PR #37, opened 2012-09-12
Real time simulator implementation.

esteban-aliverti:
The idea is to execute a simulation that actually respects the time between its steps. 
So if step1 is in t=0 and step2 is in t=1000, the simulation will wait approximately 1 second between these 2 steps.
This implementation also allows you to use a factor to reduce or increase the speed of time (believe it or not :) )


mswiderski:
closing as it did not have any attention for a while and can't be merged due to conflicts, feel free to reopen

PR #113, opened 2015-01-13
JBPM-4537 - Error when getting a primitive type process instance variable when using the REST API

jesuino:
JBPM-4537 - Error when getting a primitive type process instance variable when using the REST API

The fix will simply check if use is requesting the value of a primitive type and return build the response without trying to parse it.

Another more complex solution would be create a wrapper object for primitive types.


mrietveld:
William, 
Maciej has added this solution recently: 1b62095
Would it be okay to close this PR since we've now solved it differently? 

See this comment for why wrappers are necessary: https://bugzilla.redhat.com/show_bug.cgi?id=1119560#c6

jesuino:
Hello Marco,

This is ok for me. I saw the Maciej's solution and told customer that this will be solved in BPM Suite 6.1.

PR #114, opened 2015-01-14
kie-server: 'admin's should be able to use the remote api

psiroky:
@etirelli please take a look at this. I think the KIE Server remote API should be accessible also for users with role 'admin', to be consistent with KIE Workbench.


etirelli:
@psiroky I am not so sure about this. Admin is a workbench role, not an exec server role... lets talk on IRC.

psiroky:
Closing, outdated PR. The changes did not land in as there was not clear agreement as whether to include it or not.

PR #115, opened 2015-02-12
Remove generated sources committed by mistake

yurloc:
These were probably unintentionally added in ba979143e2 which removed kie-remote-ws-wsdl-cmd module completely.

@mrietveld please confirm and also cherry-pick to 6.2.x.


mrietveld:
Merged to master -- the files don't exist on 6.2.x! 

Thanks @yurloc!

PR #117, opened 2015-03-03
Ordered Builders: 1195772 - Remote Java API builders provide too much options

mrietveld:
@ibek and @rsynek: could you look at this and let me know what you think? 

In general, I don't like adding constructors because they're not "forward-compatibile": each constructor is unique. 

I've taken the fluent API and imposed a second set of interfaces on top of it: with the "ordered" builders, you can only call ".build()" when all of the correct information has been added, regardless of which path/configuration you choose for.

### DIFF
--- a/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/api/RemoteRuntimeEngineFactory.java
+++ b/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/api/RemoteRuntimeEngineFactory.java
@@ -9,6 +9,8 @@ import javax.naming.NamingException;
 
 import org.kie.api.runtime.manager.RuntimeEngine;
 import org.kie.remote.client.api.exception.RemoteCommunicationException;
+import org.kie.remote.client.api.order.OrderedRemoteJmsRuntimeEngineBuilder;
+import org.kie.remote.client.api.order.OrderedRemoteRestRuntimeEngineBuilder;
 import org.kie.remote.services.ws.command.generated.CommandWebService;
 
 /**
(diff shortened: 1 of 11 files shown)


ibek:
Hi Marco,

ok, the EAP SSL example test seems to be alright. However, I tried to use the builders for other use cases below where I have some questions:

# EAP example without SSL
RemoteRuntimeEngineFactory.newOrderedJmsBuilder()
       .addUserName("")
       .addPassword("")
       .addHostName("")
       .addJmsConnectorPort(1)
       .disableTaskSecurity()
       .addJbossServerHostName("") // why do we need to set jboss server host name when we already set the host name?
       .addDeploymentId("").build();
    
# WAS example
RemoteRuntimeEngineFactory.newOrderedJmsBuilder()
       .addUserName("")
       .addPassword("")
       .addHostName("")           
       .addJmsConnectorPort(1) // from my experience host name and jms connector port cannot be used with WAS or WLS because it overrides the connection factory and queues
       .disableTaskSecurity()
       .addConnectionFactory(null)
       .addKieSessionQueue(null); // further, I am not able to set other queues, deploymentId and build the builder. It just offers to set host name again :(
       
We have discussed it with Radek and he has an alternative idea which will follow my comment.

rsynek:
Marco,

I think the main problem is we need EAP specific setup - that uses HornetQ directly due to SSL requirement and we need setup for other containers as well - setup, that will use JNDI lookup to get CF and queues. How about providing a EAP specific JMS builder and general EE compatible builder (with lookup) ? Maybe it could be much simpler in such a way, we don't need any order on parameters setup, do we?

mrietveld:
@rsynek and @ibek: I haven't forgotten about this -- I've just been too busy with other things, although I've thought about this frequently. 

I think @rsynek has a good point with regards to a EAP specific setup and a EE setup -- what it made me realize is that there should probably be an API that has less options but can be used to setup any server, whether it's EAP, WAS or WLS. That means that the API won't have any shortcuts options like  `.addJbossServerHostName(String)`, for example. 

I'll rewrite this pull request and do a `git push -f ordered-builder` and then see you what you guys think.

mrietveld:
Hmm.. maybe I need to add a required(...) method to the builders? 

In essence, there are 2 problems: 
- It's not obvious or easy to see which information is required 
- More things need to be able to be changed, nothing should be hard-coded

PR #118, opened 2015-04-22
[DROOLS-729] Android support

kedzie:
* drools-android module added
  * Dex Classloaders for generated classes
  * Roboguice extension for injecting KieBases from pre-serialized packages or KieClasspathContainer
  * Robolectric testing support.  *(Shadows for the Dex classloaders)*
* drools-examples-android module added.
  * KieContainer sample *(WITH drools-compiler dependency)*
  * Roboguice KieContainer sample *(WITH drools-compiler dependency)*
  * Preserialized KieBase sample *(without drools-compiler dependency)*
  * Roboguice Preserialized KieBase sample *(without drools-compiler dependency)*
* Robolectric unit tests (verify injection and rule execution)
* Instrumentation tests (on-device tests to verify execution of rules and injection)


kedzie:
Rebased it.  We need to figure out how to handle the remaining jars not found in maven central.  Again, I included a script which install them from the local Android SDK.  Or we can deploy them to JBoss.  Gotta make a decision to move this forward.

psiroky:
@kedzie thanks for the updates. About the remaining jars. I think one of the requirements here is that the build works on checkout. We should not force people to install Android SDK just to get the build working. If it is possible (licensing) to deploy them into the JBoss repo, then that would probably work. I will look into that and let you know what would be the preferred way for us.

kedzie:
Android SDK is required to build the sample projects.  And again let me list the 3rd party jars not found in maven central.  @mariofusco has already corresponded with JBoss legal about 3rd party jars and said there wouldn't be any issues:

1. Apache Harmony *javax.beans* package. I stripped the *javax.beans* package from the harmony jars using the (jarjar)[https://code.google.com/p/jarjar/] tool. Android OS doesn't have that package. If there is another existing jar in maven central which has these that could be used. Otherwise (this)[https://github.com/kedzie/droolsjbpm-integration/blob/master/maven_repo_3rd_party/org/apache/harmony/beans-android/1.0.0/beans-android-1.0.0.jar] needs to be deployed to some JBoss Maven Repository.
2. [Android DX library (DEX Converter) version 1.10](https://github.com/kedzie/drools-android/blob/master/maven_repo_3rd_party/com/android/dx/1.10/dx-1.10.jar).  I packaged this from Android Open Source repositories (DX)[https://android.googlesource.com/platform/dalvik/dx] & (Dex)[https://android.googlesource.com/platform/libcore/dex].  It is also available in the Android SDK at *$ANDROID_HOME/build-tools/21.1.2/lib/dx.jar*.  I wrote an (install-dependencies.sh)[https://github.com/kedzie/droolsjbpm-integration/blob/master/drools-android/install-dependencies.sh] script which will install it from the Android SDK into your local maven repo.
3. Android multidex and multidex-instrumentation libraries.  They are both found in the Android SDK extra maven repositories. com.android.support:multidex:1.0.1:aar & com.android.support:multidex-instrumentation:1.0.1:aar

psiroky:
I honestly don't know what to do here. I wanted to make this work without the need for the Android SDK (to at least compile things). I still think that we can not simply put this burden on all the users that are not interested in the Android. We could somehow download the SDK as part of the build, get it to download the jars wen need (as they are not there by default). However, the SDK has more than 150MiBs, which is quite a lot. Also this is basically the same as downloading from arbitrary 3rd party repo, which we are trying hard to avoid. 

I have been also looking at the Android SDK Terms a Conditions and the following paragraph sounds like it is prohibited to copy parts of the SDK (e.g. the jars) other for backup purposes:

`3.3 You may not use the SDK for any purpose not expressly permitted by this License Agreement.  Except to the extent required by applicable third party licenses, you may not: (a) copy (except for backup purposes), modify, adapt, redistribute, decompile, reverse engineer, disassemble, or create derivative works of the SDK or any part of the SDK; or (b) load any part of the SDK onto a mobile handset or any other hardware device except a personal computer, combine any part of the SDK with other software, or distribute any software or device incorporating a part of the SDK.`


Or maybe the `Except to the extent required by applicable third party licenses,` applies here? I am not sure.

triceo:
@psiroky Did you think to look at other projects that come with Android support? Surely they must have solved the problem of Android SDK dependencies.

Alternatively, can we maybe create a separate module, perhaps disabled by default, that would only allow for Android compilation on-demand?

psiroky:
I was looking at some other projects, e.g. spring-android integration, but they usually only need to depend on publicly available jars like `com.google.android:android` (this one is available in Maven Central). The `drools-android` module, however, needs more exotic stuff like the `dx.jar`, `com.android.support:multidex` and ` com.android.support:multidex-instrumentation`. These seem to be available only in the SDK (and only after downloading some extras, they are not there with the default zip). 

If we want to release the `drools-android` module then it should be in the `full` profile (if not in the standard non-full build). That would be definitely more pleasant for people, as they are not usually using the full profile. The CI builds are mainly using them (which brings the burden of the SDK there, but at least it is more isolated). I would like to avoid situation where we have some other profile, android-only, which needs to be triggered only in some specific cases (but it may be the only feasible solution...)

triceo:
+1 for using another profile for the Android stuff. But inclusion and/or exclusion of the profile could be fully automated! Maven has profile activation conditions. So the "android-only" profile could be activated by the additional files being present somewhere, for example.

kedzie:
About the SDK default contents, there are command line tools to install additional SDK components (like the SDK maven repos, where the needed jars are found) which are commonly used for CI builds.

kedzie:
What about the harmony jar? We need the java.beans package.  Do u guys know of an existing lib in maven central which has these?

kedzie:
As far as the license, @mariofusco corresponded with JBoss legal so I would  ask him.  And again the actual SDK is only needed for the sample projects, not for drools-android itself.  Drools-android does need the DX jar though.
What if we only activated the android profile when we detect the android SDK is installed? We could activate based on existence of $ANDROID_HOME...   I dunno if profile file activation can use environment variables, but probably does.

psiroky:
Yes, we should be able to do the profile activation based on property or environment variable. I wanted to avoid this, because this would require changes in our build process, as we need to make sure the Android SDK is always present, as we want to compile/run the Android related modules frequently to catch possible issues as early as possible. 

I wasn't able to find existing library that would we could use instead of the `jarjar`-ed one.

kedzie:
I pushed the pom profile changes.  Which adds the drools-android and drools-examples-android modules based on the android SDk being present at ${env.ANDROID_HOME}.  And rebased again.

kedzie:
I wrote some documentation @psiroky .  Check it out droolsjbpm/kie-wb-distributions#65

mrietveld:
@kedzie and @psiroky: I corresponded with legal about the Android SDK jars in March. Legal gave their approval.

@psiroky I'll forward you the responses from legal.

psiroky:
I did some (minor) updates and merged the PR into master: https://github.com/droolsjbpm/droolsjbpm-integration/commit/631a8f83d76e067f50f46f36a052b01e2f307027

Summary of my additional changes:
 * dependencyManagement moved into top-level kie-platform-bom (located in droolsjbpm-build-bootstrap repo)
 * removed some duplicated plugin configuration
 * removed the 3rd party repo with jars. dx.jar and dex.jar were uploaded into JBoss 3rd party uploads repo. The beans-android.jar is not needed to build the drools-android project, so it was removed for now. I am guessing it will be needed for the examples, which still need some more care.
 * it may very well happen that the android examples are now broken, I will be looking into that
 * removed @author tags, added license headers and fixed some minor formatting issues

@kedzie thanks for the PR and sorry that it took so long to merge this, it was a tough one.

mrietveld:
@psiroky Thank you for keeping on top of this! 
@kedzie Thank you for your patience.

kedzie:
The beans-android.jar is needed during runtime for any app using drools-android (including the examples).

kedzie:
@psiroky any thoughts on how to proceed with this?  This lib is definitely needed for runtime.  Android JVM doesn't have the javax.beans classes.  Which drools uses a lot.  So this dependency needs to be there during runtime for anything using drools on android.  What our options?  What's wrong with just deploying the jar to JBoss repos like we did with the dx?

mrietveld:
@kedzie Usually, we leave runtime dependencies to the users: in this case, my opinion is that we should make sure to document (everywhere possible.. :) ) that the jar is needed. Is that an option here? 

One thing that we are unfortunately extremely strict about on is that our repositories are source code repositories: we do not store binary artifacts in our repositories.

psiroky:
I am not sure what is the best solution here. For users it would definitely be easier if the `drools-android` module brought the beans jar with it. However, I see some issues with the https://github.com/kedzie/droolsjbpm-integration/blob/36b828137567b0e3102d3041bd9bdffd4ef1da17/maven_repo_3rd_party/org/apache/harmony/beans-android/1.0.0/beans-android-1.0.0.jar . Looks like it only has some subset of the java.beans classes? At least when looking at https://openbeans.googlecode.com/files/openbeans-1.0.jar it has more classes (but maybe it is just because of the additional AWT and Swing stuff).

@kedzie could you please point to the sources of that beans-android jar?

@mrietveld I uploaded two required artifacts into the JBoss 3rd party artifacts repo. See http://repository.jboss.org/com/android/dex/ and http://repository.jboss.org/com/android/dx/. These were originally part of the PR (so committed to the source repo).

kedzie:
Documenting that developers need to package their own javax.beans classes
into a jar file doesn't seem reasonable to me...   Nor does including
broken examples.  Not to mention broken instrumentation tests within those
examples .  the automated tests prove that rules execute on an android
device,  which is he only real way to test the drools-android module.  (And
I think u agree it needs to have tests.)

i don't have the sources.. I stripped the javax.beans classes from an
apache harmony jar using the jarjar tool.  If it is incomplete that was my
mistake; we can make a new jar if you want.  I couldn't find apache harmony
in maven central or else I might have referenced that..

Bottom line is that the android JVM doesn't have the beans package and it
is required by drools.  I looked at openbeans but it uses a different
package name.. So we would need to refactor drools to use openbeans instead
of the jvm javax.beans.

Is there any reason we cannot deploy this jar like we did the dex,dx jars?

psiroky:
There reasons are currently the following:
 * I want to be sure we are not breaking any license (thus wanted to look at the sources)
 * I want to be sure the jar is complete and has all the classes that it needs. If we upload the jar it should be ideally usable also by others.

psiroky:
@kedzie I tried to use the beans-android-1.0.0.jar as a dependency, but the examples are failing for me (looks like it may not be related to this jar). I am getting `java.lang.UnsupportedOperationException: can't load this type of class file`

I was using the `serialized-kbase` example and Android Emulator (Nexus 5  with API version 21). I started the emulator and the installed the package using `./adb install serialized-kbase.ap`. After starting the app and clicking "Fire rules" nothing happened. 

See https://gist.github.com/psiroky/4a40d4254ad4571c511f for more details. Any idea what could be wrong?

kedzie:
Yes that is not related. But concerning.. Which jvm did you use for the
build?  I will try to run it in its current state.  Did you try any other
examples?  Try the robo kiecontainer.

psiroky:
I used Oracle JDK 1.7.0_80.  I've tried the `kiecontainer` example as well, with same result.

kedzie:
@psiroky sorry for late response.  Could be class version issue.. try building with JDK 6.  I know that the maven plugin should force source level 1.6 when it generates the classes.. I will make sure it does.

kedzie:
@psiroky I did verify that the maven plugin goal with serializes the rules doesn't force the source level to 1.6 with a system property.  I will add this.  For the KieContainer example, which doesn't preserialize rules, it uses the normal kie-maven-plugin kiebase caching.  I am looking to see if there is a mechanism there to force 1.6 source level.

kedzie:
@psiroky  Good news.  I did manage to fix the kie-maven-plugin to force 1.6 source level with new JDK's.     I will create a new PR with that change, I just added system properties parameter to the mojo and used it to pass the lnglevel property to force 1.6

Bad news is that when droolsjbpm-integration got merged some dependencies got removed from drools-android which broke some examples.  This consists of the harmony beans and the jaxp parser reference implementation. This is required for drools to parse the kmodule xml files.. (if using drools-compiler and the KieContainer API).

kedzie:
Im pushing a new PR which fixes all the examples.. you guys can decide what to do about the remaining 3rd party deps...

PR #119, opened 2015-04-28
BZ-1214927: adding test for decision tables in spreadsheet

sutaakar:
BZ-1214927: Test verifying that spreadsheet decision table is processed and evaluated in Execution server.


psiroky:
Thanks!

PR #120, opened 2015-04-29
kie-spring: improved PessimisticLockingSpringTest to handle different transaction timeouts

sutaakar:
Improved PessimisticLockingSpringTest to work against different databases. Can handle case when transaction timeout is set to 0, to value smaller or bigger than waiting time between start of thread 2 and commit in thread 1.
Method getProcessInstance in thread 2 is set to be read only to differentiate from cases when pessimistic locking is turned off - in which case process instance is returned immediately. Invoking method abortProcessInstance(in previous test version) contains writing to database which causes waiting on thread 1 to unlock affected rows even if pessimistic locking is turned off.

### DIFF
--- a/kie-spring/src/test/java/org/kie/spring/jbpm/PessimisticLockingSpringTest.java
+++ b/kie-spring/src/test/java/org/kie/spring/jbpm/PessimisticLockingSpringTest.java
@@ -1,5 +1,8 @@
 package org.kie.spring.jbpm;
 
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
 import java.io.File;
 import java.io.FilenameFilter;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


mswiderski:
overall looks good, just one thing that I think we don't need:
> Method getProcessInstance in thread 2 is set to be read only to differentiate from cases when pessimistic locking is turned off - in which case process instance is returned immediately.

since this test is explicitly for pessimistic lock testing don't think we should add anything that allows to run it without pessimistic lock being configured. If we want that then test should be renamed and both optimistic and pessimistic lock test methods should be implemented. Wdyt?

sutaakar:
That comment was meant as explanation why I set read-only flag in getProcessInstance method. Without it test would behave same for both types of locking and we wouldn't be able to verify that pessimistic locking is really set.
Test is done just for pessimistic locking, with optimistic locking it will fail.

mswiderski:
understood, will merge this later today. thanks @sutaakar

mswiderski:
merged into master https://github.com/droolsjbpm/droolsjbpm-integration/commit/b0ecce1e6fdd0bd973256abf8626140973faee76

PR #121, opened 2015-05-04
BZ-1214927: adding test for decision tables in spreadsheet

sutaakar:
BZ-1214927: Test verifying that spreadsheet decision table is processed and evaluated in Execution server.
Cherrypicked from master.


manstis:
Can you merge this without the applicable flags against a BZ?

Sent on the move

psiroky:
@manstis the referenced BZ should be https://bugzilla.redhat.com/show_bug.cgi?id=1217296 which has the flags. However, since the commit was simply cherry-picked from master, it references the original BZ. I agree this is confusing and I've seen that on other commits too. We need to define some policy for commits to 6.2.x branch.

PR #127, opened 2015-05-20
JBPM-4651 - kie-spring jBPM - refactored and added integration tests

sutaakar:
Implementing [https://issues.jboss.org/browse/JBPM-4651](https://issues.jboss.org/browse/JBPM-4651) .

Main changes:
- created test class for basic combination of all runtime manager types X transaction configuration: RuntimeManagerInitSpringTest - most tests from JTA(Local)...SpringTest moved there
- other tests from those test classes moved to their separate special test classes: SwimlaneSpringTest, ProcessFlowSpringTest, PerProcessInstanceSpringTest
- simplified PessimisticLockingSpringTest by adding inheritance from AbstractJbpmSpringTest
- added MultipleRuntimeManagersTest, RuntimeManagerInitNoInitialContextSpringTest to verify functionality which wasn't covered yet

Main idea behind this change is to remove code duplicities and clarify test class responsibilities. Means that now every test class has specified functionality which it tests and can set different environments for tests as it needs(using Parameterized runner).
Tried to name test classes according to what they tests, if you think up better names please tell me so.

What do you think about this change?


mrietveld:
@sutaakar This looks good. 

I'll push a commit after we merge it showing how you can avoid the Thread.sleep(..) methods in tests, and instead use java concurrent mechanisms (notify/wait/etc) to achieve the same result.

PR #129, opened 2015-05-29
DROOLS-733 - splitted and refactored work item operations test

sutaakar:
Splitted testWorkItemOperations. 
User task shouldn't be handled as work item because in such case it doesn't behave consistently: in database there was left task in ready status even when process instance was completed. That caused issues in other tests.
In this PR user task is completed using startTask/completeTask and work item operation tests are done on second task - Email work item.
Test was splitted to cover both abortWorkItem and completeWorkItem.


mswiderski:
merged into master, thanks @sutaakar

PR #130, opened 2015-06-03
DROOLS-733 - Added tests for forwardTask and delegateTask

sutaakar:


mswiderski:
merged int execution-server-unification2 branch, thanks @sutaakar and keep the PRs coming :)

PR #133, opened 2015-06-04
Changed .contains to .containsKey

selrahal:
Using contextsCache.contains(String) will compare agains values in the cache, not keys. Use .containsKey instead.


mrietveld:
Great catch! Thanks very much, Sal. 

Merged into master: 7b493520

PR #135, opened 2015-06-05
Adding a possible fix for BZ-1228853

jesuino:
Possible fix for BZ-1228853. It will not convert to lower cases query parameters that starts with var_


mrietveld:
@jesuino Thanks for the suggestion! However, the special params should also be set to lowercase, for example: "VAR_ParamName" needs to be changed to "var_ParamName". 

I'll push a fix for the BZ today.

PR #136, opened 2015-06-10
DROOLS-733 - improved UserTaskServiceIntegrationTest.testSkipUserTask test

sutaakar:
Test improvement showing possible bug.
taskSummary.getSkipable() returns false although task has defined Skippable data input which is set to true and is possible to skip task by calling client.skipTask(...).

Please try to run this test.


mswiderski:
merged as part of other changes and fixed problem with skipable property not being set by https://github.com/droolsjbpm/jbpm/commit/7df638d98c1f86f6c30fdc266041027e6e5a47b1

PR #138, opened 2015-06-15
move drools-osgi-integration back to droolsjbpm-integration

mariofusco:


psiroky:
Merged into master as we need it to correctly install the features xml into Karaf/Fuse. https://github.com/droolsjbpm/droolsjbpm-integration/commit/0d555ae709b7120d33309fd75b17e35587e1bc52

PR #139, opened 2015-06-17
Fix for wrong {version.org.drools} instead of {project.version} in fe…

lazarotti:
…atures.xml, to allow one-off patches


psiroky:
Merged into both 6.2.x and master. Thanks @lazarotti!

PR #140, opened 2015-06-17
drools-karaf-features: update cxf-specs version range to support Fuse 6.2

psiroky:
* the commit needs to be cherry-picked into 6.2.x as well


psiroky:
We need to investigate if we still need to depend on the cxf-specs. For now the safest thing seems to be just updating the version range.

PR #141, opened 2015-06-18
fixed optaplanner osgi feature

snurkabill:


snurkabill:
Geoffrey, after this I was able to install optaplanner-engine into fuse. So we still don't know if planner is runnable, but it's feature can be definitely installed

ge0ffrey:
Thanks Jiri

PR #142, opened 2015-06-19
DROOLS-821 OSGi modules clean-up + adding back Karaf integ. tests

psiroky:
@mariofusco please review.


drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/KieSpringjBPMPersistenceOnKarafKarafIntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.kie.api.KieServices;
+import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.EnvironmentName;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.manager.*;
+import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.api.task.TaskService;
+import org.kie.api.task.model.TaskSummary;
+import org.kie.internal.runtime.manager.context.EmptyContext;
+import org.ops4j.pax.exam.Configuration;
```
triceo: @winklerm I see these tests are using pax-exam, same as we are. Could we perhaps merge our stuff in here?

winklerm: @triceo  I believe we could. We'd need to:
- enhance PAX-EXAM configuration to be able to use other containers than community Karaf, 
- add a mechanism for configuring other Maven repositories for the OSGi container, (some) OSGi containers do not read repositories settings from Maven settings.xml,
- probably also set up profiles for different versions of the OSGi container (right now we need it, not sure if it will be needed for future versions).

psiroky:
I've added one more commit that uses dependency-plugin to download and unpack the *orm files that are used by the tests. This way we always get the latest updates and thus manual updates are no longer necessary.

mariofusco:
Squashed and merged https://github.com/droolsjbpm/droolsjbpm-integration/commit/33f0e40a7a3562bc1757f0b9cf97200c338374de

PR #143, opened 2015-06-22
KIE Performance Kit

ibek:
This is kie-performance-kit for performance tests of jbpm engine, remote APIs and other internal tests. It needs to be built together with other modules to be in the Maven repository, so the kit could be used anywhere.

kie-performance-kit: refactoring of executor
kie-performance-kit: Added warmup
kie-performance-kit: added runtimeManagerStrategy
kie-performance-kit: Fixed metrics during warmup
kie-performance-kit: added pessimistic locking and concurrent user count properties
kie-performance-kit: Added CsvSingleReporter and MemoryUsageGaugeSet
kie-performance-kit: extended TestConfig, added support for optional measures MemoryUsage, FileDescriptors and ThreadStates
kie-performance-kit: Format report output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,11 +4,15 @@
 
 # Eclipse, Netbeans and IntelliJ files
 /.*
-!.gitignore
+.gitignore
 /nbproject
 /*.ipr
 /*.iws
 /*.iml
+.classpath
+.project
+.settings
+target
 
 # Repository wide ignore mac DS_Store files
 .DS_Store
--- a/kie-performance-kit/pom.xml
+++ b/kie-performance-kit/pom.xml
@@ -1,0 +1,81 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.drools</groupId>
+        <artifactId>droolsjbpm-integration</artifactId>
+        <version>6.3.0-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.kie</groupId>
+    <artifactId>kie-performance-kit</artifactId>
+
+    <name>BPMS Performance Kit</name>
+
+    <properties>
+        <logback.version>1.1.2</logback.version>
+        <metrics.version>3.1.0</metrics.version>
+        <slf4j.version>1.7.7</slf4j.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.dropwizard.metrics</groupId>
+                <artifactId>metrics-core</artifactId>
+                <version>${metrics.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.dropwizard.metrics</groupId>
+                <artifactId>metrics-jvm</artifactId>
+                <version>${metrics.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>ch.qos.logback</groupId>
+                <artifactId>logback-classic</artifactId>
+                <version>${logback.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-api</artifactId>
+                <version>${slf4j.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.reflections</groupId>
+                <artifactId>reflections</artifactId>
+                <version>0.9.10</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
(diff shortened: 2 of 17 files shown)


mrietveld:
@ibek Is this module used in other repositories besides the kie-tests repository? 

If it's only used in kie-tests, would it be possible to have the module first located in the kie-tests repository? 

I'm curious about the plans for usage of this module: could you share what you want to do?

ibek:
@mrietveld I need this dependency also for internal performance use cases. That's why it had been put here, so it gets into Maven repository. Therefore it won't be only used in kie-tests. Also I can imagine that the kie-performance-kit could be used by users to evaluate their scenarios how they perform considering the performance, memory usage, etc.

mrietveld:
Ah, nice! (especially user-usage to evaluate their own scenarios). 

If you could add some preliminary documentation (either to the droolsjbpm-integration wiki as a draft) or directly to the docs modules in kie-wb-distributions, that would be great! It doesn't have to be complete -- even a very simple chapter with a couple of sentences is fine as a starting point. 

Thanks!

ibek:
Maciej, asked me for a documentation for jbpm-performance-tests yesterday, hence I created this README https://github.com/ibek/kie-tests/blob/performance-tests/performance-tests/jbpm-performance-tests/README.md  

Ok, I will write also documentation how to use kie-performance-kit on the example of jbpm-performance-tests because it is a little different topic than how to execute a test suite or scenario explained in the readme above.

mrietveld:
Cool! thanks. 

Re: kie-performance-kit: if you have a good "example" test with coments in the code, that's also fine. I'll leave it up to you. :)

ibek:
Here is the preliminary documentation for KIE Performance Kit https://github.com/droolsjbpm/droolsjbpm-integration/wiki/KIE-Performance-Kit

.gitignore
```diff
@@ -4,11 +4,15 @@
 
 # Eclipse, Netbeans and IntelliJ files
 /.*
-!.gitignore
+.gitignore
```
psiroky: Should we really ignore the `.gitignore` file? I think this file is supposed to be committed into the source repo.

kie-performance-kit/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <parent>
+        <groupId>org.drools</groupId>
+        <artifactId>droolsjbpm-integration</artifactId>
+        <version>6.3.0-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.kie</groupId>
+    <artifactId>kie-performance-kit</artifactId>
+
+    <name>BPMS Performance Kit</name>
```
psiroky: >KIE< Performance Kit?

kie-performance-kit/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    <properties>
+        <logback.version>1.1.2</logback.version>
+        <metrics.version>3.1.0</metrics.version>
+        <slf4j.version>1.7.7</slf4j.version>
```
psiroky: Is there a reason to override the versions of `logback` and `slf4j` coming from parent?

mrietveld:
@Ibek In the future, if you could rebase PR's instead of merging them, that would be great! 

We don't really have a clear policy in KIE (AFAIK?), but my personal preference is for rebases, because it makes the history much easier to read. 

If it's a *really* big PR (+20 files, +500 LoC's, etc.), then a merge is better, I think. Otherwise, rebases are better.

ibek:
Marco, I do not like the additional commits too but do you have any guide how to rebase pull requests in GitHub? I just used the merge button here but I do not know about any other option.

etirelli:
@ibek There are many ways to do that. My preference is to squash commits when merging, so I do this (assuming you are merging into master):

1. create a temporary local branch on your clone

    $ git checkout -b tempBranch

2. fetch the pull request from github as a patch (just add the .patch extension on the URL):

    $ curl -L https://github.com/droolsjbpm/droolsjbpm-integration/pull/143 | git am

3. checkout master

    $ git checkout master

4. squash commits

    $ git merge --squash tempBranch

5. check that the changes are fine, solve any conflicts, commit and push

    $ git status
    $ git commit
    $ git push

In this example I did not "rebase" the pull request, but I squashed the commits and added them as a single commit on top of master, but the result is quite similar. If you do want to rebase for some reason, then you can add the remote repository instead, fetch the branch and rebase prior to squashing and merging.

ibek:
Thank you Edson, I will try it this way next time. I did not know about the possibility to get pull request as a patch. :)

mrietveld:
@Ibek I use the following bash function: 

    gpr () { $git fetch $1 pull/$2/head:$3; git checkout $3; }

(I'm using git 2.4.1, I'm not sure which version started supporting the 
`git fetch <repo> pull/<num>/head:<branch-name>` operation.)
 
If I were to rebase this pull request, for example, I would do the following: 

    $ git remote update 
    $ gpr upstream 143 name-of-pull-request-branch # the bash function above
    $ git checkout name-of-pull-request-branch
    $ git rebase upstream/master
    $ git checkout master
    $ git merge --ff name-of-pull-request-branch
    $ git push upstream master  

I hope that helps?

PR #147, opened 2015-07-01
kie-performance-kit: removed overriding dependency versions and enabl…

ibek:
…ed creating new folder for report data


psiroky:
+1 for the dependency related changes.

+0 for the the report directory ones (I don't have enough context there).

PR #148, opened 2015-07-03
BZ-1221491 - added tests covering commands newly supported in execution server

sutaakar:
Test coverage for BZ 1221491.


etirelli:
Merged.

PR #151, opened 2015-07-13
DROOLS-733 - Added test verifying task properties setting.

sutaakar:
Test showing marshalling issues of methods:
 - setTaskDescription
 - setTaskExpirationDate
 - setTaskName

Also there is one inconsistency, for expiration date retrieval from task object is used method getExpirationTime, but for setting value in client is used setTaskExpirationDate.
Could be good to unify it, choose one of Date or Time.


mswiderski:
as it comes for the inconsistency, it was done on purpose as you work in different contexts. When you're in taskInstance context there is not much sense to repeat the Task word in the method name, but when client is the context stating explicitly that we are setting task details and not e.g. deadline, assignments etc is rather useful in my opinion. 

Will merge the test and fix the issues. thanks @sutaakar

mswiderski:
test incorporated, inconsistency in names fixed as we discussed on irc and problem with marshaling fixed as well. thanks @sutaakar and keep the PRs coming :)

PR #153, opened 2015-07-15
DROOLS-733 - Added test verifying task attachments.

sutaakar:
Tests showing 2 possible problems:
 - Attachment objects returned by client don't have attribute name (element attachment-name) set, didn't find a way how to set it.
 - Sending byte array as attachment fails for JAXB and JSON, should it be supported?


mswiderski:
just to keep this PR updated, I merged and fixed the issues locally but will push it tomorrow with bunch of other changes that I have in place ... so will keep it open until the push, thanks @sutaakar

mswiderski:
pushed to execution-server-unification2 branch

PR #154, opened 2015-07-17
DROOLS-733 - Added tests covering client functionality.

sutaakar:
Tests verifying exitTask() functionality and and correlation keys.
Shows error in getProcessInstance() method which doesn't return correlation key - test testStartCheckProcessWithCorrelationKey().


mswiderski:
test cases merged and issue fixed, thanks @sutaakar

PR #156, opened 2015-07-20
DROOLS-733 - Fixed nondeterminism in jbpm executor tests.

sutaakar:
I didn't find a way how to get rid of Thread.sleep(), but at least I got rid of nondeterminism caused by timing of job executor.


sutaakar:
Added job executor test for getRequestsByCommand(), minor client fix.

mswiderski:
merged into execution server unification branch, thanks @sutaakar

PR #159, opened 2015-08-03
DROOLS-733: kie-server-client refctored tests with wiremock to execut…

mswiderski:
…e them without a need to have running kie server

### DIFF
--- a/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/util/PaxExamWithWireMock.java
+++ b/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/util/PaxExamWithWireMock.java
@@ -1,0 +1,665 @@
+package org.drools.karaf.itest.util;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.net.ServerSocket;
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.model.InitializationError;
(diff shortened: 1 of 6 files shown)


mswiderski:
following our discussion on improving kid-server-client tests in osgi/karaf, here is a PR that does that using suggested WireMock. @psiroky if you could take a look whenever you have some time.

drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/util/PaxExamWithWireMock.java
```diff
@@ -0,0 +1,665 @@
...
+import com.github.tomakehurst.wiremock.WireMockServer;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.model.InitializationError;
+import org.ops4j.pax.exam.junit.PaxExam;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.*;
+
+public class PaxExamWithWireMock extends PaxExam {
+
```
psiroky: Maybe we could use `PaxExamParameterized` it basically combines PaxExam with default jUnit's Parametrized runner. We could hook the wiremock as the tests are usually the same for XML and JSON?

psiroky:
+1. After talking to @mswiderski over IRC, it is not easily possible to use the `PaxExamParametrized` (will be explaining why as a comment in the test class).

mswiderski:
merged into master and 6.3.x

PR #161, opened 2015-08-10
DROOLS-733 - added tests for findProcessInstancesByCorrelationKey

sutaakar:
+ minor enhancements.

Paging doesn't work, it would need change in RuntimeDataService - adding QueryContext parameter for getProcessInstancesByCorrelationKey method.


mswiderski:
issues with correlation key paging fixed and this pr merged into master and 6.3.x, thanks @sutaakar

PR #162, opened 2015-08-11
BPMSPL-202 - REST Query APIs with POST method

mrietveld:
Functionality to invoke REST Query APIs with POST method. 

Tested here: https://github.com/mrietveld/kie-tests/commit/2a012b58416091f6585150932b04397d32ba9c59 (Waiting for a PR (commit before this) to push it to the blessed repo).

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/QueryResourceImpl.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/QueryResourceImpl.java
@@ -28,6 +28,7 @@ import java.util.Map.Entry;
 import javax.enterprise.context.RequestScoped;
 import javax.inject.Inject;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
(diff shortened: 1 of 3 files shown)


mswiderski:
Marco, isn't this sort of abuse of REST API to use POST for queries (which are essentially read operation that should be implemented as GET)? I see this as similar thing as we have the execute end point that does not really fit into REST api.

Wouldn't it be better to somehow encrypt the parameters instead of using post? I read the jira and security issue has some value but the other with xml for staring process does not belong here - start process is already POST so it should always use body for the data.

mrietveld:
Rescinding this PR because of Maciej's comment: I indeed do _not_ want to extend the REST API in this way.

mrietveld:
Hmm.. actually, after rereading the BPMPSL issue, there are 2 valid issues: 

1. Even if you use HTTPS/SSL, using GET means that the query parameters are in the URL and thus are exposed. 

2. POST allows the user to post a far larger content (with regads to query parameters) than with GET. 

@mswiderski both of these points are valid with regards to the Query API. I agree with your points here, but given that both points are valid (and that I also think that security is an important issue), I'm thinking of going through this anyways..

mswiderski:
> Even if you use HTTPS/SSL, using GET means that the query parameters are in the URL and thus are exposed. 

is this really valid for REST operations? Usually you don't expose REST endpoints in like a browser URL bar or similar. You either use JavaScript (like XmlHttpRequest) or some other type of client library that do not expose the URL any way.

> POST allows the user to post a far larger content (with regads to query parameters) than with GET.

that's true but what is the use case to send big amount of data when quering for items via given end point? Isn't the idea of the GET endpoints to return (possibly big) data?

mrietveld:
Closing this pull request, since I don't think we want to implement this (because using POST in this case is an anti-pattern).

PR #163, opened 2015-08-12
JBPM-4682 kie-performance-kit: added cpu usage metrics

ibek:


kie-performance-kit/src/main/java/org/kie/perf/Executor.java
```diff
@@ -168,7 +171,18 @@ public static void main(String[] args) {
...
+                
+                CPUUsageHistogramSet cpuusage = null;
+                boolean cpuusageEnabled = tc.getMeasure().contains(Measure.CPUUSAGE);
+                if (cpuusageEnabled) {
+                    cpuusage = CPUUsageHistogramSet.getInstance(scenario.getClass());
+                    cpuusage.start();
+                }
                 testSuite.startScenario(scenario);
+                if (cpuusageEnabled) {
+                    cpuusage.stop();
```
mrietveld: Cool!

mrietveld:
+10 looks very interesting! 

Feel free to merge when you're ready.

ibek:
To the commit I added also KPKConstraint so the L100MultiInstanceSubprocesses scenario and other tests are run only when the constraints are met - for that scenario it is when strategy is set to PerProcessInstance for example.

PR #165, opened 2015-08-14
[BZ-1203351] add jbpm-services-cdi into bpms-distribution

psiroky:
* the BZ is already closed as the artifact was manually added into the maven repo, but this should be the proper fix


psiroky:
@krisv, @mswiderski does this change look good?

mswiderski:
looks good to me

psiroky:
Cherry-picked into 6.3. as well.

PR #167, opened 2015-08-17
BZ1200628 - kie-remote-client: fix and tests for adding possibly null objects to commands

sutaakar:


kie-remote/kie-remote-client/src/main/java/org/kie/services/client/api/command/AbstractRemoteCommandObject.java
```diff
@@ -228,15 +228,19 @@ void preprocessParameterCommand( Object cmdObj, List<Object> extraClassInstanceL
 
     void addPossiblyNullObject( Object inputObject, List<Object> objectList ) {
         if( inputObject != null ) {
-            if( inputObject instanceof List )  { 
-                objectList.addAll((List) inputObject);
+            if( inputObject instanceof List )  {
+                for( Object obj : (List<?>)inputObject ) {
+                    addPossiblyNullObject(obj, objectList);
+                }
```
mrietveld: Could you add a check to see if the object isn't already present in `objectList` -- users sometimes do crazy things, including trying to submit self-referential nested lists, which would lead to a stack-overflow in this case. :/ (If the object has already been added, an exception should probably be thrown). 

Good catch, Karel!

sutaakar: I think that that check if the object isn't already present in objectList could be tricky as client may want to send same object several times, for example as output content with different content keys.
About self-referencing issue - good point, will try to think up some solution which would cover that.

mrietveld: I have an.. algorithm to do this, so to speak. I'll merge it and add it to the code -- unless you want to fix it yourself? :)

sutaakar: Ok, sounds good, please merge and add it :)

mrietveld:
Closed with https://github.com/droolsjbpm/droolsjbpm-integration/commit/11921ce80154985f2064e30b4e144481f5ca92e6

See https://github.com/droolsjbpm/droolsjbpm-integration/commit/fc81ecf5ec2c89c88aa8fb82b911d2db86ccbc9b for the algorithm I mention above.

PR #168, opened 2015-08-18
DROOLS-733 - aligned persistence and REST endpoints for all supported…

mswiderski:
… containers grouped by EE spec (EAP, WebLogic, WebSphere - EE6), (Wildfly - EE7) and Tomcat (web container)

this PR makes the KIE Server war assemblies to be container agnostic and be bound only to the EE spec it supports (either EE 6 or EE7). With this we cover all supported containers (especially important for EE6 where we have EAP, WLS, WAS).

Worth noting items in this PR:
- use PersistanceUnitInfo to create EntityManagerFactory in container fashion though without relying on container to parse persistence.xml - kie server does the parsing and manually creates the EntityManagerFactory - reason for this is:

1. we abstract the kie server from application specific configuration that resides in persistence.xml and this requires manual changes to that file or to have independent war files per container (similar as we have for workbench) 
2. we do not use any EE frameworks (CDI, EJB) so persistence unit does not have to be managed by container as it won't be able to inject it any way
3. such setup allows us to easily change persistence configuration (data source name, dialect, transaction manager platform, etc) without changing the war file - all done via system properties or configuration parameters of kie server

- assemblies exclude all EE spec api to allow portable deployment across different containers
- REST endpoint reconfiguration because Jersey used in Weblogic does not allow to have multiple resources classes that have same @Path set on class level even though they do not cause duplicates or conflicts as method level @Path are unique 
- special cofiguration is needed to switch class loader for WLS and WAS

1. WLS is handled automatically via weblogic.xml file that allows to configure class loader preference on individual packages
2. WAS requires to use parent last class loader setting on application level
these class loader settings are enforced by older version of Jackson libraries shipped with WLS and WAS


mswiderski:
anyone any comments or good to merge?

mswiderski:
merged into master and 6.3.x

etirelli:
A bit late, but it does look good. +1

PR #169, opened 2015-08-19
kie-server-tests: enable custom settings.xml when building kjars for testing

psiroky:
* the settings.xml file can be used to define additional repos
   to download artifacts from (e.g. when testing staged build)

@mswiderski, @sutaakar please review


mswiderski:
looks good to me (though haven't executed tests with it)

sutaakar:
Looks good to me too (didn't try it yet either).

Just one small idea, any special reason to define default value for build settings parameter? You may use constructor without default value - default will be null then.

psiroky:
The only reason was that I don't like `null`s :) But looking at the other config options now, they are not using any default value, so I will fix that to make it consistent. Thanks for pointing this out!

psiroky:
Fixed - using the constructor without default value.

mswiderski:
thanks, I'll merge locally and retest all before push

PR #171, opened 2015-08-21
kie-server-tests: fix issue when Tomcat hangs during deployment

psiroky:


mswiderski:
@psiroky will this have any impact on users running tests on windows? or we don't expect people to run these tests on windows at all? Main question is will it fail when running on windows :)

psiroky:
I haven't tried myself, but from what I could read this should work on Windows as well. To be 100 % sure, I will try on Win 2012.

psiroky:
Works as expected on Win 2012, so I am assuming it should work on any Windows machine.

PR #174, opened 2015-08-24
DROOLS-733 - align KIE Server Rest URIs to be consistent and follow s…

mswiderski:
…ame pattern - plurals in resource names


psiroky:
This will break the backwards compatibility between 6.3.x and 6.4.x, is that correct? (assuming it will only be merged into master). Or is it scheduled for 6.3.x as well?

mswiderski:
the idea is to put it into 6.3 as well to make sure it will not be a breaking change after 6.3, so yes, it should go to 6.3.x as well

PR #175, opened 2015-08-25
DROOLS-733 - Controller tests and test infrastructure

sutaakar:
First draft of controller test and infrastructure around:

Controller testing WAR files - maven project for that added to kie-server-parent/kie-server-controller folder, not sure if it is the best place for such thing, ideas for better place?

Added controller embedded server for local testing.

Controller tests: So far basic happy path tests. Deployment structure for tests is: 1 controller, 1 kie server - didn't try to emulate more kie servers on one container - not sure about configuration possibilities there, what about KieServerEnvironment values in such case.
Controller client - simple implementation based on Resteasy to make testing easier, inspired by kie server one.

Small controller fixes and current test improvements:
KieServerInstanceInfo - same @XmlRootElement value as KieServerInstance, changed to prevent marshalling issues
RestKieServerControllerAdminImpl - response code correction, delete request corrections

There are test failures:
KieControllerIntegrationTest.testGetContainer - Should container resource returned from controller return also container status?
JSON tests when running with real servers: JSON implementation in controller returns KieServerInstance instance marshalled according to @XmlElement attributes. Default Resteasy JSON marshaller isn't able to unmarshall response using those attributes, throwing UnrecognizedPropertyException. Should unmarshaller on client side support this features?

Controller tests were run on Tomcats and Wildfly 8.2, will check other containers too.

Looking forward for ideas for changes/improvements.


sutaakar:
Btw found that there were Resteasy connections in rest only tests which didn't release connection upon finish - that may had caused those rare timeouts in REST tests - connections are released now.

sutaakar:
One more point about setting org.kie.server.location/org.kie.server.controller: kie server URL is set as complete URL with path defined in @Path annotation, controller URL is set as URL without part from @Path. Maybe it would be nicer to unify it.

mswiderski:
thanks @sutaakar will take a look at this tomorrow and try to fix remaining issues to add it to the release as it does make sense to have it fixed and well tested.

others feel free to comment as well, the more feedback the better

sutaakar:
Added tests for fail states, test for verification postStartup/preShutdown behavior.

JSON tests seems to work now - maybe local issue or fixed by someone.

Found issues:
- Controller doesn't return fail state in case we try to create duplicit container or delete not existing container.
- If all containers are deleted in controller with offline kie server, kie server after connecting to controller assumes that there isn't connection to controller and use local list of containers, see testContainerDisposedAfterStartup

PR #176, opened 2015-08-25
DROOLS-733 - add jms/ejb support ofr timers and executors for jBPM to…

mswiderski:
… properly run on WAS and WLS with background threads


mswiderski:
merged into master and 6.3.x

PR #177, opened 2015-08-26
Enable drools-karaf-itest execution against custom container

jiripetrlik:
This pull request enables drools-karaf-itest to be executed against custom karaf or karaf like container (for example JBoss Fuse). The path to container can be specified using "-Dkaraf.dist.file=path_to_container". Moreover there are few additional options added. It is possible to keep pax exam runtime folder after the test execution is done ("-Dkaraf.keep.runtime.folder"). This feature is very useful in case of debugging. The last additional option "-Dkaraf.maxpermsize" is able to increase the maximal size of perm gen space for karaf container in Java 7.

### DIFF
--- a/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/AbstractKarafIntegrationTest.java
+++ b/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/AbstractKarafIntegrationTest.java
@@ -28,12 +28,19 @@ import java.io.File;
 import java.util.List;
 import java.util.Properties;
 
+import org.ops4j.pax.exam.karaf.options.KarafDistributionBaseConfigurationOption;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.keepRuntimeFolder;
 import static org.ops4j.pax.exam.CoreOptions.*;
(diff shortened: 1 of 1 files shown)


psiroky:
Very nice!

Just to make sure I understand this correctly: nothing basically changes for the default build, when I run `mvn clean install` it will automatically resolve Karaf distro using Maven GAV and run the tests there (so no need to specify the zip file URL anywhere). In case I want to run the test on JBoss Fuse, the only thing needed is the the property definition with the URL. Is this correct?

jiripetrlik:
Yes, it is as you said. The default build will work as before. You must define additional parameters to use different Karaf or Fuse as container.

psiroky:
Merged into master and 6.3.x

cmoulliard:
Very nice PR. Are the options that we can pass as System Properties documented somewhere ?

psiroky:
I don't think so -- only in the code. I guess we could create README file inside the `drools-osgi-itests` module and place that info there? WDYT? I believe these props should not be directly in the project docs, since they are only used for testing.

jiripetrlik:
Thank you for comments. Currently there is no description for these options. I will create README in the drools-osgi-itests as psiroky recommended and send it as new PR.

PR #180, opened 2015-09-02
DROOLS-733 - delay container installation until controller is availab…

mswiderski:
…le whenever controller is configured, ensure bootstrap of kieserver on all server, rename jbpm executor sueue to now clash with workbench in case they share same container

@etirelli take a look - it includes the background connection to controller before installing containers as we discussed


etirelli:
+1 to merge

mswiderski:
merged into master and 6.3.x

PR #182, opened 2015-09-04
DROOLS-733 - added group task test, other improvements

sutaakar:
Added test for task assigned to group.
Fixed issue with callbackConfig - feel free to improve.
Small test improvements and cleanup.


sutaakar:
Working on Wildfly, Tomcat 7 throwed issue when claiming group task with user john - need to check what is wrong there.

PR #185, opened 2015-09-08
DROOLS-733 - added REST api docs for kie server to be auto generated …

mswiderski:
…and included in war file with complete description of all endpoints - based on JRAPIDoc

Not protected resource that exposes all REST api endpoints with all details via the kie server web application. It will always show the actual URL on which it runs and can be reached at:
```
http://host:port/contextpath/docs/
```
for example:
```
http://localhost:8230/kie-server/docs/
```
and this is how it looks like (just first part of it)
![Kie Server REST api]
(http://i.imgur.com/0HZNsbv.png)
Note that all of this is generated from source so no need for any manual manipulation (besides the look and feel of it)

@ibek thanks for great tool and support, comments more than welcome


ibek:
very nice :) ... I like how you changed the look & feel - that you removed the text field and so on - I believe we can merge it as it is.

I have no additional comments since we decided not to use jrapidoc annotations for now.

mswiderski:
@etirelli frankly speaking no idea what TAG is and what sort of approval we need. Although looking at swagger and jrapidoc and compare feature/requirement wise sounds like a good idea.
I have never used swagger and I recall it was withdrawn at some point due to dependencies/license issues but that might have been fixed and is not an issue any way. Has anyone already used swagger?

@ibek thanks, I made it simpler for now. we might decide to add annotation to provide some human readable description on top of these endpoints and methods. What I really liked a lot about this is it can be easily embedded into application and does provide all details such as query parameter names and their types. and in the end was able to make it respond with actual URL the kie server is running on so no need to replace any strings to actually invoke it.

mrietveld:
+1 for merging! I'll probably steal this code for the remote-api. ;)

mswiderski:
@etirelli does it make sense to merge it or at least only the generated part so we have rest api docs for kie server

etirelli:
@mswiderski if it is possible to commit only the static docs, without any dependency to jRAPIDoc at this point, it should be fine. Until we have approval from TAG, we should not add a dependency to it in the pom files.

mswiderski:
merged without jrapidoc generation so no auto updates when rest endpoints changes but at least we do include into build war file the documentation. We will revisit this after TAG approval

PR #186, opened 2015-09-09
BZ-1250323 - Not able to register KIE server if we use https protocol

mswiderski:
extension to KieRemoteHttpRequest to allow use of kie server client over https/ssl when running on localhost.

@mrietveld what do you think? is this a valid addon? it does certainly help when kie server client runs within workbench where users cannot set globally the host name verifier and thus cannot use workbench and kie server on localhost over https.
I believe (though have not tested) use of hostname instead of localhost might be a workaround as well for local machine, but it might be more tricky than it sounds.


mrietveld:
+1 You're welcome to add this to the code. 

FYI: the `KieRemoteHttpRequest` was a workaround because of issues relating to using RestEasy on the client side. I'd like to move us to the current version of apache httpclient as soon as possible, which will probably be 7.0?

mrietveld:
Cherry-picked to 6.3.x @ https://github.com/droolsjbpm/droolsjbpm-integration/commit/48522bc85ba335083b6fb8b8218b8d7c325ca331

PR #188, opened 2015-09-09
DROOLS-733 - fixed NPE when looking for nonexisting process definition

sutaakar:
One small thing found, error returned in testGetProcessDefinitionByContainerAndNonExistingId test has response code 500, I think some 4xx should be better in this case.


sutaakar:
Added NPE fix for nonexisting process instances.

PR #190, opened 2015-09-11
[BZ-1257739] move kie-maven-plugin from drools repo + add deps to ena…

psiroky:
…ble building of all resources OOTB

 * added all the optional deps for user experience. The plugin is now able to build
   every supported resource type

 * using takari libs for plugin testing. See http://takari.io/book/70-testing.html
   for more info

 * removed the "unit" test and using only integration tests as they can be
   spawned in own JVM to eliminate dependency clashes

 * related PR in drools repo: https://github.com/droolsjbpm/drools/pull/495


mariofusco:
Merged with https://github.com/droolsjbpm/droolsjbpm-integration/commit/d5a6b52d506db7eee09e8665e9c661fab952dc80

PR #194, opened 2015-09-17
DROOLS-733 - extended processing time for JMS job executor

sutaakar:
This test can become nondeterministic on slower computers/using slower containers like WLS.
For example running this test on WLS using JMS configuration for kie server client results in durationTime value around 1700ms which is quite close to the limit.
For predicable results I suggest to raise processing time limit to some safer value, like 5000ms.

What do you think?

It is a small change, I suggest to add it to some other commit to avoid branch spamming.


mswiderski:
+1

psiroky:
Merged into master and 6.3.x to fix intermittent failures.

PR #195, opened 2015-09-21
DROOLS-925 - fixed cannot deploy kie-server on wildfly 10.x.x

pokpitch:
DROOLS-925 - fixed cannot deploy kie-server on wildfly 10.x.x due to HornetQ is to be deprecated.

### DIFF
--- a/kie-server-parent/kie-server-wars/kie-server/src/main/shared-ee6-ee7-resources/META-INF/kie-server-jms.xml
+++ b/kie-server-parent/kie-server-wars/kie-server/src/main/shared-ee6-ee7-resources/META-INF/kie-server-jms.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<messaging-deployment xmlns="urn:jboss:messaging-deployment:1.0">
-  <hornetq-server>
+<messaging-deployment xmlns="urn:jboss:messaging-activemq-deployment:1.0">
+  <server>
     <jms-destinations>
 
       <!-- Kie Server REQUEST queue -->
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-wars/kie-server/src/main/shared-ee6-ee7-resources/META-INF/kie-server-jms.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<messaging-deployment xmlns="urn:jboss:messaging-deployment:1.0">
-  <hornetq-server>
+<messaging-deployment xmlns="urn:jboss:messaging-activemq-deployment:1.0">
```
psiroky: Wouldn't his break deployment to older versions of WildFly (8+9)?

pokpitch: Its only effected on WildFly 10+

etirelli:
@pokpitch Not sure if I missed your answer, but as Petr asks above, will this change work for Wildfly 8 and 9 as well?

mswiderski:
-1 this does not work on Wildfly 8.2.0.Final and moreover it modifies file that affects both ee6 and ee7 (meaning both EAP and Wildfly) so will actually break all assemblies from JBoss

error that happened on wildfly 8.2.0.Final
```
Caused by: javax.xml.stream.XMLStreamException: ParseError at [row,col]:[2,1]
Message: Unexpected element '{urn:jboss:messaging-activemq-deployment:1.0}messaging-deployment'
```

PR #197, opened 2015-09-23
DROOLS-733 - update storage when creating container

sutaakar:
Fixes BZ-1265635.


etirelli:
Looks good. Merged. Thanks!

PR #199, opened 2015-09-24
[BZ-1201370] add dependencies used in osgi features to pom file

mariofusco:


psiroky:
@mariofusco would it make sense to add dependencies also on the KIE artifacts, like kie-api, drools-core, etc?

mariofusco:
@psiroky I was not sure. I gave kie dependencies for granted, but also to be conform to other poms, now I'm thinking that yes, they should be there. Can you please take care of adding them?

psiroky:
This PR has been replaced by https://github.com/droolsjbpm/droolsjbpm-integration/pull/201

PR #200, opened 2015-09-24
Karaf version is detected using Maven.

jiripetrlik:
Karaf version is detected using Maven. Previously there was constant 2.4.2 in the code. The user has to specify a version of Karaf container in the case he wants to test against custom Karaf container.

### DIFF
--- a/drools-osgi/drools-karaf-itests/README
+++ b/drools-osgi/drools-karaf-itests/README
@@ -4,6 +4,8 @@ These additional properties can be used to execute tests against custom karaf
 or karaf like container (for example JBoss FUSE). 
 
 karaf.dist.file - path to karaf distribution file (if not defined, the default karaf is downloaded from maven central)
+karafVersion - Version of Karaf container. This parameter is mandatory when a custom Karaf distribution file is specified.
+                (for example Jboss Fuse 6.2. uses Karaf container version 2.4.0)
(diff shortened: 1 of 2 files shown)


drools-osgi/drools-karaf-itests/README
```diff
@@ -4,6 +4,8 @@ These additional properties can be used to execute tests against custom karaf
 or karaf like container (for example JBoss FUSE). 
 
 karaf.dist.file - path to karaf distribution file (if not defined, the default karaf is downloaded from maven central)
+karafVersion - Version of Karaf container. This parameter is mandatory when a custom Karaf distribution file is specified.
```
psiroky: We could maybe use `karaf.version`, so that the name is consistent with other properties. WDYT?

jiripetrlik:
It is because it was possible to set the Karaf version using this property even before my PR. I wanted it to be backward compatible. However, maybe I should change it to karaf.version to be more consistent with other settings. What is your opinion Peter?

psiroky:
I am not sure that property was actually used at all (we were always using the Karaf version defined in the project). I would go with changing the property to `karaf.version`.

jiripetrlik:
Ok, I've renamed that system property.

psiroky:
Merged into master. Let me know if this needs to go into 6.3.x as well (the cherry-pick wasn't clean, so there are probably some other commits missing).

PR #201, opened 2015-09-24
[BZ-1201370] add deps used in OSGi features to POM

psiroky:
@mariofusco I added the KIE deps. I would also like to add simple test which would take deps from the features.xml and the POM and compared those to make sure they are in sync. I will look at that tomorrow.


mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/7320cbf048313f6b4156e2b54d71cfd049f1d24c

PR #202, opened 2015-09-25
[DROOLS-729] add system properties parameter to kie-maven-plugin

psiroky:
* changes originally done by @kedzie, thanks!

@mariofusco please take a look. I believe the system properties will be useful even for non-Android use cases.

### DIFF
--- a/kie-maven-plugin/pom.xml
+++ b/kie-maven-plugin/pom.xml
@@ -28,12 +28,14 @@
             <goals>
               <goal>helpmojo</goal>
             </goals>
+            <phase>process-classes</phase>
           </execution>
           <execution>
             <id>generate-descriptor</id>
             <goals>
               <goal>descriptor</goal>
             </goals>
+            <phase>process-classes</phase>
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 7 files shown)


mrietveld:
Thanks for keeping on top of this, @psiroky!

psiroky:
@kedzie do we need this kind of additional configuration? After talking to @mariofusco we believe the same can be achieved using the `<properties>` directly in the maven pom. Or are there cases where this would not work?

kedzie:
The properties in the maven pom are NOT system properties..

kedzie:
On second thought I don't think they are published by maven as system properties.. but I don't know that for sure.  I could explore that a bit.

psiroky:
@kedzie my apologies, you are right. We can't just use the generic Maven properties.

Looking at the available options to define additional properties for the plugin, I would prefer this way:
```
<configuration>
  <properties>
    <drools.dialect.java.compiler.lnglevel>1.6</drools.dialect.java.compiler.lnglevel>
  </properties>
</configuration>
```
instead of 
```
<configuration>
  <properties>
    <property>
      <name>drools.dialect.java.compiler.lnglevel</name>
      <value>1.6/value>
    </property>
  </properties>
</configuration>
```

psiroky:
I just pushed the changes mentioned in previous comment. @kedzie please let me know if this all right from you point of view. I believe we are ready to merge this.

psiroky:
@mariofusco from my point of view this is ready to be merged.

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/a3ef2b6b156c7bbeffa29c8385a44b41713ebbe3

PR #204, opened 2015-10-06
DROOLS-733 - Extended tests to support Oracle database

sutaakar:
tested Oracle database has some differences compared to H2 used in tests:
- returns null instead of empty String for values which weren't set
- returns results in random order - tests modified for attachments and comments , will prepare another PR for other sorting issues
- Adjusted timeouts as they were too short for remote database


mswiderski:
@sutaakar please update this PR with other sorting issues so I'll merge all at the same time. thanks

sutaakar:
Updated.
Added sorting according to ProcessInstanceId for runtime data.

PR #205, opened 2015-10-14
DROOLS-733 - WebLogic test integration profile

sutaakar:
Changed controller URL as there cannot be 2 servlets with same url-pattern (also aligned to URL used in business central controller).

Please try it.
Ideas/improvements are welcome.

For running WebLogic profile you need to declare just profile and weblogic home, i.e.:
mvn clean install -Poracle-wls-12 -Dweblogic.home=/home/ksuta/wls/wls12130

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-test-war/pom.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-test-war/pom.xml
@@ -73,6 +73,19 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>compile-sources</id>
+            <phase>compile</phase>
+            <goals>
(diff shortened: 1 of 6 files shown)


mrietveld:
Does this only impact the execution-server?

mswiderski:
yes, this is only for kie server

psiroky:
Yes, only kie-server for now. Once we tweak the configs and make sure everything is stable, we will do the same for kie-wb and kie-drools-wb as well.

mrietveld:
Okay, when you tweak the `kie-wb` and `kie-drools-wb`, please (obviously :) ) make sure that both the exec-server and workbench REST services work with the new configuration. Thanks!

psiroky:
@mrietveld just to clarify, we will probably need to tweak the Cargo config. There are no plans to touch the exec-server or workbech REST configs. I am assuming they are already work on WebLogic as they are being tested there.

Main purpose of these Cargo changes is to simplify the (local) testing on WebLogic.

mrietveld:
Ahh.. okay, thanks!

PR #209, opened 2015-10-18
DROOLS-733 - enhancements for executor tests

sutaakar:
Making tests more reliable and cleaner.

Raised timeouts to make tests work with slow connection on DBs.

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
@@ -39,6 +39,7 @@ public class KieServerConstants {
     public static final String CFG_PERSISTANCE_DS = "org.kie.server.persistence.ds";
     public static final String CFG_PERSISTANCE_TM = "org.kie.server.persistence.tm";
     public static final String CFG_PERSISTANCE_DIALECT = "org.kie.server.persistence.dialect";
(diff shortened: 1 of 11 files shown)


sutaakar:
Don't merge yet, will update PR.

sutaakar:
Additional changes:
- brought new constant allowing configuration of hibernate.default_schema (needed for DB2 as this database tend to use random schemas if this is not configured)
- Changes to make failed assertions/errors more descriptive.
- Removed controller URL setting for kie server in pom - startup interaction between kie server and controller is tested using embedded kie server which has separate configuration, mixing embedded kie server with deployed one could make tests randomly fail
- try/catch block in KieControllerStartupIntegrationTest - Jenkins builds tends to randomly fail in that place, I believe it is caused by way how Java translate localhost address to IP, this may be fixed by correct configuration of Jenkins. Best solution would be to get rid of embedded server in this case and use same container as for controller for this test.
- Changed assertion of timestamp to just notNull as some databases(mssql or mysql, not sure now) rounds timestamps to whole seconds in default configuration

mrietveld:
None of the failing tests are related to this PR. 

Drools-simulator tests fixed by: https://github.com/droolsjbpm/drools/pull/519
kie-remote tests will be fixed by a PR created by me (soon.. :) ).

sutaakar:
Questions, comments and improvements welcome.

PR #212, opened 2015-10-21
BZ-1273109 - BPM cluster: Job executor fails with NPE while

mswiderski:
this PR enables efficient and reliable job status checks within cluster. Since deployment and undeployment commands/jobs are executed asynchronously (jbpm executor) they might be executed on any cluster member. Thus when running clustered environment default map based cache did not work properly when request for checking job status was sent to other instance then the job was actually executed.
This PR uses executorService (component that actually executed the job) to find it in db in case it does not exist in the local cache. That way job status check is cluster aware and will cover efficient job execution distribution across cluster members while still allowing to find the job status regardless where it was executed. 
Especially important when kie-wb instances are running behind load balancer.

this is one of three PRs for this BZ - others are in guvnor (https://github.com/droolsjbpm/guvnor/pull/167) and jbpm (https://github.com/droolsjbpm/jbpm/pull/317), though they don't have any compile dependency

### DIFF
--- a/kie-remote/kie-remote-jaxb/src/main/java/org/kie/services/client/serialization/jaxb/impl/deploy/JaxbDeploymentJobResult.java
+++ b/kie-remote/kie-remote-jaxb/src/main/java/org/kie/services/client/serialization/jaxb/impl/deploy/JaxbDeploymentJobResult.java
@@ -15,6 +15,7 @@
 
 package org.kie.services.client.serialization.jaxb.impl.deploy;
 
+import java.io.Serializable;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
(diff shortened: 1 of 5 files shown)


kie-remote/kie-remote-jaxb/src/main/java/org/kie/services/client/serialization/jaxb/impl/deploy/JaxbDeploymentJobResult.java
```diff
@@ -27,7 +28,7 @@
 @XmlRootElement(name="deployment-job-result")
 @XmlAccessorType(XmlAccessType.FIELD)
 @JsonIgnoreProperties({"jobId"})
-public class JaxbDeploymentJobResult {
+public class JaxbDeploymentJobResult implements Serializable {
```
mrietveld: Why is `Serializable` needed?

mswiderski: because it is variable/data for executor and thus needs to be serialized and stored as part of the job to be loaded on execution time - async way

mrietveld:
+1 feel free to merge.

PR #214, opened 2015-10-23
DROOLS-733 - enabled auto docs genration for KIE Server via jrapidoc

mswiderski:
@etirelli as discussed here is jrapidoc enabled to auto generate KIE Server docs


etirelli:
@mswiderski Great! thanks! 

+1 to merge.

PR #217, opened 2015-10-30
- fix webservices exception ("getReader() has already been called for this request")

mrietveld:
Fixes the failing webservice test for the smoke tests. 

It turns out it's practically impossible to open the stream (and parse it for the deployment id) and then "reset" the stream so that the REST *and* WS frameworks can (re)open the stream in their code.


mrietveld:
eh.. hold off on merging, double checking something.

mswiderski:
is this still under investigation or?

mrietveld:
@mswiderski Please merge. I've confirmed that this fixes problems on master. (via `kie-tests` integration tests).

PR #220, opened 2015-11-03
[BZ-1276379] fix kie module files discovery when dependencies are installed in JBoss EAP Module

mariofusco:


mswiderski:
only comment from my end is that I still would like to see the setting of application context via method setApplicationContext, in the end this is the purpose of this method :)

looks like rely on spring configuration files is the safest bet as they should always be within the application (war file)
+1

mariofusco:
@mswiderski Agreed. I'll add it.

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/85c894f10920bd825637bf8ad8225a6ecd2c9b09

PR #224, opened 2015-11-09
JBPM-4678 - Constructor of RemoteConfiguration does not process Url

mrietveld:
https://issues.jboss.org/browse/JBPM-4678

### DIFF
--- a/kie-remote/kie-remote-client/src/main/java/org/kie/services/client/api/command/RemoteConfiguration.java
+++ b/kie-remote/kie-remote-client/src/main/java/org/kie/services/client/api/command/RemoteConfiguration.java
@@ -3,7 +3,7 @@
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * 
+ *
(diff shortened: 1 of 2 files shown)


mswiderski:
the change look ok, just one question - as the jira says it's not processing the URL in constructor wouldn't that mean it requests to add the processing of URL in the constructor rather than removing it from there?

Not knowing the internals there feel free to skip this comment, just thought that URL for remote configuration is something common to be used anon default value e.g. localhost

mrietveld:
See https://issues.jboss.org/browse/JBPM-4822

Users should not be touching the `RemoteConfiguration` class: it's _very_ internal, and fooling around with the configuration of this class (or other internal classes in `kie-remote-client`) can lead to ugly problems. 

The fact that the `url` parameter was not processed in the constructor is intentional, because the url can either be for the REST or the WS client, and should be set differently depending on which one is being targetd. The solution here was to make sure that it's simply not used in the configuration.

mswiderski:
excellent, thanks for explanation Marco! will merge in few minutes

PR #225, opened 2015-11-10
BZ-1278722 - Marshalling Error when executing command through SOAP interface

mrietveld:
Because I have the unique (and lovely! /8( ) situation of having a client whose classpath must be different than the server side, easy integration tests for webservice (and REST) issues are not easy without a _lot_ of work (think a custom classloader framework, just for the test!). 

The test for this is available in https://github.com/droolsjbpm/kie-tests/pull/14

However, the fix in this PR is another hack. It works, but it's another band-aid/strand of spaghetti. For this reason, please do _not_ cherry-pick this to master. 

In master, classes will slowly be centralized in either `kie-core-jaxb` (one of the new `kie-core` modules, resulting from @mariofusco 's work to separate drools and jbpm from eachother) or in `kie-internal`. I will submit a PR for that fix later this week.

### DIFF
--- a/kie-remote/kie-remote-client/pom.xml
+++ b/kie-remote/kie-remote-client/pom.xml
@@ -285,6 +285,7 @@
               org.jsoup.*;resolution:=optional,
               *
             </Import-Package>
+            <Include-Resource>wsdl/=target/classes/wsdl</Include-Resource>
           </instructions>
         </configuration>
       </plugin>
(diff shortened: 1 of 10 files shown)


mrietveld:
@mswiderski Please merge this ASAP! It needs to be in by EOD.

mswiderski:
as soon as the build is done with success I'll merge

mswiderski:
merged into 6.3.x with https://github.com/droolsjbpm/droolsjbpm-integration/commit/0b4b05ca4f5439290bdf5145d2661c6c9810ea09

PR #226, opened 2015-11-10
BZ-1273109 - BPM cluster: Job executor fails with NPE while creating a deployment unit

mswiderski:
This pr comes with two commits
- first to fix the problem of possible NPE when deployment job is executed on another node than the request came it - should check both input and output of the job for completeness and not only the output
- second is to remove most recent jobs cache as it might serve outdated information as the cache is only maintained in memory. so when there were two jobs - first to deploy and second to undeploy and these jobs were performed on cluster nodes the cache for most recent jobs will be outdated on one of the nodes and thus providing misleading response. This happens especially when cluster nodes have not sync yet

@mrietveld could you please have a look?

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/DeployResourceBase.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/DeployResourceBase.java
@@ -186,11 +186,11 @@ public class DeployResourceBase extends ResourceBase {
         }
         
         // Most recent job? 
-        JaxbDeploymentJobResult jobResult = jobResultMgr.getMostRecentJob(deploymentId);
-        if( jobResult != null ) { 
-            jaxbDepUnit = jobResult.getDeploymentUnit();
-            return jaxbDepUnit;
-        }
+//        JaxbDeploymentJobResult jobResult = jobResultMgr.getMostRecentJob(deploymentId);
+//        if( jobResult != null ) {
+//            jaxbDepUnit = jobResult.getDeploymentUnit();
+//            return jaxbDepUnit;
+//        }
         
         // Nonexistent? 
         String [] gavKK = deploymentId.split(":");
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/DeployResourceBase.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/DeployResourceBase.java
@@ -184,15 +184,7 @@ public class DeployResourceBase extends ResourceBase {
                 return jaxbDepUnit;
             } 
         }
-        
-        // Most recent job? 
-//        JaxbDeploymentJobResult jobResult = jobResultMgr.getMostRecentJob(deploymentId);
-//        if( jobResult != null ) {
-//            jaxbDepUnit = jobResult.getDeploymentUnit();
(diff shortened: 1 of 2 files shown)


kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/DeployResourceBase.java
```diff
@@ -207,7 +199,7 @@ public JaxbDeploymentUnit determineStatus(String deploymentId, boolean checkDepl
         default:
             throw KieRemoteRestOperationException.notFound("Invalid deployment id: " + deploymentId);
         }
-        jaxbDepUnit.setStatus(JaxbDeploymentStatus.NONEXISTENT);
+        jaxbDepUnit.setStatus(JaxbDeploymentStatus.UNDEPLOYED);
```
mrietveld: +1 Good idea!

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/async/JobResultManager.java
```diff
@@ -162,19 +131,60 @@ public JaxbDeploymentJobResult getJob(String jobId) {
...
+        }
+
+        if (ctx != null && ctx.getData(itemName) != null) {
+            return ctx.getData(itemName);
         }
+
         return null;
     }
 
+    protected Object getItemFromRequestOutput(String itemName, RequestInfo requestInfo) {
```
mrietveld: +1 for putting this in a method, thanks!

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/async/JobResultManager.java
```diff
@@ -162,19 +131,60 @@ public JaxbDeploymentJobResult getJob(String jobId) {
...
-     * @param deploymentId The id of the deployment
-     * @return The {@link JaxbDeploymentJobResult} with the job information
-     */
-    public JaxbDeploymentJobResult getMostRecentJob(String deploymentId) {
-        logger.debug( "Getting most recent job for '{}'", deploymentId);
-        String jobId = deploymentIdMostRecentJobIdMap.get(deploymentId);
-        if( jobId != null ) { 
-            return getJob(jobId);
+
+    protected Object getItemFromRequestInput(String itemName, RequestInfo requestInfo) {
```
mrietveld: This is just nitpicking, feel free to ignore this: 

I'd much rather see 1 method that has a class/type-based if clause at the end: 

    if( result instanceof CommandContext ) { 
      return ((CommandContext) result).getData(itemName);
    } else if ( result instanceof ExecutionResults ) { 
      return ((ExecutionResults) result).getData(itemName);
    }

    return null;

There's no reason to check whether or not `.getData(itemName)` is null or not, seeing as we're returning null no matter what. :)

mrietveld:
Feel free to merge. 

If you have the time and are inspired, please do add a test. However, I understand that this is a piece of logic that's hard to test, so no problems if you do not.

mrietveld:
Also, of course, please rebase into 1 commit. :)

mswiderski:
commits squashed and merged to both master and 6.3.x. Haven't changed the methods to move to single type based clause at the moment. thanks for review @mrietveld !

PR #228, opened 2015-11-12
Upgrade takari-plugin-testing to 2.8.0 to fix failing tests

psiroky:
* tests were failing in case of using custom settings.xml with
   additional repositories. The tooling in version 2.1.0 would
   not pass the content of settings.xml to the project under test
   which resulted in errors due to missing dependencies.

 * the versions will be moved to build-bootstrap/ip-bom

@mbiarnes this fixes the failing tests we saw in the latest two "prod" builds.

### DIFF
--- a/kie-maven-plugin/pom.xml
+++ b/kie-maven-plugin/pom.xml
@@ -97,12 +97,12 @@
       <dependency>
         <groupId>io.takari.maven.plugins</groupId>
         <artifactId>takari-plugin-testing</artifactId>
-        <version>2.1.0</version>
+        <version>2.8.0</version>
       </dependency>
       <dependency>
         <groupId>io.takari.maven.plugins</groupId>
         <artifactId>takari-plugin-integration-testing</artifactId>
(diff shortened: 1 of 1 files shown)


psiroky:
I am looking into the test failures as they are caused by the takari stuff version upgrade.

psiroky:
The problem was that the Jenkins slaves did not have a default settings.xml file under ~/.m2. I added empty ones to fix the issue for now. In ideal case the takari tooling should work even if the file is not there, I created https://github.com/takari/takari-plugin-testing-project/issues/11 for that.

mbiarnes:
+1 - test run now

psiroky:
Merged.
 
master: https://github.com/droolsjbpm/droolsjbpm-integration/commit/f67dc98aa83effed05037724d20e4144963ac73b
6.3.x: https://github.com/droolsjbpm/droolsjbpm-integration/commit/24dd6580f979c1d4013e4ec44016c93fc5e1ffd2 (wrong 'closes' comment there)

PR #236, opened 2015-11-23
BZ-1280313 - Multiple client command execution through SOAP interface…

mswiderski:
… does not authenticate correctly

this is intended to share the finding and might not be complete solution. @mrietveld please have a look

this is caused by http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6626700 and that's why such a dirty solution...

clearHttpUrlConnectionAuthCache might require synchronized keyword to make sure we don't override the cache with race condition.

### DIFF
--- a/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/ws/KieRemoteWsAuthenticator.java
+++ b/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/ws/KieRemoteWsAuthenticator.java
@@ -15,8 +15,10 @@
 
 package org.kie.remote.client.ws;
 
+import java.lang.reflect.Method;
 import java.net.Authenticator;
 import java.net.PasswordAuthentication;
+import java.util.HashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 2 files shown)


triceo:
Is there any test coverage that would detect problems in this area early on?

I worry that, since we're in ```sun.*``` territory, any of these classes might disappear from any of the supported JDKs without notice, and so it'd be great if we had an easy way to catch that.

mswiderski:
agreed, we should have that covered in community tests as well (as it's already covered in qe tests). I believe that will be added to kie-tests repo but will leave that to @mrietveld as he knows it better

mrietveld:
@mswiderski Updated. Would you please look through this? 

@triceo I have tests in kie-tests and will also add it to kie-wb-smoke-tests. I also have a test in the PR, but it's a "always" positive at the moment. 

@mswiderski Any ideas on making sure that the AuthenticatorTest does what it should do? (Test in PR).

mrietveld:
oops, wrong PR. Please see https://github.com/droolsjbpm/droolsjbpm-integration/pull/237

mrietveld:
Please close this PR (and do _not_ merge it.).

PR #237, opened 2015-11-24
BZ-1280313 - Multiple client command execution through SOAP interface does not authenticate correctly

mrietveld:


mrietveld:
@mswiderski Updated. Would you please look through this?

@triceo I have tests in kie-tests and will also add it to kie-wb-smoke-tests. I also have a test in the PR, but it's a "always" positive at the moment.

@mswiderski Any ideas on making sure that the AuthenticatorTest does what it should do? (Test in PR).

mswiderski:
Marco, the test works as expected because only first client creation will load wsdl from the server (and this is where it fails for the first bad credentials). Next stemple will take wsdl fron cxf cache of wsdl definitions. So last creation of the client is not reaching the server and thus passes. If you call any command on this client it then should fail. 

Will take closer look at the changes when in fron of the computer

mrietveld:
Maciej, the test passes even when the changes have not been applied (or are commented out).

mswiderski:
you mean entire test or just the last bits?

mrietveld:
This test: https://github.com/mrietveld/droolsjbpm-integration/blob/pr-BZ-1280313/kie-remote/kie-remote-client/src/test/java/org/kie/remote/client/ws/AuthenticatorTest.java

mswiderski:
I believe this is expected as this line fails at loading wsdl as it uses wrong credentials, then as soon as this line [2] is invoked which succeeds from that time it will cache good credentials and reuse them for all subsequent calls.

[1] https://github.com/mrietveld/droolsjbpm-integration/blob/pr-BZ-1280313/kie-remote/kie-remote-client/src/test/java/org/kie/remote/client/ws/AuthenticatorTest.java#L144
[2] https://github.com/mrietveld/droolsjbpm-integration/blob/pr-BZ-1280313/kie-remote/kie-remote-client/src/test/java/org/kie/remote/client/ws/AuthenticatorTest.java#L152

mrietveld:
Thanks.

kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/jaxb/ClientJaxbSerializationProvider.java
```diff
@@ -112,6 +112,19 @@ public static JaxbSerializationProvider newInstance(Collection<Class<?>> extraJa
...
+            classes = new ArrayList<Class<?>>(extraJaxbClass.length);
+            for( Class jaxbClass : extraJaxbClass ) {
+                classes.add(jaxbClass);
+            }
+        }
+        ClientJaxbSerializationProvider jaxbSerProvider = new ClientJaxbSerializationProvider(classes);
+        jaxbSerProvider.initialize();
+        return jaxbSerProvider;
+    }
+
```
mswiderski: is this related to the soap client issue?

kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/ws/KieRemoteWsAuthenticator.java
```diff
@@ -11,54 +11,99 @@
...
-    
+
     @Override
-    protected PasswordAuthentication getPasswordAuthentication() { 
+    protected PasswordAuthentication getPasswordAuthentication() {
         return threadLocalPasswordAuthentication.get();
     }
-    
+
+    private static AtomicBoolean authenticatorSet = new AtomicBoolean(false);
```
mswiderski: might be worth renaming the variable as it's not used for authenticator being set anymore, though a minor thing

kie-remote/kie-remote-client/src/test/java/org/kie/remote/client/ws/AuthenticatorTest.java
```diff
@@ -0,0 +1,352 @@
...
+
+        String user = ((JaxbStringListResponse) cmdResp).getResult().get(0);
+        String password = ((JaxbStringListResponse) cmdResp).getResult().get(1);
+        assertEquals("User did not match request!", GOOD_USER, user);
+        assertEquals("Password did not match request!", GOOD_PASSWORD, password);
+
+        // bad client creation
+        exceptionThrown = false;
+        try {
+            createClient(BAD_USER, BAD_PASSWORD);
```
mswiderski: let's call a command here via execute method and that should make the authentication fail with the fix and succeed without the fix

mswiderski:
besides these minor comment all looks good to me

PR #239, opened 2015-11-25
kie-remote-client: use explicit dependency to fix parallel builds

psiroky:
@mrietveld could you please double check this? I believe it is safe as we are using the same on master and 6.3.x for some time now.


mswiderski:
+1 from me, thanks @psiroky

mrietveld:
@mswiderski Thanks! yes, +1

PR #241, opened 2015-11-27
kie-remote-client: 'better' fix for parallel builds

psiroky:
@mswiderski this what we were talking about yesterday. The same is on 6.2.x already.

### DIFF
--- a/kie-remote/kie-remote-client/pom.xml
+++ b/kie-remote/kie-remote-client/pom.xml
@@ -151,8 +151,9 @@
       <groupId>org.kie.remote</groupId>
       <artifactId>kie-remote-jaxb-gen</artifactId>
       <classifier>sources</classifier>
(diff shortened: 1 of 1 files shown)


mswiderski:
+1

mrietveld:
@psiroky @mswiderski Just to be clear: kie-remote-jaxb-gen should ****NEVER**** be exposed to users or anyone. It is an INTERNAL hack and should NOT be available via nexus.

psiroky:
@mrietveld I wasn't aware of that. It is currently build and deployed as any other module, so you will find it in nexus.

mswiderski:
@psiroky shall this go to 6.3.x considering we are in blockers only mode?

psiroky:
It does not change the code itself and is only build-related. So I would vote for yes.

mswiderski:
merged, thanks

PR #243, opened 2015-12-02
DROOLS-733 - added profile for smoke tests

sutaakar:
Also reworked maven repository initialization to use minimal pom file without any dependency.


mswiderski:
+1

PR #244, opened 2015-12-03
DROOLS-733 - added profile for smoke tests

sutaakar:
Is useful to have this available also for 6.3.


mswiderski:
+1 from me though I'd like to get confirmation from @etirelli and @krisv if it is ok to merge it to 6.3.x. Note this only affects tests and not the actual code

etirelli:
The changes look fine. +1 to merge. The only thing is: we are done with DROOLS-733, so it would be good to open a new ticket for the commits from now on. It can be a umbrella top level ticket for 6.4 and then we can create subtasks as necessary for everything else.

mswiderski:
merged into 6.3.x

PR #245, opened 2015-12-03
DROOLS-994 - Improve extensibility of drools extension in KIE Server

mswiderski:
@etirelli could you have a look - it's only refactoring to allow better extensibility as described in the [articles](http://mswiderski.blogspot.com/2015/12/extending-kie-server-capabilities.html) illustrating how kie server can be extended on various levels


etirelli:
Looks good. +1

PR #246, opened 2015-12-03
JBPM-4867 - fixed value for disabling executor service

sutaakar:


mswiderski:
looks good to me, let's wait for jenkins and if all good then merge it

PR #247, opened 2015-12-03
[DROOLS-999] remove usage of mockito-all uberjar

psiroky:
* all touched modules already contain mockito-core,
   which is sufficient


mrietveld:
+1

PR #249, opened 2015-12-04
Added deployment descriptor test

sutaakar:
So far just checking global variable processing from deployment descriptor.


mswiderski:
+1 will merge soon

PR #252, opened 2015-12-08
Per Request strategy deployment descriptor test

sutaakar:
Verification done by checking of stored object by new request.

@mswiderski Do you think it is good test for checking such thing?

Also found one issue during implementation of this test - JSON marshaller doesn't correctly unmarshall ArrayList containing custom class objects - to simulate it run new test.


mswiderski:
it does, based on this we verify that different ksession is used

I'll fix the json marshaller issue and merge this, thanks

mswiderski:
provided new PR that includes this commit with fix for marshaller: https://github.com/droolsjbpm/droolsjbpm-integration/pull/253 thus closing this one as once merged it will include this work as well, thanks @sutaakar

PR #254, opened 2015-12-14
JBPM-4877 - fixes CNFE during smart JAXB context initialization

mrietveld:
This fixes the CNFE that @agiertli describes in his support mail about interfaces and JAXB.

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
@@ -39,7 +39,9 @@ import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.Validator;
 
 import org.jbpm.kie.services.api.DeploymentIdResolver;
+import org.jbpm.kie.services.impl.KModuleDeploymentUnit;
 import org.jbpm.services.api.DeploymentEvent;
+import org.jbpm.services.api.model.DeploymentUnit;
 import org.jbpm.services.cdi.Deploy;
 import org.jbpm.services.cdi.Undeploy;
 import org.kie.remote.services.exception.KieRemoteServicesDeploymentException;
@@ -137,7 +139,10 @@ public class DynamicJaxbContext extends JAXBContext {
     public void addOnDeploy(@Observes @Deploy DeploymentEvent event) {
         String deploymentId = event.getDeploymentId();
         deploymentClassesMap.put(deploymentId, event.getDeployedUnit().getDeployedClasses());
-        setupDeploymentJaxbContext(deploymentId);
+        DeploymentUnit deploymentUnit = event.getDeployedUnit().getDeploymentUnit();
+        ClassLoader projectClassLoader = ((KModuleDeploymentUnit) deploymentUnit).getKieContainer().getClassLoader();
+
+        setupDeploymentJaxbContext(deploymentId, projectClassLoader);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mswiderski:
can we have a jira for this and test if possible

mrietveld:
A test is practically impossible, making a jira now.

mrietveld:
jira created: https://issues.jboss.org/browse/JBPM-4877

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
```diff
@@ -338,7 +343,7 @@ private void removeClassFromJaxbContextClassList( List<Class> classList, Illegal
                String className = classLocs.listIterator(classLocs.size()).previous().toString();
                Class removeClass = null;
                try {
-                   removeClass = Class.forName(className);
+                   removeClass = projectClassLoader.loadClass(className);
```
mswiderski: out of curiosity as I don't know it myself - what is better to be used:
classLoader.loadClass(className) or Class.forName(className, init, classLoader)?

mrietveld: Personally, I prefer `ClassLoader.loadClass(className)` simply because I know exactly what is going on there: the `ClassLoader.loadClass(className)` method is directly calling the Project ClassLoader. 

I'm pretty sure that the class does not have to be initialized (seeing as the project classloader has already loaded it). 

Class.forName(clasName, init, classLoader) is doing security checks which we don't really need.

mswiderski: roger that, thanks

+1 will just wait for jenkins build to complete

etirelli: @mrietveld @mswiderski the main difference is really that Class.forName(...) initializes the loaded class. We did have problems a couple years back because we used ClassLoader.loadClass(...) and we had to replace them by Class.forName(...), but I don't recall the actual scenario. I default to pretty much always use Class.forName(...) ever since, but @mrietveld seems to have a reason not to? Here are a couple articles about it:
http://www.javaworld.com/article/2077332/core-java/get-a-load-of-that-name.html
http://stackoverflow.com/questions/8100376/class-forname-vs-classloader-loadclass-which-to-use-for-dynamic-loading

etirelli: Just to clarify, the problems we had were on drools-core/drools-compiler.

mrietveld:
@etirelli, @mswiderski 

Changed to use `Class.forName(String, boolean, ClassLoader)`, just to be safe. 

I'm pretty sure that the problems that Edson is describing were happening before we had the current Project Classloader. 

Looking up a github link to the code that I think fixes that..

mrietveld:
Oops, I was wrong! Good catch, @mswiderski, @etirelli 

https://github.com/droolsjbpm/drools/blob/master/drools-core/src/main/java/org/drools/core/common/ProjectClassLoader.java

It provides `loadClass(String, boolean)`, but does not override `loadClass(String)`. 

The default `loadClass(String)` method calls `loadClass(String, false)`.

mswiderski:
@mrietveld looks like there are some test failures:
```
java.lang.NullPointerException: null
	at org.kie.remote.services.rest.jaxb.DynamicJaxbContext.addOnDeploy(DynamicJaxbContext.java:143)
	at org.kie.remote.services.rest.jaxb.JaxbContextResolverTest.deploy(JaxbContextResolverTest.java:105)
	at org.kie.remote.services.rest.jaxb.JaxbContextResolverTest.cacheJaxbContextTest(JaxbContextResolverTest.java:194)
```
though jenkins just started another build, most likely after last update of the code...

mrietveld:
jenkins, please retest

mrietveld:
@mswiderski yup, test framework needed to be updated. Rebuilding now.

PR #259, opened 2016-01-06
kie-server: added withVars query attribute to getProcessInstanceById …

ibek:
…operation

KIE Server already provides getProcessInstance operation with query attribute withVars but only for an active process instance [1]. This PR adds this functionality to RuntimeDataResource to be able to query completed and aborted processes including their variables.

[1] https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/ProcessResource.java#L218


mswiderski:
Few things:
- no test for this new capability
- no client method to use this endpoint
- might need some impl to support it over jms
- no jira in commit message

@ibek do you mind working on these?

ibek:
Hi Maciej, yes, sorry about that... Right now I will file JIRA and I will get back to it in February.

ibek:
Here is the JIRA https://issues.jboss.org/browse/DROOLS-1020

mswiderski:
can we then close this PR and submit new one with updates as it's not going to be done this month?

mrietveld:
@ibek FYI: the current kie-wb remote-api will be migrated to the exec-server. 

If you're adding functionality that's identical, it would be great if you could add it under the same URL?

PR #260, opened 2016-01-07
BZ-1296017:  "IllegalAnnotationsException JAXB can t handle interface…

jesuino:
BZ-1296017:  IllegalAnnotationsException "JAXB can't handle interfaces." in kie-server


mswiderski:
Based on this I have extended verification of extra classes to handle not only interfaces but other types that should not be added + added test case to make sure we cover it. See PR https://github.com/droolsjbpm/droolsjbpm-integration/pull/261
thanks @jesuino

PR #264, opened 2016-01-12
DROOLS-1022 - WebSphere: unable to deploy Kie server war

mswiderski:


mrietveld:
The list of tests failing (`org.kie.server.integrationtests.jbpm.ImageServiceIntegrationTest.test`*) seems to be related to this change? 

Once the tests have been fixed, +1.

mswiderski:
correct, this was caused by the changed in this PR. Updated PR to use external xercesImpl:
- EAP and Wildfly it's taken from the modules
- Tomcat has included in WEB-INF/lib as it has it's own (webc) assembly
- WebSphere does not need it as app server delivers it
- WebLogic needs to have added it manually to DOMAIN_HOME/lib this is standard weblogic way to add application libs to server - requires to be documented

So this is again dependency hell when it comes to supporting multiple app servers with single war file... any other ideas are more than welcome.

mswiderski:
by it I meant xercesImpl-2.9.1.jar - regarding weblogic libs

triceo:
@rsynek I think you'll find Maciej's information interesting for our automation.

PR #265, opened 2016-01-19
DROOLS-719 - added Cargo WAS profile for Kie server, WLS profile fix

sutaakar:
Usage: mvn clean install -Pwas85x -Dwebsphere.home="websphere home folder"
Tested with JDK 1.7 and WAS 8.5.5.7.

Side note: WebSphere Cargo plugin is really slow, can take several minutes to initialize and configure new profile (delete old profile, create new profile, configure profile, start WebSphere, add users and groups). So if initialization looks like stuck it may be caused by this. 
WebSphere is run as separate process so if something gets wrong and WebSphere isn't turn off with maven then it has to be shut down manually.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
@@ -50,6 +50,7 @@
 
         <include>com.sun.xml.bind:jaxb-impl</include>
         <include>com.sun.xml.bind:jaxb-core</include>
+        <include>org.slf4j:slf4j-api</include>
       </includes>
       <outputDirectory>WEB-INF/lib</outputDirectory>
       <useTransitiveFiltering>true</useTransitiveFiltering>
--- a/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/java/org/kie/server/controller/common/KieServerControllerApplication.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/java/org/kie/server/controller/common/KieServerControllerApplication.java
@@ -15,9 +15,23 @@
 
 package org.kie.server.controller.common;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
 
+import org.kie.server.controller.rest.RestKieServerControllerAdminImpl;
+import org.kie.server.controller.rest.RestKieServerControllerImpl;
+
 @ApplicationPath("/")
 public class KieServerControllerApplication extends Application {
+
+    @Override
+    public Set<Class<?>> getClasses() {
+        Set<Class<?>> classes = new HashSet<Class<?>>();
+        classes.add(RestKieServerControllerImpl.class);
+        classes.add(RestKieServerControllerAdminImpl.class);
+        return classes;
+    }
 }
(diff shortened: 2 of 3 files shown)


kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
```diff
@@ -50,6 +50,7 @@
 
         <include>com.sun.xml.bind:jaxb-impl</include>
         <include>com.sun.xml.bind:jaxb-core</include>
+        <include>org.slf4j:slf4j-jdk14</include>
```
psiroky: I am not 100% sure, but this could cause troubles on EAP 6 where we want the container's jboss-logmanager to be used. @sutaakar would you be able able to test this?

I know it is basically impossible to get this completely right since we use the same WAR for multiple containers...

psiroky:
Just a minor question/concern about the slf4j-jdk14 (see inline). Otherwise looks very good, thanks @sutaakar!

kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
sutaakar: I will try to check it.

mswiderski: why we need to add it here? Didn't logging work well on all containers with current libraries included?

mswiderski:
could we use new jira for this work instead of reusing long time resolved jira issue...??

sutaakar:
Well, we can.
This issue [1] is about enabling of automated testing of Kie server on WebSphere, it is still open, so I found appropriate to use it.


[1] https://issues.jboss.org/browse/DROOLS-719

psiroky:
I think https://issues.jboss.org/browse/DROOLS-719 is exactly the one use here (and it is not yet resolved).

mswiderski:
my bad, opened wrong jira issue... frankly speaking no idea how... then there is no need to change the jira 
+1 to merge

kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
sutaakar: Without this dependency WebSphere fails to deploy RestKieServerControllerImpl with:
 Caused by: java.lang.ClassNotFoundException: org.slf4j.Logger

mswiderski: shouldn't slf4j-api be enough in that case? instead of given (jdk) impl of slf4j...

sutaakar: Can be, will check it.
It makes me think... this dependency is used also in Kie server ee6 WAR. Is there some special reason why it is used there or slf4j-api should be more appropriate there too?

psiroky: It should be generally enough to include just the slf4j-api. The logging impl should be then supplied by the container. 

Although I am wondering if there isn't something weird going on here, because slf4j-jdk14 does _not_ contain the org.slf4j.Logger class, that one is indeed in slf4j-api.

sutaakar: slf4j-api is dependency of slf4j-jdk14 so this is how it was included.
Using slf4j-api makes WebSphere work, will update PR to use it.

PR #267, opened 2016-01-19
DROOLS-1032 - WebSphere JMS on Kie server produces errors in log

mswiderski:
fixed handling WebSphere (specific) authentication when JMS is used

@sutaakar could you please take a look at this?


sutaakar:
Looks good to me. JMS authentication works with this commit.

PR #275, opened 2016-01-25
DROOLS-1040 - Kie server on Tomcat with JACCValve fails authorisation

sutaakar:


mswiderski:
+1 will merge later on. @sutaakar shall we have this on 6.4.x as well? If you don't mind (and agree wishing having it on 6.4.x ) submitting another PR for 6.4 that would be great

thanks a lot for this @sutaakar

sutaakar:
sure, will prepare PR for 6.4

PR #277, opened 2016-01-25
DROOLS-1040 - Kie server on Tomcat with JACCValve fails authorisation

sutaakar:


mswiderski:
merged into 6.4.x thanks @sutaakar

PR #280, opened 2016-01-27
JBPM-4900 - Enhance TaskList Search capability - query service based …

mswiderski:
…on dashbuilder data sets

depends on https://github.com/droolsjbpm/jbpm/pull/387

more description on this work can be found here:
http://mswiderski.blogspot.com/2016/01/advanced-queries-in-kie-server.html


mswiderski:
@psiroky @mrietveld didn't find better way to refer to KIE dependencies from test project to be reliable than what is in this PR: https://github.com/droolsjbpm/droolsjbpm-integration/pull/280/files#diff-a9ab380a14ce9202cae1c82c8fe65427R22

if using regular scope (like provided) it failed on downloading snapshots for every day builds so it was certainly not the right option. If you can think of better way to secure it I am more than happy to change :)

will see if this even work on jenkins... first verification on it's way

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/filtered-resources/kjars-sources/query-definition-project/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+  <version>1.0.0.Final</version>
+  <name>query-definition-project</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-services-api</artifactId>
+      <version>${version.org.kie}</version>
+      <scope>system</scope>
+      <systemPath>${user.home}/.m2/repository/org/jbpm/jbpm-services-api/${version.org.kie}/jbpm-services-api-${version.org.kie}.jar</systemPath>
```
psiroky: This will likely fail when the local maven repository is somewhere else that ~/.m2 (this is true for the Jenkins PR builds). I will think about how we could do this better (no good ideas off the top of my head).

mswiderski:
@psiroky fixed by using settings.localRepository property: https://github.com/droolsjbpm/droolsjbpm-integration/pull/280/files#diff-a9ab380a14ce9202cae1c82c8fe65427R16 build is green. 

Now it's based on properties so the paths can be overridden if needed too. If you have any better idea I am all open :)

psiroky:
@mswiderski thanks for info. I will take a look early next week. I would like to avoid the system-scoped deps, but I am currently not aware of any better alternative.

mswiderski:
any objections to merge this?

psiroky:
@mswiderski sorry, I still haven't had a chance to look at that maven system scoped deps. Could you give one more day? If I don't find anything better, we would go with the current solution (with possibility of  improving sometime later).

mswiderski:
@psiroky sure thing, I hold it back till tomorrow

psiroky:
Go ahead with the merge, should be fine for now. I will let you if I come up with something better in the future.

PR #282, opened 2016-02-01
DROOLS-1034 - Kie server test improvement for Oracle RAC

jakubschwan:
fix failing test

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
@@ -900,8 +900,9 @@ public class RuntimeDataServiceIntegrationTest extends JbpmKieServerBaseIntegrat
 
         try {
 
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
```diff
@@ -900,10 +900,17 @@ public void testFindTaskEvents() throws Exception {
 
         try {
 
-            List<TaskSummary> tasks = taskClient.findTasks("yoda", 0, 10);
+            List<TaskSummary> tasks = taskClient.findTasks("yoda", 0, 50);
```
mswiderski: @jakubschwan maybe instead of using findTasks we could use findTasksByStatusByProcessInstanceId (https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java#L119) instead so we could avoid this loop and increased number of records to load. Wdyt?

jakubschwan: Good idea. This return only one task, that we need. I think it should be work OK.
I'll do changes.

PR #283, opened 2016-02-01
[DROOLS-1049] fix config after upgrading jaxb2-maven-plugin from 1.6 to 2.2

psiroky:
This upgrade _should_ fix intermittent build failures we see on Jenkins, related to jaxb2-maven-plugin:schemagen.

Needs to be merged together with https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/141

@mrietveld could you please take a look?

### DIFF
--- a/kie-remote/kie-remote-jaxb-gen/pom.xml
+++ b/kie-remote/kie-remote-jaxb-gen/pom.xml
@@ -257,32 +257,8 @@
       </plugin>
 
       <plugin>
-        <!-- We use this plugin to ensure that our usage of the
-             jaxb2-maven-plugin is JDK 8 compatible. -->
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>properties-maven-plugin</artifactId>
-        <version>1.0-alpha-2</version>
-        <executions>
-          <execution>
-            <id>set-additional-system-properties</id>
-            <goals>
-              <goal>set-system-properties</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <properties>
-            <property>
-              <name>javax.xml.accessExternalSchema</name>
-              <value>file,http</value>
-            </property>
-          </properties>
-        </configuration>
-      </plugin>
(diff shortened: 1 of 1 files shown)


kie-remote/kie-remote-jaxb-gen/pom.xml
```diff
@@ -292,11 +268,10 @@
...
             <configuration>
-              <includes>
-                <include>org/drools/**/*.java</include>
-                <include>org/jbpm/**/*.java</include>
-              </includes>
-              <verbose>true</verbose>
+              <sources>
+                <source>${project.basedir}/src/main/java/org/drools</source>
+                <source>${project.basedir}/src/main/java/org/jbpm</source>
+              </sources>
```
mrietveld: @psiroky Pleaes keep `**/*.java`. The jars may eventually contain other classes besides `.java` classes, and I want to make sure that only `.java` classes are extracted into these directores. 

Good idea on adding the `${project.basedir}`!

mrietveld:
+1 as long as `**/*.java` is used in the pom.xml, instead of (only) the root `/org/jbpm/` or `/org/drools/` directories. 

Thanks!

kie-remote/kie-remote-jaxb-gen/pom.xml
psiroky: I've tried that, but could not make it work. The `source` expect either directory or specific file path. Using GLOBs does not seem to be supported.

psiroky: The plugin docs (http://www.mojohaus.org/jaxb2-maven-plugin/Documentation/v2.2/schemagen-mojo.html#sources) says that `sources` is: "Parameter holding List of paths to files and/or directories which should be recursively searched for Java source files".

mrietveld: Ah, apologies -- I probably should have RTFMed that myself. :) +1 on merging.

PR #287, opened 2016-02-05
JBPM-4928 - Clean up the KIE Remote Client (Java) API

mrietveld:
@mswiderski Would you mind looking at this? 

(Or @manstis?) 

It's just clean up, nothing has (/should have) changed on the functional side. 

@droolsjbpm/qe FYI: in case any of your tests reference the Factory classes or kie-remote-client internal classes. 

@dvirgiln FYI: I don't think fuse-bxms-integ/fuse/switchyard is using internal classes, anymore, but just in case. 

Depended on by: 
- https://github.com/droolsjbpm/kie-wb-distributions/pull/188


mswiderski:
looks ok to so +1 when build completes without errors

PR #288, opened 2016-02-08
PLANNER-505: initial implementation - ready for review

etirelli:
This is now ready for review and merging. 

A couple details to keep in mind:

1. There are 3 ignored tests. Geoffrey will review them after merging and we will fix them.

2. JMS implementation not ready. PM asked us to proceed with the merging as this will be part of the DR2 build.

### DIFF
--- a/kie-server-parent/kie-server-api/pom.xml
+++ b/kie-server-parent/kie-server-api/pom.xml
@@ -24,10 +24,15 @@
       <artifactId>kie-api</artifactId>
       <!-- scope>provided</scope -->
     </dependency>
+
+    <!-- we need to find a way to remove the following 2 dependencies from this jar -->
     <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
-      <!-- scope>provided</scope -->
+    </dependency>
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-core</artifactId>
     </dependency>
 
     <!-- json -->
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/OptaplannerServicesClientBuilder.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/OptaplannerServicesClientBuilder.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.client.helper;
+
+import org.kie.server.client.KieServicesConfiguration;
+import org.kie.server.client.RuleServicesClient;
+import org.kie.server.client.SolverServicesClient;
+import org.kie.server.client.impl.RuleServicesClientImpl;
+import org.kie.server.client.impl.SolverServicesClientImpl;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class OptaplannerServicesClientBuilder
+        implements KieServicesClientBuilder {
+
+    @Override
+    public String getImplementedCapability() {
+        return "OptaPlanner";
+    } // we should consolidate this into a constant in the kie-server-api jar
+
+    @Override
+    public Map<Class<?>, Object> build(KieServicesConfiguration configuration, ClassLoader classLoader) {
+        Map<Class<?>, Object> services = new HashMap<Class<?>, Object>();
+
+        services.put(SolverServicesClient.class, new SolverServicesClientImpl(configuration, classLoader));
+
+        return services;
+    }
+}
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.client.impl;
+
+import org.kie.api.command.Command;
+import org.kie.server.api.commands.CallContainerCommand;
+import org.kie.server.api.commands.CommandScript;
+import org.kie.server.api.model.KieServerCommand;
+import org.kie.server.api.model.ServiceResponse;
+import org.kie.server.api.model.instance.SolverInstance;
+import org.kie.server.api.model.instance.SolverInstanceList;
+import org.kie.server.api.rest.RestURI;
+import org.kie.server.client.KieServicesConfiguration;
+import org.kie.server.client.RuleServicesClient;
+import org.kie.server.client.SolverServicesClient;
+import org.optaplanner.core.api.domain.solution.Solution;
+
+import java.util.Collections;
+
+public class SolverServicesClientImpl
+        extends AbstractKieServicesClientImpl implements SolverServicesClient {
+
+    public SolverServicesClientImpl(KieServicesConfiguration config) {
+        super(config);
+    }
+
+    public SolverServicesClientImpl(KieServicesConfiguration config, ClassLoader classLoader) {
+        super(config, classLoader);
+    }
+
+    @Override
+    public ServiceResponse<SolverInstanceList> getSolvers(String containerId) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId );
+            return makeHttpGetRequestAndCreateServiceResponse( uri, SolverInstanceList.class );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    @Override
+    public ServiceResponse<SolverInstance> createSolver(String containerId, String solverId, SolverInstance instance) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        checkMandatoryParameter( "SolverId", solverId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId, solverId );
+            return makeHttpPutRequestAndCreateServiceResponse( uri, instance, SolverInstance.class );
+        } else {
+//            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+//            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    @Override
+    public ServiceResponse<SolverInstance> getSolverState(String containerId, String solverId) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        checkMandatoryParameter( "SolverId", solverId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId, solverId );
+            return makeHttpGetRequestAndCreateServiceResponse( uri, SolverInstance.class );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    @Override
+    public ServiceResponse<SolverInstance> getSolverBestSolution(String containerId, String solverId) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        checkMandatoryParameter( "SolverId", solverId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId, solverId ) + RestURI.SOLVER_BEST_SOLUTION;
+            return makeHttpGetRequestAndCreateServiceResponse( uri, SolverInstance.class );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    @Override
+    public ServiceResponse<SolverInstance> updateSolverState(String containerId, String solverId, SolverInstance instance) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        checkMandatoryParameter( "SolverId", solverId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId, solverId );
+            return makeHttpPostRequestAndCreateServiceResponse( uri, instance, SolverInstance.class, getHeaders( instance ) );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    @Override
+    public ServiceResponse<Void> disposeSolver(String containerId, String solverId) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        checkMandatoryParameter( "SolverId", solverId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId, solverId );
+            return makeHttpDeleteRequestAndCreateServiceResponse( uri, Void.class );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            throw new UnsupportedOperationException( "JMS is not supported at the moment." );
+        }
+    }
+
+    private String getURI(String containerId) {
+        return (baseURI + "/" + RestURI.SOLVER_URI ).replace( "{"+RestURI.CONTAINER_ID+"}", containerId );
+    }
+
+    private String getURI(String containerId, String solverId) {
+        return (baseURI + "/" + RestURI.SOLVER_URI + RestURI.SOLVER_ID_URI).replace( "{"+RestURI.CONTAINER_ID+"}", containerId ).replace( "{"+RestURI.SOLVER_ID+"}", solverId );
+    }
+
+    private void checkMandatoryParameter(String parameterName, String parameter ) {
+        if ( parameter == null || parameter.isEmpty() ) {
+            throw new IllegalArgumentException( parameterName + " can not be null or empty." );
+        }
+    }
+
+}
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-optaplanner/src/main/java/org/kie/server/remote/rest/optaplanner/resources/Messages.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-optaplanner/src/main/java/org/kie/server/remote/rest/optaplanner/resources/Messages.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.remote.rest.optaplanner.resources;
+
+public class Messages {
+
+    public static final String UNEXPECTED_ERROR = "Unexpected error during processing: {0}";
+
+}
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.services.optaplanner;
+
+import org.kie.server.api.KieServerConstants;
+import org.kie.server.services.api.*;
+import org.kie.server.services.impl.KieServerImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.concurrent.*;
+
+public class OptaplannerKieServerExtension
+        implements KieServerExtension {
+
+    private static final Logger logger = LoggerFactory.getLogger( OptaplannerKieServerExtension.class );
+
+    public static final String EXTENSION_NAME = "OptaPlanner";
+
+    private static final Boolean droolsDisabled = Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_DROOLS_SERVER_EXT_DISABLED, "false"));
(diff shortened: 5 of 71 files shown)


mswiderski:
@etirelli when it comes to kie-server-api and domain objects I opt for having them specialized as opposed to reuse public api from optaplanner. Main reason is that they can be designed in the way remote requires it - both in terms of structure and data that should be send over the wire.

long time to create container usually indicates maven is resolving kjar's dependencies and in case snapshots are involved this might take long time in case there are lots of transitive deps which might be the case when adding our core libs (in this case optaplanner-core and optaplanner-persistence-xstream). Adding classes to jaxb context is not time consuming once the list of classes is known.

Personally I would like to avoid any asynchronous calls for creating containers as this brings in lots of complexity and in my opinion it's not worth the effort due to following:
- introducing asynchronous requires use backend support from app server which usually is ejb based to accommodate WAS support
- next once we have async jobs then we need to provide support for polling for job status which becomes difficult in clustered environments behind load balancer as you can hit another server instance that the job was submitted to
- async for containers brings not so much value as the timeout is client side only meaning creation of container will still happen even if client times out. Moreover when using controller which will be responsible for creating containers with be done in async way and we can use long timeout value there. New UI for controller includes messages from server/container which will bring in the details about status of create container calls.

the URI for optaplanner should be similar as we have it for example for processes: https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/ProcessResource.java#L52
that would give use something like this:
/server/containers/{containerId}/solvers/{solverId}
so to configure this on OptaplannerResource class use following:
in RestURI define the PLANNER_URI
```
public static final String PLANNER_URI = "containers/{" + CONTAINER_ID + "}/solvers";
```
next in the OptaplannerResource class:
```
@Path("server/" + PLANNER_URI)
```

Last question, is use of ServiceResponse really needed to be wrapper on top of Solver results?

ge0ffrey:
On 08/02/16 04:18, Edson Tirelli wrote:
    
    
      @ge0ffrey 
      
        
          We are facing some difficulties due to the fact that
            optaplanner-core contains both the public API and domain
            interfaces and the implementations (e.g: Solution, Score,
            etc). Would it be possible to split them into an
            optaplanner-api jar that does not bring the whole set of
            dependencies that are there due to the implementation? In
            particular, we can't add optaplanner-core as a dependency to
            kie-server-api jar. So unless we can split these interfaces
            into a separate optaplanner-api jar, we will need to
            duplicate the interfaces in the kie-server-api jar and use
            them there copy&pasting data during marshalling. 
          
            See: https://github.com/droolsjbpm/droolsjbpm-integration/compare/6.4.x...etirelli:PLANNER-505?expand=1#diff-c03f98ee0c28a670eff894409b97a1e2R120
          
        
      
    
    
    I don't think I can split the public api from the internal impl.
    - I clearly separate them in separate package (core.api.* and
    core.impl.*), so there is a real public api cut out.
    - but just like java.* and sun.*, the public api has internal code
    that directly uses the impl classes (so no reflection tricks etc
    which aren't always realistic)...
    With Jigsaw (JDK9), this will still be 1 module, but the public api
    is public and the impl is module encapsulated (so it's really
    hidden).
    
    Let's talk about this in real-time to discuss the problem, the cause
    and the potential fixes.
    
    
      
        
          
          
        
        
          Related to that, do we need to actually marshall a complete
            Solution instance? Or can we marshall simply the Number[]
            and reconstruct the Solution during unmarshalling?
          
            See: https://github.com/droolsjbpm/droolsjbpm-integration/compare/6.4.x...etirelli:PLANNER-505?expand=1#diff-35fdf428277cbe7951029bf67ae1495aR29
          
        
      
    
    A Solution cannot be turned into Number[].
    A Score can, but I see no reason why you'd do that.
    A Score is usually 2 numbers, but sometimes 3 or 5 numbers (rarely
    more than 5).
    A Solution has a Score.
    
    In the first phase, I suggest marshaling the complete Solution.
    For most use cases, this won't be a problem (although there are
    notable exceptions, we 'll deal with that in the next phase).
    For example in cloud balancing, a Solution could be 100 Computers
    and 300 Processes. Some idea's on size in XML:
    
    
    
    
      
        
          
          
        
        3. The cloud balancing example you sent me by e-mail is not
          working it seems. It fails during annotation scanning. Can you
          please take a look and fix it?
          
            See: https://github.com/etirelli/droolsjbpm-integration/tree/PLANNER-505/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance
          
        
      
    
    That's PLANNER-522, which I discovered on Friday. I've got a
    reproducer in optaplanner-core and I am trying to fix now
    (it's caused because KieContainer's mfs:URL protocol is not
    compatible with org.reflections).
    I am expecting to fix it today or tomorrow:
      https://issues.jboss.org/browse/PLANNER-522
    
    
    
      With kind regards,
        Geoffrey De Smet

etirelli:
@mswiderski and @ge0ffrey and @sutaakar

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-optaplanner/src/main/java/org/kie/server/remote/rest/optaplanner/resources/Messages.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.remote.rest.optaplanner.resources;
+
+public class Messages {
+
+    public static final String PROCESS_DEFINITION_NOT_FOUND = "Could not find process definition \"{0}\" in container \"{1}\"";
```
mswiderski: I believe majority of these messages are not used in optaplanner so maybe worth removing them?

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverInstanceContext.java
```diff
@@ -0,0 +1,37 @@
+package org.kie.server.services.optaplanner;
```
mswiderski: missing license header

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
```diff
@@ -0,0 +1,234 @@
...
+      <artifactId>httpclient</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>fluent-hc</artifactId>
+      <scope>test</scope>
+    </dependency>
+   
+    <!-- persistence -->
```
mswiderski: minor thing - persistence is not used in optaplanner I believe...

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/btm-config.properties
```diff
@@ -0,0 +1,4 @@
+bitronix.tm.serverId=tomcat-btm-node0
```
mswiderski: not needed for optaplanner, same for btm-resources.properties

kie-server-parent/kie-server-api/pom.xml
```diff
@@ -24,11 +24,18 @@
...
+    <!-- we need to find a way to remove the following 2 dependencies from this jar -->
     <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
       <!-- scope>provided</scope -->
     </dependency>
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-core</artifactId>
+      <!-- scope>provided</scope -->
```
mswiderski: shall we make it optional?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public ServiceResponse<SolverInstanceList> getSolvers(String containerId) {
+        checkMandatoryParameter( "ContainerID", containerId );
+        if( config.isRest() ) {
+            String uri = getURI( containerId );
+            return makeHttpGetRequestAndCreateServiceResponse( uri, SolverInstanceList.class );
+        } else {
+            //            CommandScript script = new CommandScript( Collections.singletonList((KieServerCommand) new CallContainerCommand(id, payload)) );
+            //            return (ServiceResponse<String>) executeJmsCommand( script ).getResponses().get( 0 );
+            return null;
```
mswiderski: maybe more meaningful would be to throw UnsupportedOperationException as null might get people worried that it does not work while it's simply not supported/implemented

mswiderski:
looks good to me besides these minor comments added inline.

psiroky:
Looks like there were some changes which broke the kie-server-client tests on Karaf (see the test result on Jenkins).

kie-server-parent/kie-server-api/pom.xml
triceo: Let's do whatever is done for Drools and jBPM also. I'm not aware of Planner getting any sort of special status - am I wrong?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/AbstractKieServicesClientImpl.java
```diff
@@ -62,7 +62,7 @@
 
     protected String baseURI;
     protected final KieServicesConfiguration config;
-    protected final Marshaller marshaller;
+    public final Marshaller marshaller;
```
etirelli: Missed this. I changed just for a quick debug. Will revert to protected.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-optaplanner/src/main/java/org/kie/server/remote/rest/optaplanner/resources/Messages.java
etirelli: Done.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverInstanceContext.java
etirelli: Done.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
etirelli: Damn copy&paste. Done.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/btm-config.properties
etirelli: Done.

kie-server-parent/kie-server-api/pom.xml
etirelli: As I mentioned before, I would rather try to remove the dependency completely, but we can try with optional for now.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
etirelli: Done.

etirelli:
PR updated based on previous comments.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/OptaplannerServicesClientBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class OptaplannerServicesClientBuilder
+        implements KieServicesClientBuilder {
+
+    @Override
+    public String getImplementedCapability() {
+        return "Optaplanner";
```
ge0ffrey: It should be either "OptaPlanner" or "optaplanner"

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
```diff
@@ -0,0 +1,140 @@
...
+            return;
+        }
+        this.registry = registry;
+        // the following threadpool will have a max thread count equal to the number of cores on the machine.
+        // if new jobs are submited and all threads are busy, the reject policy will kick in.
+        this.threadPool = new ThreadPoolExecutor(2, // core size
+                                                 Runtime.getRuntime().availableProcessors(), // max size
+                                                 120, // idle timeout
+                                                 TimeUnit.SECONDS,
+                                                 new ArrayBlockingQueue<Runnable>(Runtime.getRuntime().availableProcessors())); // queue with a size
```
ge0ffrey: From experience, I 'd recommend to take a more conservative number for the threadpool size:

    threadPoolSize = (availableProcessors < 4 ? availableProcessors : availableProcessors - 1);

Why? If you have 8 processors and 8 solver threads running in parallel, which each take 100% of their processor (remember that they don't do IO, so they never block), then any REST request coming in will need to rely on the JVM's round-robin thread scheduling to park at least 1 of the running solver threads to get some time to be handled (and for the OS and for EAP threads). I believe this will affect response time (= latency). If we have 4 or more processors, I think the tradeoff is worth to reserve 1 processor to optimize latency over throughput (note that such a reservation is virtual because solver threads jump to another cpu all the time, but you will see that total cpu time isn't maxed out).

In benchmarks, I am even more conservative, only using half of the processorSize by default. This is mainly because on hardware with hyper-threading, stressing more than half of the CPU's clearly reduces the average speed per CPU.

In any case: in the long run that thread pool size will need to be configurable for the SysOps (through JMX I suppose).

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
```diff
@@ -0,0 +1,140 @@
...
+    public <T> T getAppComponents(Class<T> serviceType) {
+        if ( serviceType.isAssignableFrom( solverServiceBase.getClass() ) ) {
+            return (T) solverServiceBase;
+        }
+        return null;
+    }
+
+    @Override
+    public String getImplementedCapability() {
+        return "Optaplanner";
```
ge0ffrey: => OptaPlanner or optaplanner

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -0,0 +1,323 @@
...
+
+    public ServiceResponse<SolverInstanceList> getSolvers(String containerId) {
+        try {
+            List<SolverInstance> sl = getSolversForContainer( containerId );
+            return new ServiceResponse<SolverInstanceList>(
+                    ServiceResponse.ResponseType.SUCCESS,
+                    "Solvers list successfully retrieved from container '" + containerId + "'",
+                    new SolverInstanceList( sl ) );
+        } catch ( Exception e ) {
+            logger.error( "Error retrieving solvers list from container '" + containerId + "'", e );
```
ge0ffrey: I wondering if we can do anything to do all these exception handling code through an aspect?
It's always the same boilerplate, making the code harder to read.
With CDI we might be able to add a simple interceptor for all these methods I think.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -0,0 +1,323 @@
...
+
+                    solvers.put( instance.getSolverInstanceKey(), sic );
+
+                    logger.info( "Solver '" + solverId + "' successfully created in container '" + containerId + "'" );
+                    return new ServiceResponse<SolverInstance>(ServiceResponse.ResponseType.SUCCESS,
+                                                               "Solver '" + solverId + "' successfully created in container '" + containerId + "'",
+                                                               instance );
+
+                } catch( Exception e ) {
+                    logger.error("Error creating solver factory for solver " + instance, e);
```
ge0ffrey: Why duplicate this exception handling inside and outside the synchronized statement? I don't see any race condition if it was simply all handled by the outer exception handling.

kie-server-parent/kie-server-api/pom.xml
sutaakar: I think we cannot make it optional as classes from optaplanner-core are defined in JaxbMarshaller.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -0,0 +1,326 @@
...
+        instance.setSolverConfigFile( SOLVER_1_CONFIG );
+        assertSuccess( client.createContainer( CONTAINER_1_ID, new KieContainerResource( CONTAINER_1_ID, kjar1 ) ) );
+        assertSuccess( solverClient.createSolver( CONTAINER_1_ID, SOLVER_1_ID, instance ) );
+
+        instance.setStatus( SolverInstance.SolverStatus.NOT_SOLVING );
+        instance.setPlanningProblem( loadPlanningProblem( 50, 300 ) );
+        assertSuccess( solverClient.updateSolverState( CONTAINER_1_ID, SOLVER_1_ID, instance ) );
+    }
+
+    @Test @Ignore("Ignoring for now. Geoffrey will review why this is failing.")
```
sutaakar: I think this is failing because solver is started asynchronously:
Method updateSolverState in SolverServiceBase starts solver in new thread and then update state of SolverInstanceContext in updateSolverStatus. There is condition solver.isSolving(), but result of that condition depends on whether thread with solver already started or not.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -0,0 +1,326 @@
...
+        assertSuccess( client.createContainer( CONTAINER_1_ID, new KieContainerResource( CONTAINER_1_ID, kjar1 ) ) );
+
+        ServiceResponse<SolverInstance> response = solverClient.createSolver( CONTAINER_1_ID, SOLVER_1_ID, instance );
+        assertSuccess( response );
+        assertEquals( SolverInstance.SolverStatus.NOT_SOLVING, response.getResult().getStatus() );
+
+        // the following status starts the solver
+        instance.setStatus( SolverInstance.SolverStatus.SOLVING );
+        instance.setPlanningProblem( loadPlanningProblem( 5, 10 ) );
+        response = solverClient.updateSolverState( CONTAINER_1_ID, SOLVER_1_ID, instance );
```
sutaakar: What about creating own separate methods in client for starting solver and terminating it? It would be more intuitive and user friendly from my point of view. Also it would show what parameters are mandatory for each action.

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/filtered-resources/kjars-sources/deployment-descriptor-project/src/main/resources/conditional-event.bpmn2
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<bpmn2:definitions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://www.omg.org/bpmn20" xmlns:bpmn2="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:bpsim="http://www.bpsim.org/schemas/1.0" xmlns:color="http://www.omg.org/spec/BPMN/non-normative/color" xmlns:dc="http://www.omg.org/spec/DD/20100524/DC" xmlns:di="http://www.omg.org/spec/DD/20100524/DI" xmlns:drools="http://www.jboss.org/drools" id="_TqMGgFrcEeWYPtWtxMzMaA" xsi:schemaLocation="http://www.omg.org/spec/BPMN/20100524/MODEL BPMN20.xsd http://www.jboss.org/drools drools.xsd http://www.bpsim.org/schemas/1.0 bpsim.xsd" exporter="jBPM Designer" exporterVersion="6.2.0" expressionLanguage="http://www.mvel.org/2.0" targetNamespace="http://www.omg.org/bpmn20" typeLanguage="http://www.java.com/javaTypes">
```
ge0ffrey: Did you intend to add this bpmn2 file to this pull request?

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/src/main/java/org/kie/server/testing/Labeled.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.testing;
+
+public interface Labeled {
```
ge0ffrey: Overkill, I 'll send a PR to remove all code in the copy of the example that's overkill.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/src/main/java/org/kie/server/testing/CloudComputerStrengthComparator.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.kie.server.testing;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import org.apache.commons.lang3.builder.CompareToBuilder;
+
+public class CloudComputerStrengthComparator implements Comparator<CloudComputer>, Serializable {
```
ge0ffrey: Overkill, I 'll send a PR to remove all code in the copy of the example that's overkill.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/src/main/java/org/kie/server/testing/CloudProcessDifficultyComparator.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.kie.server.testing;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+import org.apache.commons.lang3.builder.CompareToBuilder;
+
+public class CloudProcessDifficultyComparator implements Comparator<CloudProcess>, Serializable {
```
ge0ffrey: Overkill, I 'll send a PR to remove all code in the copy of the example that's overkill.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/src/main/resources/META-INF/kmodule.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<kmodule xmlns="http://jboss.org/kie/6.0.0/kmodule">
```
ge0ffrey: As of 6.4.0.Beta1, the kmodule xmlns namespace has switched to "http://www.drools.org/xsd/kmodule"
We should never use "http://jboss.org/kie/6.0.0/kmodule" again (except in the regression tests in drools itself because it still works for backwards compatibility)

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -0,0 +1,323 @@
...
+                        if( instance.getStatus() == SolverInstance.SolverStatus.SOLVING ) {
+                            if( instance.getPlanningProblem() == null ) {
+                                return new ServiceResponse<SolverInstance>(ServiceResponse.ResponseType.FAILURE,
+                                                                           "Planning-problem is a mandatory field when starting the solver.",
+                                                                           instance );
+                            }
+                            startSolver( sic, instance );
+                        } else if( instance.getStatus() == SolverInstance.SolverStatus.NOT_SOLVING ) {
+                            return new ServiceResponse<SolverInstance>(ServiceResponse.ResponseType.SUCCESS,
+                                                                       "Solver '" + solverId + "' on container '" + containerId + "' is already executing.",
```
sutaakar: It seems that this response text should be switched with response text on line 199. Here it is returned in case original status is NOT_SOLVING and updated status is NOT_SOLVING too. So response here should be that solving didn't started yet or already terminated.

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/filtered-resources/kjars-sources/deployment-descriptor-project/src/main/resources/conditional-event.bpmn2
mswiderski: yes, it is required as we made jbpm extension to activate only where there are processes in kjar and this test kjar requires jbpm extension to be enabled

kie-server-parent/kie-server-api/pom.xml
mswiderski: very good point @sutaakar and looks like this was the reason build failed on jenkins. My wrong thinking was that optional applies only when used as dependency and thus would make it optional as transitive but anyway since it was compiled with reference (in import section) that would fail as well on runtime. So we have to go back to non optional dependency for optaplanner

kie-server-parent/kie-server-api/pom.xml
```diff
@@ -24,10 +24,17 @@
       <artifactId>kie-api</artifactId>
       <!-- scope>provided</scope -->
     </dependency>
+
+    <!-- we need to find a way to remove the following 2 dependencies from this jar -->
     <dependency>
       <groupId>org.drools</groupId>
       <artifactId>drools-core</artifactId>
-      <!-- scope>provided</scope -->
+      <scope>optional</scope>
```
ge0ffrey: optional is not a scope, should be
<optional>true</optional>
instead
I am preparing a PR for edson's fork.

etirelli:
This is now squashed and ready for review/merge.

etirelli:
/assign @mswiderski @ge0ffrey @psiroky @sutaakar 

Tests succeed for me locally.

mswiderski:
+1 looks good to me

sutaakar:
+1

etirelli:
Merged.

PR #291, opened 2016-02-09
[DROOLS-1055] Fix ClassCastException when running with Fuse 6.2.1 (bl…

grgrzybek:
…ueprint-core 1.4.4)


grgrzybek:
@psiroky mind that I'm only fixing particular bug in Fuse, I don't know the *big picture* of drools/jbpm/brms

psiroky:
ok to test

psiroky:
ok to test

psiroky:
+1 to merge. There is no test, but as uncovered by @grgrzybek all the current tests are disabled (yikes), because they are failing (likely test setup issue). I will first need to fix those. Then we can take a look at adding specific test case for this case.
 
I will merge this shortly.

PR #293, opened 2016-02-10
DROOLS-1056 - test fix for mysql55

sutaakar:
Queries in QueryDataServiceIntegrationTest failed when I tried them with MySQL5.5 database.
This changes will make them work.


mswiderski:
+1
this indicates that mysql was running in case sensitive mode meaning same fixes needs to be added in jbpm tests for query services

^^ @ibek just so you know as we discussed about these tests recently

@sutaakar if you find some time please submit PR for 6.4.x as that will be needed anyway, thanks!

mswiderski:
merged as part of #299

PR #294, opened 2016-02-10
DROOLS-1056 - test fix for mysql55

sutaakar:
Cherrypick of https://github.com/droolsjbpm/droolsjbpm-integration/pull/293 for 6.4.x


mswiderski:
jenkins retest this

mswiderski:
these test failures do not seem to be at all related to this PR... do we have any issues with build on jenkins?

sutaakar:
@mswiderski I have noticed such errors even in my local build. I am not sure what is causing them, they don't seem to be related to this change.

psiroky:
I am not aware of any Jenkins related issues. It may be that the test started to fail because of some underlying change in drools/jbpm.

psiroky:
This one looks interesting https://github.com/droolsjbpm/drools/commit/183bf11f753257f7d06f20f350d8cfbf74da5be3

mswiderski:
thanks @psiroky @sutaakar it looks we would have to revisit all our tests and replace repository.deploy with repository.install apparently ...will try this later today. Anyway don't see a point of blocking this PR because of that.

psiroky:
Yes, no need to block this one. The tests should probably use install, but there could be other changes needed as well (if I am remembering correctly the way the maven test projects are used in the kie-server tests).

sutaakar:
Seems that only tests using SNAPSHOT test artifact version fails.

mswiderski:
merged int 6.4.x, thanks @sutaakar

PR #297, opened 2016-02-11
[DROOLS-1060] upgrade Servicemix POI bundle from 3.9_2 to 3.11_1

psiroky:
* this is a temporary upgrade as we ultimately need to use 3.13.
   However, the 3.13_1 bundle is borked and useless, because it
   self-imports package which does exist anymore.  We need
   3.13_2 whis is not yet availabe. Also at the same time, recent
   changes in drools-decisiontables caused incompatibility with
   current, ancient versin 3.9, so we need to ugprade now.


psiroky:
@mariofusco could you please quickly check this? I need this on 6.4.x to fix the fuse-bxms-integ which is currently failing.

mswiderski:
jenkins retest this

mswiderski:
same here +1 once the build succeeds

PR #300, opened 2016-02-12
drools-karaf-itests: add test for Spring context in META-INF/spring

winklerm:
Added test for case when Spring auto-detects application context stored in META-INF/spring/ in a bundle.

This use-case failed for me because of my mistake (not using OSGi-enabled KModuleBeanFactoryPostProcessor), adding the test just to prevent possible regression in future.

### DIFF
--- a/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/KieSpringMetaInfAppContextOnKarafKarafIntegrationTest.java
+++ b/drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/KieSpringMetaInfAppContextOnKarafKarafIntegrationTest.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2015 Red Hat
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


drools-osgi/drools-karaf-itests/src/test/java/org/drools/karaf/itest/KieSpringMetaInfAppContextOnKarafKarafIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.junit.Assert.assertNotNull;
+import static org.ops4j.pax.exam.CoreOptions.streamBundle;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.configureConsole;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.keepRuntimeFolder;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.logLevel;
+import static org.ops4j.pax.tinybundles.core.TinyBundles.bundle;
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerMethod.class)
+public class KieSpringMetaInfAppContextOnKarafKarafIntegrationTest extends AbstractKarafIntegrationTest {
```
psiroky: There is a "Karaf" two times in a row in the test name, which looks a bit weird.

psiroky: Ah, I see that it probably comes from the other test names. Let me fix that when merging the PR.

psiroky:
+1, looks good. Will merge it once Jenkins approves.

psiroky:
Test failures are not related to this PR.

PR #303, opened 2016-02-15
JBPM-4900 - added test covering replaceQuery

sutaakar:
There is NPE for JAXB run of test, it is thrown in https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskInstanceList.java#L55 .
JAXB marshaller doesn't distinguish between marshalling empty array and null when sending arrays in Kie server. Maybe we can fix it by adding @XmlElementWrapper annotation for those arrays.


sutaakar:
@ibek FYI

mswiderski:
@sutaakar take a look at #307 that uses slightly different approach - simple null check to avoid changes in response structure. Adding XmlElementWrapper will affect the returned content for both JAXB ad JSON and this I want to avoid. Closing this PR as #307 includes this commit too

PR #304, opened 2016-02-16
JBPM-4926 Enhance kie-performance-kit to run soak tests

ibek:


mswiderski:
+1

if that shall be backported to 6.4.x please submit separate PR, thanks

ibek:
I intended this enhancement only to 7.x - Also I plan to do some structural changes due to the different test suites, so it will be better to leave to 7.x

PR #313, opened 2016-02-18
[6.4.x] [BZ-1305206] kie-maven-plugin: upgrade takari-lifecycle-plugin to 1.11.11.jbossorg-1

psiroky:
* this is only temporary until there is a proper 1.11.12 release
   directly from Takari


mbiarnes:
+1 build droolsjbpm-integration with this change and executed all tests

PR #317, opened 2016-02-19
Allow BigIntegers to be defined in the /process/*/start/ map parameters

xmrk:
This incremental improvement on the object marshaling of numeric map parameters allows BigIntegers to be placed on the process instance's parameters.  These big integers can better accommodate primary keys on relational databases.


psiroky:
@mrietveld could you please take a look at this one?

mrietveld:
@psiroky Please close. 

@xmrk Unfortunately, we're deprecating this code for 7.0. However, I've copied this code so that we can use it for the new code where the REST API is implemented.

mswiderski:
closing based on @mrietveld, request

PR #319, opened 2016-02-22
kie-server tests: remove additional dependencies for kie-maven-plugin

psiroky:
* no need to include thhose additional dependencies, they are already
   brought in by the kie-maven-plugin by default


mswiderski:
+1 to merge

PR #320, opened 2016-02-23
kie-server tests: remove additional dependencies for kie-maven-plugin

psiroky:
* no need to include those additional dependencies, they are already
   brought in by the kie-maven-plugin by default


psiroky:
Same as #319, but for master.

mswiderski:
+1 once the build finishes

psiroky:
I see it has been copy-pasted into the optaplanner tests as well (https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/common-parent/pom.xml#L25). Will remove it there as well.

PR #321, opened 2016-02-23
[6.3.x] [DROOLS-1055, BZ-1307079] Fix ClassCastException when running with Fu…

psiroky:
…se 6.2.1 (blueprint-core 1.4.4)


krisv:
+1 for picking up in 6.2.2

PR #333, opened 2016-02-29
BZ-1312887 - kie-server JSON doesn't work with singleton extra classes

mswiderski:
@etirelli could you take a look?


etirelli:
Maciej, we might need to clone it in the marshaller as well:

https://github.com/mswiderski/droolsjbpm-integration/blob/BZ-1312887/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java#L82

The marshaller can be instantiated by itself, what would trigger the same issue...

mswiderski:
in theory it can be though I'd be more tempted to leave it to fail if it was not created via KieServiceClient. 

wdyt?

PR #335, opened 2016-03-01
PLANNER-505: adding support for JMS

etirelli:
Adding support for JMS transport. 

@ge0ffrey not sure why the kjar with the planner cloudbalance example is raising compilation failures. It is not related to my changes, as it was already happening. Will investigate tomorrow.

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/CreateSolverCommand.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/CreateSolverCommand.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.api.commands.optaplanner;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import org.kie.server.api.model.KieServerCommand;
+
+import javax.xml.bind.annotation.*;
+
+@XmlRootElement(name = "create-solver")
+@XStreamAlias("create-solver")
+@XmlAccessorType(XmlAccessType.NONE)
+public class CreateSolverCommand
+        implements KieServerCommand {
+
+    private static final long serialVersionUID = -1803374525440238478L;
+
+    @XmlAttribute(name = "container-id")
+    @XStreamAlias("container-id")
+    private String containerId;
+
+    @XmlAttribute(name = "solver-id")
+    @XStreamAlias("solver-id")
+    private String solverId;
+
+    @XmlElement(name = "solver-config-file")
+    @XStreamAlias("solver-config-file")
+    private String solverConfigFile;
+
+    public CreateSolverCommand() {
+        super();
+    }
+
+    public CreateSolverCommand(String containerId, String solverId, String solverConfigFile) {
+        this.containerId = containerId;
+        this.solverId = solverId;
+        this.solverConfigFile = solverConfigFile;
+    }
+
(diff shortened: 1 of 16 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/CreateSolverCommand.java
```diff
@@ -0,0 +1,88 @@
...
+            return true;
+        if ( obj == null )
+            return false;
+        if ( getClass() != obj.getClass() )
+            return false;
+        CreateSolverCommand other = (CreateSolverCommand) obj;
+        if ( instance == null ) {
+            if ( other.instance != null )
+                return false;
+        } else if ( !instance.equals( other.instance ) )
```
ge0ffrey: this will hardly ever be equal as SolverInstance doesn't implement equals (so they are only equal if they are the same instance, which is never).

I suggest removing the equals/hashcode code from this Command for now - until it's sure that we need to have such code.

etirelli:
@ge0ffrey @mswiderski @sutaakar this is ready for merging. All tests passed locally. Can you please review?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/UpdateSolverStateCommand.java
```diff
@@ -0,0 +1,108 @@
...
+    @XStreamAlias("container-id")
+    private String containerId;
+
+    @XmlAttribute(name = "solver-id")
+    @XStreamAlias("solver-id")
+    private String solverId;
+
+    @XmlElement
+    @XStreamAlias("solver-instance")
+    private String instance;
```
ge0ffrey: Why is solver-instance of type String?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
```diff
@@ -100,13 +104,15 @@ public SolverServicesClientImpl(KieServicesConfiguration config, ClassLoader cla
     public ServiceResponse<SolverInstance> updateSolverState(String containerId, String solverId, SolverInstance instance) {
         checkMandatoryParameter( "ContainerID", containerId );
         checkMandatoryParameter( "SolverId", solverId );
+        checkMandatoryParameter( "instance", instance );
```
sutaakar: I think that SolverInstance don't need to be mandatory when terminating solver.

Maybe best solution would be to separate starting and terminating of solvers into separate client methods with own mandatory parameter validations.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/UpdateSolverStateCommand.java
etirelli: Because the instance may contain user-provided classes (for instance, the problem to solve). This requires that attribute to be unmarshalled with the kie-container classloader, not the kie-server classloader. Unfortunately, this is the only way we found to do it. It is the same artifice drools/jbpm use.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SolverServicesClientImpl.java
etirelli: The update method requires the solver state attribute that is inside the solver instance, so it is mandatory.

I agree that having client methods for specific operations would make it easier to use, but as of today, this is the same across drools/jbpm as well. We can introduce the additional client operations as part of the next version.

sutaakar:
+1 for me

mswiderski:
changes looks ok to me though we have quite a few test failures related to this changes, or did I missed something?

etirelli:
:( I will try to see what's going on. Tests succeeded locally, but I had a bunch of environment issues before.

etirelli:
OK, additional fixes in, squashed, rebased and re-submitted. 

@sutaakar @mswiderski @ge0ffrey in case you want to check again, tests should succeed now.

sutaakar:
Tests failing because of changed behaviour of REST DELETE request in Kie client.

etirelli:
Hi Karel, yes, I fixed it locally, and I am re-running all the tests now.

etirelli:
Jenkins, be nice to me! :)

etirelli:
YEY, all tests succeeded! :) ready to merge.

mswiderski:
+1 to merge

etirelli:
merged.

PR #337, opened 2016-03-01
DROOLS-1073 Option to bypass start task operation in REST/JMS

ibek:
this PR depends on https://github.com/droolsjbpm/jbpm/pull/416

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
@@ -35,6 +35,8 @@ public interface UserTaskServicesClient {
 
     void completeTask(String containerId, Long taskId, String userId, Map<String, Object> params);
 
+    void completeAutoProgress(String containerId, Long taskId, String userId, Map<String, Object> params);
+
     void delegateTask(String containerId, Long taskId, String userId, String targetUserId);
 
     void exitTask(String containerId, Long taskId, String userId);
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
@@ -86,11 +86,24 @@ public class UserTaskServiceBase {
     public void complete(String containerId, Number taskId, String userId, String payload, String marshallerType) {
 
         userId = getUser(userId);
+        
         logger.debug("About to unmarshal task outcome parameters from payload: '{}'", payload);
         Map<String, Object> parameters = marshallerHelper.unmarshal(containerId, payload, marshallerType, Map.class);
 
         logger.debug("About to complete task with id '{}' as user '{}' with data {}", taskId, userId, parameters);
         userTaskService.complete(taskId.longValue(), userId, parameters);
+
+    }
+
+    public void completeAutoProgress(String containerId, Number taskId, String userId, String payload, String marshallerType) {
+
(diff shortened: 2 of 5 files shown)


kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
```diff
@@ -35,6 +35,8 @@
 
     void completeTask(String containerId, Long taskId, String userId, Map<String, Object> params);
 
+    void claimStartCompleteTask(String containerId, Long taskId, String userId, Map<String, Object> params);
+
```
mswiderski: I'd prefer to call to it completeTaskAutoProgress, wdyt?

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
```diff
@@ -86,6 +87,33 @@ public void claim(String containerId, Number taskId, String userId) {
...
+            logger.debug("About to start task automatically with id '{}' and user '{}'", taskId, userId);
+            userTaskService.start(taskId.longValue(), userId);
+            
+        } else if (task.getTaskData().getStatus() == Status.Reserved) {
+            
+            logger.debug("About to start task automatically with id '{}' and user '{}'", taskId, userId);
+            userTaskService.start(taskId.longValue(), userId);
+            
+        }
+
```
mswiderski: the issue with this code is that it creates several transactions:
- first to get task
- second to claim
- third to start
- forth to complete

of course it depends on the state of the task but still in worse case it will create and complete 4 transactions which is not efficient. I'd prefer to move this to jbpm services as this kind of functionality would make sense in embedded uses cases too.

ibek: Ok, I will look at how it could be implemented in the jbpm services.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
ibek: well, I thought that it would be clearer to call it claimStartCompleteTask - since it exactly says what it will do .... completeTaskAutoProgress is quite abstract but if it is ok and you think it's better, I will change it.

mswiderski: true, it does that at the moment but if we change the internals then it might not be as exact any more ;)

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
ibek: I should implement the functionality only into this service right? https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java#L170

mswiderski: create new method in UserTaskService to do the auto progress completeTask

mswiderski:
@ibek could you please take a look at the failure? looks like a compilation error...

ibek:
@mswiderski hurrayy, it passed - the error was strange - there was no Map on that line so I rerun the tests
I will create 6.4.x PRs, then

PR #340, opened 2016-03-02
kie-server tests: separated BTM for Tomcat builds

sutaakar:
Removed BTM config files from integration tests which doesn't use BPM extension.
Java 6 fix for OptaPlanner integration test.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
@@ -238,4 +238,47 @@
       </plugin>
     </plugins>
   </build>
+
+  <profiles>
+    <profile>
+      <id>tomcat7x</id>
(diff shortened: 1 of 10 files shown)


mswiderski:
do we need this for 6.4.x too?

sutaakar:
Yes, it should be cherrypicked into 6.4.x too.

Btw locally I got an exception when running OptaPlanner test:
	org/kie/server/testing/Rule_requiredCpuPowerTotal2067845905.java (2:169) : Only a type can be imported. org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScoreHolder resolves to a package
...

It seems to be unrelated to changes made in this commit, got this error on Tomcat runs and also WildFly.

sutaakar:
That error seems to happen just on 6.4.x branch.

mswiderski:
jenkins retest this

PR #353, opened 2016-03-14
Collection of improvements for KIE Server

mswiderski:
this PR comes with three updates for KIE Server mainly from jBPM point of view:
- add quartz to war files to allow to be used as timer service
- allow to set UserInfo via system properties same as for UserGroupCallback
- makes the load of containers from controller synchronous (it's configurable) to make sure that kie server is not yet available until it receives containers and they're fully loaded to avoid issues with too early use of the server

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
@@ -143,7 +143,14 @@ public class KieServerImpl {
                                                                                         containerManager,
                                                                                         this), "KieServer-ControllerConnect");
             connectToControllerThread.start();
-
+            if (Boolean.parseBoolean(currentState.getConfiguration().getConfigItemValue(KieServerConstants.CFG_SYNC_DEPLOYMENT, "true"))) {
+                logger.info("Containers were requested to be deployed synchronously, holding application start...");
+                try {
+                    connectToControllerThread.join();
+                } catch (InterruptedException e1) {
+                    logger.debug("Interrupt exception when waiting for deployments");
+                }
+            }
         }
 
         if (readyToRun) {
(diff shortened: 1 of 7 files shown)


mswiderski:
@sutaakar couldn't come up with any tests for these changes so if you have any idea that would be more than welcome

mrietveld:
I don't know that much about these, but from what I can see, +1

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
```diff
@@ -143,7 +143,14 @@ public KieServerImpl() {
                                                                                         containerManager,
                                                                                         this), "KieServer-ControllerConnect");
             connectToControllerThread.start();
-
+            if (Boolean.parseBoolean(currentState.getConfiguration().getConfigItemValue(KieServerConstants.CFG_SYNC_DEPLOYMENT, "true"))) {
+                logger.info("Containers were requested to be deployed synchronously, holding application start...");
+                try {
+                    connectToControllerThread.join();
+                } catch (InterruptedException e1) {
+                    logger.debug("Interrupt exception when waiting for deployments");
```
yurloc: @mswiderski I think you should re-interrupt the thread immediately after catching the InterruptedException otherwise it might be impossible to kill the server because of swallowing the interrupted flag at this point. I tried this but I didn't even see the debug log message so the flag might have been cleared somewhere else but anyway it seems that re-setting the flag in situations like this is the only correct approach, wdyt? http://stackoverflow.com/a/12339487/1691152

mswiderski: @yurloc I agree, though I am not 100% sure how to deal properly with this thread.join to be app server friendly and still secure the requirement for this feature... any ideas are more than welcome ... and PRs of course

PR #356, opened 2016-03-15
BZ-1310510 - 6.2.0: Remote java API: cannot get tasks assigned to different users

mrietveld:


mswiderski:
+1 to merge

nice that it does use org.kie.task.insecure property as default value :)

mrietveld:
@mswiderski Feel free to merge, when you get the chance. Thanks!

mswiderski:
merged into 6.4.x

ivenxu:
Do I need setting "org.kie.task.insecure" system property in server side? I assume the this PR is only setting correct value for rest client. According the https://github.com/droolsjbpm/droolsjbpm-integration/blob/0a38844ec0203a2b7f3a6ecf1bcf564db0ebf275/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/util/ExecuteCommandUtil.java#L44, when the server side getting the request, it evaluate the system property as well.

PR #357, opened 2016-03-15
BZ 1310439: add reproducer

winklerm:
Added reproducer for BZ 1310439.

The problem is when you install jbpm feature after drools-module feature (or kie-api, more precisely) had been installed before, and someone attempted to initialize jBPM both before jBPM is installed and after it was installed.


mswiderski:
closing this one as it's incorporated in https://github.com/droolsjbpm/droolsjbpm-integration/pull/363 to verify the changes in kie-api and jbpm

PR #367, opened 2016-03-17
Improve getBestSolution testing (backport to 6.4.x)

ge0ffrey:
JSON testing for that particular method is ignored for now as it's broken due the marshalling issue (and was already broken before these changes, the new tests just expose it).


mswiderski:
merged into 6.4.x

PR #371, opened 2016-03-21
kie-server: JSON marshaller fixes to support known use cases (drools, jbpm, optaplanner)

mswiderski:
@ge0ffrey finally ;)

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
@@ -16,12 +16,15 @@
 package org.kie.server.api.marshalling.json;
 
 import java.io.IOException;
+import java.lang.reflect.Array;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -110,38 +119,50 @@ public JSONMarshaller(Set<Class<?>> classes, ClassLoader classLoader) {
...
 
             for (Class<?> clazz : classes) {
                 mod.addSerializer(clazz, customObjectSerializer);
+
             }
 
             objectMapper.registerModule(mod);
-            customSerializationMapper.registerModule(mod);
-        }
+//            customSerializationMapper.registerModule(mod);
```
ge0ffrey: Is it intended to keep this commented line?

mswiderski: not any more, will remove it

ge0ffrey:
+1 to merge - Thanks Maciej!

mswiderski:
jenkins retest this

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -266,20 +266,14 @@ public void testGetBestSolution() throws Exception {
...
-            assertSuccess(solutionResponse);
-            solution = solutionResponse.getResult().getBestSolution();
-            // Only once the solution's score is null, the solution is fully uninitialized
-            // TODO add "|| solution.getScore().isInitialized()" once PLANNER-405 is fixed
-            if (solution != null && solution.getScore() != null) {
-                break;
-            }
-            Thread.sleep(1000);
-        }
+        Thread.sleep(5000);
```
ge0ffrey: 5000ms is both too much and too little. On a fast machine it unnecessarily slows down the test by 5 seconds (that's a lot, because 3 formats make 15secodns - if other tests would have this same philosophy, it would take so long to run all tests, that no one runs all tests before committing). On slow machines, it can fail with a false negative. The original code is intended to take 1000ms on a fast machine and 15000ms on a really slow one (and I believe we have to increase the loop to 60, so on really really slow machine 60000ms is still acceptable - well actually we have to make it an infinite loop and set a @Test.timeout of 60 seconds - I 'll send a PR later)

The fact that this change fixes the problem, means that something is not adhering to the contract. The contract states that the solution.getScore() is null as long as the solution isn't initialized. Furthermore, I believe the solution is null if the solver hasn't started yet. I suspect **the race condition we are seeing is because we're getting the solution of the previous run, if the solver hasn't started yet, instead of null.**

mswiderski:
jenkins retest this

ge0ffrey:
+1 to merge

PR #373, opened 2016-03-21
[BZ-1310039] use the bundle class loader when creating a KieContainer in kie-aries-blueprint

mariofusco:


psiroky:
Thanks @mariofusco, will take a look at this asap.

psiroky:
The failed test case does not seem to be related.

+1 to merge. 

@mariofusco quick question, am I assuming correctly that the

PR #374, opened 2016-03-21
[ENTESB-5168] drools-osgi: upgrade javax.servlet to 3.1.0

psiroky:
* this is needed to support Fuse 6.3
 * the javax.servlet:javax.servlet-api:3.1.0 is artifact used in
   Fuse 6.3.x. Works in Fuse 6.2.x as well.


psiroky:
This actually does not work with Fuse 6.2.x. Closing for now as we need to come with better way to support both versions of Fuse.

PR #381, opened 2016-03-24
BZ 1316966: add reproducer

winklerm:
Separated reproducer for the above BZ from https://github.com/droolsjbpm/droolsjbpm-integration/pull/315.


mswiderski:
included in #390, closing

PR #382, opened 2016-03-24
BZ-1319546 - Marshalling of results are done twice resulting in a mal…

mswiderski:
…formed response

@etirelli since this mainly affects drools parts of the kie server, could you please review?


etirelli:
+1, looks good

PR #384, opened 2016-03-29
BZ-1319546 - test refactoring and extension

sutaakar:
Removed testCommandScript() test method as it is already included in KieServerDroolsIntegrationTest.


mswiderski:
looks good to me, +1 once jenkins build succeed.

PR #390, opened 2016-03-30
BZ-1316966 - OSGi: drools-persistence-jpa throws NPE when starting a process via RuntimeManager

mswiderski:
@winklerm could you please take a look at this? Is this what you had in mind?

PR #393, opened 2016-03-31
[6.4.x] BZ-1316966 - OSGi: drools-persistence-jpa throws NPE when starting a process via RuntimeManager

mswiderski:
back port of #390


mswiderski:
merged into 6.4.x

PR #395, opened 2016-04-01
[6.4.x] [ENTESB-5168] create different features file for Fuse 6.3

psiroky:
* we need different servlet-api bundle for Fuse 6.2
   and Fuse 6.3. Separate feature files seem to be the
   only way to do so. From now on, the feature files
   should include the target Fuse version in the name.
   Keeping features.xml (for Fuse 6.2) because of
   backward compatibility

@mariofusco please take a look

@dvirgiln could you please double check this is what you meant?

@winklerm, @jiripetrlik FYI


dvirgiln:
+1 it is ok!

mariofusco:
+1

PR #397, opened 2016-04-01
[ENTESB-5152] add support for OSGi WAR bundles

psiroky:
* so far we only supported jar-based bundles. WARs
   are, however, supported as well on Karaf/Fuse.
   The code is a hack basically. It tries to search
   search resources with WEB-INF/classes prefix to
   cover the WAR use case. For 7.x we should look
   into making this easier to undersntad, by e.g.
   having two (or more) different classes
   - OsgiJarKieModule and OsgiWarKieModule. They
   would then hide the implementation details
   of WEB-INF/classes, etc. Implementing this for
   6.x is a risk at this stage.

This is a forward port of #368. We need to come up with better approach, bot for now let's at least keep in sync with 6.4.x

### DIFF
--- a/drools-osgi/drools-osgi-integration/pom.xml
+++ b/drools-osgi/drools-osgi-integration/pom.xml
@@ -70,6 +70,21 @@
     </dependency>
 
     <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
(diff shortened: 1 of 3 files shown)


mariofusco:
+1

PR #399, opened 2016-04-03
[6.4.x] [DROOLS-1060] osgi: upgrade Apache POI from 3.11_1 to 3.13_2

psiroky:
* this is just to synp up with non-osgi world which uses
   version 3.13 for quite some time now


mariofusco:
+1

PR #400, opened 2016-04-04
[ENTESB-5168] create different features file for Fuse 6.3

psiroky:
* we need different servlet-api bundle for Fuse 6.2
   and Fuse 6.3. Separate feature files seem to be the
   only way to do so. From now on, the feature files
   should include the target Fuse version in the name.
   Keeping features.xml (for Fuse 6.2) because of
   backward compatibility


mariofusco:
+1

PR #402, opened 2016-04-04
6.4.x: BZ-1310739 - Performance degradation with REST API on WebSphere: JAXB Context CDI use

mrietveld:
For 6.4.x. 

@mswiderski Could you review this? 

This is *not* a blocker: so, please merge this _after_ the release this week.

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
@@ -17,21 +17,12 @@ package org.kie.remote.services.rest.jaxb;
 
 import static org.kie.remote.services.cdi.DeploymentInfoBean.emptyDeploymentId;
 import static org.kie.remote.services.rest.jaxb.DynamicJaxbContextFilter.DEFAULT_JAXB_CONTEXT_ID;
+import static org.kie.remote.services.rest.jaxb.DynamicJaxbContextManager.deploymentClassesMap;
 import static org.kie.services.client.serialization.JaxbSerializationProvider.configureMarshaller;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
-import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.event.Observes;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Marshaller;
@@ -39,20 +30,10 @@ import javax.xml.bind.Unmarshaller;
 import javax.xml.bind.Validator;
 
 import org.jbpm.kie.services.api.DeploymentIdResolver;
-import org.jbpm.kie.services.impl.KModuleDeploymentUnit;
-import org.jbpm.services.api.DeploymentEvent;
-import org.jbpm.services.api.model.DeploymentUnit;
-import org.jbpm.services.cdi.Deploy;
-import org.jbpm.services.cdi.Undeploy;
-import org.kie.remote.services.exception.KieRemoteServicesDeploymentException;
 import org.kie.remote.services.exception.KieRemoteServicesInternalError;
 import org.kie.remote.services.jaxb.ServerJaxbSerializationProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.sun.xml.bind.v2.runtime.IllegalAnnotationException;
-import com.sun.xml.bind.v2.runtime.IllegalAnnotationsException;
-import com.sun.xml.bind.v2.runtime.Location;
 
 /**
  * <h3>READ THIS BEFORE WORKING ON THIS CLASS!</h3>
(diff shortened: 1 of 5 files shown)


kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
```diff
@@ -86,27 +70,22 @@
...
  */
-@ApplicationScoped
+
 public class DynamicJaxbContext extends JAXBContext {
 
     private static final Logger logger = LoggerFactory.getLogger(DynamicJaxbContext.class);
 
     // The contextsCache needs to be a ConcurrentHashMap because parallel operations (involing *different* deployments) can happen on it.
-    private static ConcurrentHashMap<String, JAXBContext> contextsCache = new ConcurrentHashMap<String, JAXBContext>();
+    static ConcurrentHashMap<String, JAXBContext> contextsCache = new ConcurrentHashMap<String, JAXBContext>();
```
mswiderski: maybe better to put this contextsCache in manager and then let this class access manager instead? That way would could avoid static map in this class and make it close to be "pure" JAXBContext impl/extension.

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
```diff
@@ -86,27 +70,22 @@
...
     static {
         setupDefaultJaxbContext();
```
mswiderski: similar for this, if that could be moved to manager class then it could be done as part of @PostConstruct method instead of static block

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContextManager.java
```diff
@@ -0,0 +1,209 @@
...
+@ApplicationScoped
+public class DynamicJaxbContextManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(DynamicJaxbContextManager.class);
+
+    private static AtomicInteger instanceCreated = new AtomicInteger(0);
+
+    public DynamicJaxbContextManager() {
+        if( ! instanceCreated.compareAndSet(0, 1) ) {
+            logger.debug("Instance {} of the {} created!", instanceCreated.incrementAndGet(), DynamicJaxbContext.class.getSimpleName() );
```
mswiderski: does this prevent of creating another class or was it just for debugging?

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContextManager.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    static Map<String, Collection<Class<?>>> deploymentClassesMap = new ConcurrentHashMap<String, Collection<Class<?>>>();
```
mswiderski: does it have to be static map? can't we rely on @ApplicationScoped behavior? or is it accessed from another class...

mswiderski:
@mrietveld looks good in general. Just few minor comments about static fields that seems to be used across classes in same package - can be difficult to manage that long term. Let me know what do you think about these comments as I might have missed some references or concept.

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContextManager.java
mrietveld: Debugging -- sort of. An @ApplicationScoped bean should only be created once -- but sometimes it's not. _If_ that creates problems, I want to see that in the log.

krisv:
We shouldn't postpone this until after our next tag, as that would only increase the risk.  We should try to get this merged today if possoible (or postpone until after 6.3 GA otherwise) imho.

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContextManager.java
mswiderski: in that case maybe increase the log level (info to be same as on 6.3.x branch) or even working otherwise might not be noticed.

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
mrietveld: I disagree with this, seeing as the `contextsCache` field is being used more in the `DynamicJaxbContext` class than in any other class. With regards to clarity of code, it makes more sense to keep it here.

kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContextManager.java
mrietveld: Increased to error, good point.

mrietveld:
Updated so that the `contextsCache` static field is only used in the `DynamicJaxbContext` class.

mrietveld:
@mswiderski I've just updated the code -- if you really feel that the `contextsCache` field should be moved to another class, let me know. I can definitely move it if you feel strongly about it.

mswiderski:
thanks @mrietveld 

as long as the contextsCache is not used across classes then it's all good. will wait for the build to finish after updates and merge. Shall there be PR for master too or we skip it there?

mrietveld:
Thank you, @mswiderski  :)

On master, we'll be deleting `kie-remote-services` (we're talking about that on thursday), so no need to merge to master, I think.

PR #403, opened 2016-04-04
6.3.x: BZ-1310739 - Performance degradation with REST API on WebSphere: JAXB Context CDI use

mrietveld:
For 6.3.x. 

@mswiderski Could you review this? 

I'm pretty sure this should be added to the 6.3.x branch -- however, I'm not sure _when_ this should be merged.  I'll add a comment once I figure it out.

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
@@ -3,7 +3,7 @@
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
  *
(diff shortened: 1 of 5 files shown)


mswiderski:
same comments as for #402 - only static map use and reference them between classes. Other than that looks ok

mrietveld:
@mswiderski The short answer is that I do not think changing the code here with regards to the `contextsCache` static field is a good idea. 

(It has to do with the fact that the jaxb context servlet filter is calling the `org.kie.remote.services.rest.jaxb.DynamicJaxbContext.setDeploymentJaxbContext(String)` method and that the `ThreadLocal` `org.kie.remote.services.rest.jaxb.DynamicJaxbContext.requestJaxbContext` field should be part of the `DynamicJaxbContext` class, not another one. )

However, maybe I've missed another solution? If there's a way to refactor things that you think is better, please let me know. I'll be glad to do that then.

mrietveld:
@mswiderski or @psiroky: Could you merge when you get a chance? 

I got permission from @lazarotti to add to the 6.3.x branch. 
 
Thanks.

PR #404, opened 2016-04-04
Fix compilation error due to Solution interface disappearing.

ge0ffrey:


triceo:
How could this be correct? ```Solution``` is optional - therefore it should not be hardcoded anywhere.

ge0ffrey:
@triceo I 'll refactor that later, but for now the code needs to compile :) It's a good way to test if the legacy way actually works (writing a unit test for that in optaplanner core too)

PR #408, opened 2016-04-05
Fixes after ip-bom 7.0.0.CR1 upgrade

psiroky:


mswiderski:
+1

kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee6-container.xml
```diff
@@ -40,9 +40,6 @@
 
   <dependencySets>
     <dependencySet>
-      <excludes>
-        <exclude>org.jboss.resteasy:*</exclude>
-      </excludes>
```
psiroky: This exclusion (and the other one) were in fact not excluding anything. assembly plugin was complaining about that.

PR #409, opened 2016-04-05
PLANNER-560 stopgap fix random failure of testGetBestSolution()

ge0ffrey:


ge0ffrey:
Random failure still occurs. Looking into it.

PR #411, opened 2016-04-05
[6.4.x] DROOLS-1114 - Add ConversationId http header/jms property to allow dy…

mswiderski:
…namic routing of requests

(cherry picked from commit 7b496c4675f2d909e8aeca569d8548e70fd6341c)

@etirelli here is the conversationId commit for 6.4.x, do you mind reviewing it?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.api;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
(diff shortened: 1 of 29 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
```diff
@@ -0,0 +1,127 @@
...
+            String kieServerId = conversationIdElements[0];
+            String containerId = conversationIdElements[1];
+            ReleaseId releaseId = new ReleaseId(conversationIdElements[2], conversationIdElements[3], conversationIdElements[4]);
+            String uniqueString = conversationIdElements[5];
+
+            return new ConversationId(kieServerId, containerId, releaseId, uniqueString);
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
etirelli: Maciej, the only concern I have is that I don't think we filter/escape ":" in the kieServerId/containerId. If one uses ":" in there, it will break the code. Maybe we should escape any ":" in these components? Or alternatively, wrap them in ""?

etirelli:
+1 to merge.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
mswiderski: this was done based on email from david to keep it as simple as possible. though we can change it as don't think we need to have it 6.4.0.Final community as it needs bit more tests to make sure it's stable. I just made additional fixes as the build failed.

etirelli:
+1 to merge

manstis:
Just a thought does the UI prevent Kie Server or Container names from containing a colon; otherwise is it possible ConversationId parsing could fail? IDK if REST operations to manage servers and containers could equally be at risk?

etirelli:
That was precisely my point @manstis. :) But @mswiderski said he will fix it in the next patch. :)

etirelli:
I merged this as is. Maciej, lets look at how to properly escape ":" in the IDs.

PR #415, opened 2016-04-06
[6.4.x] DROOLS-1114 - Add ConversationId http header/jms property to allow dy…

mswiderski:
…namic routing of requests - changed conversationId format

(cherry picked from commit e8d59b9c521957630af2f2fad1bf9ddfd661692f)

@etirelli @manstis here is a change in the conversationId format to have its parts included in '' so it can use separation character (:) in the name as well. wdyt?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/ConversationId.java
@@ -34,6 +34,9 @@ import org.kie.server.api.model.ReleaseId;
  */
 public class ConversationId {
 
+    private static final String WRAP = "'";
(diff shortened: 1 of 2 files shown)


manstis:
@mswiderski Well, IMO, it makes the issue possibly less likely to happen but only fixes the symptom not the cause... to prevent the cause we'd need validation of KIE Server ID and Container ID presumably in both the UI and REST calls (*if* there's REST to create/rename etc these items.. IDK).. Of course the fix for that is significantly more than this and it may be considered to be suitably low risk to not worry about.. and on that basis your "improved demarcation" lowers the chance of an issue so +1.

etirelli:
I think it looks good Maciej. +1 to merge.

PR #416, opened 2016-04-06
[BZ-1324017] fix dealing with URL suffix on Windows

psiroky:
* TLDR; URLs can only contain forward slashes - '/' and
   filesystem pathas contain os-specific hashes. The code
   needs to deal with URLs and paths slightly differently!

### DIFF
--- a/kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
+++ b/kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
@@ -58,7 +58,11 @@ public class KModuleBeanFactoryPostProcessor implements BeanFactoryPostProcessor
 
     private static final Logger log = LoggerFactory.getLogger(KModuleBeanFactoryPostProcessor.class);
 
-    private static final String WEB_INF_CLASSES_FOLDER = "WEB-INF" + File.separator + "classes" + File.separator;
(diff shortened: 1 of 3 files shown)


kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
```diff
@@ -271,47 +279,47 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
-    private URL tryGetRootUrlForEapContext(ApplicationContext applicationContext) throws IOException {
-        Enumeration<URL> rootUrls = applicationContext.getClassLoader().getResources("/");
-        if (containsEapSpecificUrl(rootUrls)) {
-            while (rootUrls.hasMoreElements()) {
-                URL url = rootUrls.nextElement();
-                if (url.toString().endsWith(WEB_INF_CLASSES_FOLDER)) {
-                    return url;
-                }
+    URL tryGetRootUrlForEapContext(Enumeration<URL> rootUrls) {
+        boolean containsEapSpecificUrl = false;
```
psiroky: @mariofusco do we need to check for the "EAP-specific" URL? It seems that we could just search for the WEB-INF/classes URL and use that (but I may be overlooking something).

mariofusco:
+1

PR #418, opened 2016-04-06
DROOLS-1114 - Add ConversationId http header/jms property to allow dy…

mswiderski:
…namic routing of requests - changed conversationId format


mswiderski:
jenkins retest this

PR #419, opened 2016-04-07
Do not merge yet: PLANNER-560 stopgap fix random failure of testGetBestSolution()

ge0ffrey:
We'll also need @sutaakar 's +1 on this before merging.


sutaakar:
@ge0ffrey Seems that I found out why tests fails.
It is good old marshalling issue.
When best solution isn't yet initialized then Kie server returns empty score, for example in JAXB it is "<score/>". But it is unmarshalled into HardSoftScore with values -2147483648hard/-2147483648soft.

mswiderski:
looks like it still suffers from the same instability issue, shall this be kept open?

ge0ffrey:
Let's close it for now (I will close and the other one). I need to fix the initialized score problem first, I hope to have time for that in 3 weeks.

PR #422, opened 2016-04-08
PLANNER-565 - Marshall best solution score test

sutaakar:
Showing issues described in referenced JIRA.

### DIFF
--- a/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/ScoresMarshallingTest.java
+++ b/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/ScoresMarshallingTest.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.api.marshalling;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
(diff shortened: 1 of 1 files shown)


mswiderski:
@ge0ffrey this looks like problem with*Score classes that cannot be properly marshaled as the fields are final and can be set only via constructor. This is not really valid for JAXB that uses either field access of setters. So if I am not mistaken this has to be changed on optaplanner side to be JAXB/JSON friendly - either by providing setters/getters or alternative way to be set - like JAXB adapters.

ge0ffrey:
Ok, we need optaplanner-persistence-jaxb :)
btw, we already have optaplanner-persistence-xstream, but the exe server doesn't use it yet (it should)

mswiderski:
@ge0ffrey are you going to build the optaplanner-persistence-jaxb module? Otherwise I don't think we have any option to solve these issues...

ge0ffrey:
@mswiderski Yes - but it's currently not in my short time list and only aimed for 7.0. How high is the urgency on this?

mswiderski:
@ge0ffrey I guess it's mainly up to you to asses as it might limit usage of optaplanner over kie server.

mswiderski:
do we still need this/plan to merge?

sutaakar:
Rebased to contain latest changes. There are still issues with JSON marshalling for bendable score. Should it be supported?

ge0ffrey:
These tests should move into optaplanner-persistence-*, which contains already some:
there are already xstream and jaxb tests, for each score implementation.

JSON support for Score's is still an open issue...

sutaakar:
Right, will prepare a PR there.

ge0ffrey:
Interestingly enough, this one tests the marshalling of SolverInstance too. So this might prove or disprove PLANNER-604. SolverInstance is a Execution Server object, so we can't do that test in optaplanner core.

@sutaakar Karel, let's talk on irc/bluejeans about this PR and #425

ge0ffrey:
Relates to PLANNER-612

mswiderski:
any plans to merge this or...

ge0ffrey:
Sorry Matej for losing track of this, thanks for the reminder. Looking at it.

ge0ffrey:
jenkins retest this

ge0ffrey:
This open PR from Jun with failing tests in Jun might now have green tests thanks to this merged PR from October:
  https://github.com/droolsjbpm/droolsjbpm-integration/pull/695/files

I still think that testing the different score types here is pointless (= maintenance), as we're already doing that in optaplanner-persistence for both XStream, JAXB and Jackson for each score type. for example:
  https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-persistence/optaplanner-persistence-jaxb/src/test/java/org/optaplanner/persistence/jaxb/api/score/buildin/hardsoft/HardSoftScoreJaxbXmlAdapterTest.java

However testing one SolverInstance with a HardScore marshaling/unmarshalling is probably useful to have 

@mcimbora Wdyt?

sutaakar:
Refactored to check marshalling of SolverInstance with a HardSoftScore.

PR #427, opened 2016-04-11
make a RemoteKieServerOperation class protected instead of public

mswiderski:
@porcelli @etirelli @csadilek see the change ... I might be stupid and don't understand this but how it can work on one wildfly version and not the other....????


etirelli:
@mswiderski I will let you know once we get the acks to merge, but +1 to the change.

PR #429, opened 2016-04-13
[DROOLS-1075] allow to define and execute batch commands also on stateless ksession

mariofusco:


mswiderski:
+1

the failure is on jbpm side and is unrelated to this change but looks like something changed on kie-ci level MavenRepository.deployArtifact, I am going to look at it now...

mswiderski:
jenkins retest this

PR #430, opened 2016-04-14
DROOLS-1114 - uncomment tests

sutaakar:
Uncommented accidentally commented tests.


mswiderski:
+1 as it's my mistake I believe we need this for 6.4.x too

PR #432, opened 2016-04-15
DROOLS-1128 - Concurrent call fails when creating container

sutaakar:
Also refactored conversation Id integration tests: moved conversation Id check to separated test method in common module.


mswiderski:
+1

PR #437, opened 2016-04-19
kie-server: test improvements and stabilization

sutaakar:
- raised timeout for QueryDataServiceIntegrationTest and OptaPlanner integration tests, current ones were low for case when is used slower repo(improvement idea - separate API to reduce to reduce dependency tree complexity and size)
- updated JobServiceIntegrationTest - to remove dependency on job order returned by Kie server
- updated ProcessServiceIntegrationTest - to remove dependency on job order returned by Kie server, also some DBs round timestamp so test became unreliable for them

- response code changes in REST tests - some containers like WebLogic 12.1.3 use jersey 1.x. In such case response code for DELETE requests from Kie server is 200 instead of 204. It seems to be caused by setting of response entity for these requests, see http://stackoverflow.com/questions/31655486/jersey-nocontent-response-returns-200-instead-of-204 .
Is this something what should be fixed? Or it can be acceptable behavior.


mswiderski:
@sutaakar could you please take a look at the test failures as they seem related...

sutaakar:
Sure, these errors look really weird.

mswiderski:
jenkins retest this

PR #439, opened 2016-04-19
BZ-1319553 - Response header for kie-server call always returns appli…

mswiderski:
…cation/xml

@mrietveld could you please take a look as it changes slightly the use of getVariant which is used by kie-remote as well

@etirelli @sutaakar comments are welcome from your end too

### DIFF
--- a/kie-remote/kie-remote-common/src/main/java/org/kie/remote/common/rest/RestEasy960Util.java
+++ b/kie-remote/kie-remote-common/src/main/java/org/kie/remote/common/rest/RestEasy960Util.java
@@ -55,14 +55,15 @@ public class RestEasy960Util {
                 fixedAcceptHeaders.add(header.replaceAll("q=\\.", "q=0.")); 
             }
             acceptHeaders = fixedAcceptHeaders;
-        }
-        negotiation.setAcceptHeaders(acceptHeaders);
-        negotiation.setAcceptCharsetHeaders(requestHeaders.get(ACCEPT_CHARSET));
-        negotiation.setAcceptEncodingHeaders(requestHeaders.get(ACCEPT_ENCODING));
-        negotiation.setAcceptLanguageHeaders(requestHeaders.get(ACCEPT_LANGUAGE));
+            negotiation.setAcceptHeaders(acceptHeaders);
+            negotiation.setAcceptCharsetHeaders(requestHeaders.get(ACCEPT_CHARSET));
+            negotiation.setAcceptEncodingHeaders(requestHeaders.get(ACCEPT_ENCODING));
+            negotiation.setAcceptLanguageHeaders(requestHeaders.get(ACCEPT_LANGUAGE));
 
-        return negotiation.getBestMatch(variants);
-        // ** use below instead of above when RESTEASY-960 is fixed **
-        // return restRequest.selectVariant(variants); 
+            return negotiation.getBestMatch(variants);
+            // ** use below instead of above when RESTEASY-960 is fixed **
+            // return restRequest.selectVariant(variants);
+        }
+        return null;
     }
 }
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
@@ -42,14 +42,14 @@ public enum MarshallingFormat {
     }
 
     public static MarshallingFormat fromType( String type ) {
-        if ("xstream".equalsIgnoreCase( type )) {
+        if ("xstream".equalsIgnoreCase( type ) || "application/xstream".equalsIgnoreCase( type ) ) {
             return XSTREAM;
-        } else if ("xml".equalsIgnoreCase( type )) {
+        } else if ("xml".equalsIgnoreCase( type ) || "application/xml".equalsIgnoreCase( type ) ) {
             return JAXB;
-        } else if ("json".equalsIgnoreCase( type )) {
+        } else if ("json".equalsIgnoreCase( type ) || "application/json".equalsIgnoreCase( type ) ) {
             return JSON;
         } else {
-            return null;
+            return valueOf(type);
         }
     }
 }
--- a/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
@@ -50,8 +50,6 @@ import org.kie.server.controller.api.model.spec.ServerTemplateList;
 
 public class ControllerUtils {
 
-    public static Variant defaultVariant = Variant.mediaTypes(MediaType.APPLICATION_JSON_TYPE).add().build().get(0);
-
     private static Marshaller jsonMarshaller = MarshallerFactory.getMarshaller(getMinimalModelClasses(), MarshallingFormat.JSON, ControllerUtils.class.getClassLoader());
     private static Marshaller jaxbMarshaller = MarshallerFactory.getMarshaller(getModelClasses(), MarshallingFormat.JAXB, ControllerUtils.class.getClassLoader());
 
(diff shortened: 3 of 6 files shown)


etirelli:
@mswiderski it seems jenkins is failing some tests.

mswiderski:
@etirelli correct, 2 of them are related to this PR and 26 others are not as other PRs failed on them too. Trying to figure out what might be the cause of it

sutaakar:
looks ok for me

mswiderski:
ok the root cause is this:
```
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:179) ~[maven-core-3.2.5.jar:3.2.5]
	... 41 common frames omitted
Caused by: java.lang.ExceptionInInitializerError: null
	at org.drools.compiler.compiler.ProcessBuilderFactory.initializeProvider(ProcessBuilderFactory.java:47) ~[drools-compiler-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.drools.compiler.compiler.ProcessBuilderFactory.<clinit>(ProcessBuilderFactory.java:27) ~[drools-compiler-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.kie.maven.plugin.BuildMojo.execute(BuildMojo.java:124) ~[na:na]
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:132) ~[maven-core-3.2.5.jar:3.2.5]
	... 41 common frames omitted
Caused by: java.lang.RuntimeException: Unable to build kie service url = jar:file:/Users/maciejswiderski/.m2/repository/org/optaplanner/optaplanner-core/7.0.0-SNAPSHOT/optaplanner-core-7.0.0-SNAPSHOT.jar!/META-INF/kie.conf
	at org.drools.core.util.ServiceDiscoveryImpl.discoverFactories(ServiceDiscoveryImpl.java:57) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.kie.internal.utils.ServiceRegistryImpl.initServiceDiscovery(ServiceRegistryImpl.java:228) ~[kie-internal-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.kie.internal.utils.ServiceRegistryImpl.init(ServiceRegistryImpl.java:217) ~[kie-internal-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.kie.internal.utils.ServiceRegistryImpl.<init>(ServiceRegistryImpl.java:66) ~[kie-internal-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.kie.internal.utils.ServiceRegistryImpl.<clinit>(ServiceRegistryImpl.java:54) ~[kie-internal-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	... 45 common frames omitted
Caused by: java.lang.ClassCastException: org.optaplanner.core.impl.solver.kie.KieSolverAssemblerService cannot be cast to org.kie.internal.assembler.KieAssemblerService
	at org.drools.core.util.ServiceDiscoveryImpl.processKieAssemblers(ServiceDiscoveryImpl.java:113) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.drools.core.util.ServiceDiscoveryImpl.processKieConf(ServiceDiscoveryImpl.java:85) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.drools.core.util.ServiceDiscoveryImpl.processKieConf(ServiceDiscoveryImpl.java:79) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.drools.core.util.ServiceDiscoveryImpl.processKieConf(ServiceDiscoveryImpl.java:74) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
	at org.drools.core.util.ServiceDiscoveryImpl.discoverFactories(ServiceDiscoveryImpl.java:55) ~[drools-core-7.0.0-SNAPSHOT.jar:7.0.0-SNAPSHOT]
```

and this was introduced by this commit: 
https://github.com/droolsjbpm/drools/commit/9940af3d917639313c51940b690fb9c3993c057e

@mariofusco @ge0ffrey is this org.optaplanner.core.impl.solver.kie.KieSolverAssemblerService really in use? Looking at the class itself it does not do much...
https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/kie/KieSolverAssemblerService.java
 so if it's not in use it should not be declared in kie.conf file as it now breaks kie-maven-plugin usage, most likely this file https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-core/src/main/resources/META-INF/kie.conf should be removed from optaplanner for now...

once it's removed kie server tests are back to normal

@etirelli @psiroky @sutaakar FYI

sutaakar:
I have noticed one thing when I was playing with this PR. 
I set Accept header to "application/json,application/xml" as written by Alessandro in his last comment in BZ. Then if I have Content-Type set as "application/json" I get answer marshalled as JSON, but Content-Type header in response has value "application/xml".

You can simulate it in JbpmRestIntegrationTest, just add mentioned accept header to request.

kie-remote/kie-remote-common/src/main/java/org/kie/remote/common/rest/RestEasy960Util.java
```diff
@@ -55,14 +55,15 @@ public static Variant getVariant(HttpHeaders headers) {
                 fixedAcceptHeaders.add(header.replaceAll("q=\\.", "q=0.")); 
             }
             acceptHeaders = fixedAcceptHeaders;
-        }
```
mrietveld: Good catch, @mswiderski. I didn't see the NPE until I looked at the [ServerDrivenNegotiation method](https://github.com/mswiderski/droolsjbpm-integration/blob/BZ-1319553/kie-remote/kie-remote-common/src/main/java/org/kie/remote/common/rest/variant/ServerDrivenNegotiation.java#L136).

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
```diff
@@ -42,14 +42,14 @@ public static MarshallingFormat fromId( int id ) {
     }
 
     public static MarshallingFormat fromType( String type ) {
-        if ("xstream".equalsIgnoreCase( type )) {
+        if ("xstream".equalsIgnoreCase( type ) || "application/xstream".equalsIgnoreCase( type ) ) {
             return XSTREAM;
-        } else if ("xml".equalsIgnoreCase( type )) {
+        } else if ("xml".equalsIgnoreCase( type ) || "application/xml".equalsIgnoreCase( type ) ) {
```
mrietveld: Maybe use `MediaType.APPLICATION_XML` and `MediaType.APPLICATION_JSON`? Not necessary though, just cleaner.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
```diff
@@ -106,15 +104,19 @@ public static Response createCorrectVariant(String responseObj, HttpHeaders head
     public static Variant getVariant(HttpHeaders headers) {
         Variant v = RestEasy960Util.getVariant(headers);
         if( v == null ) {
-            v = defaultVariant;
+            v = Variant.mediaTypes(getMediaType(headers)).add().build().get(0);;
```
mrietveld: This seems a convoluted way to do the same thing? But I'm probably missing the whole reason that you did this, which fixes the bug, I'm assuming. :)

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
```diff
@@ -118,6 +122,28 @@ public static String getContentType(HttpHeaders headers) {
         return contentType;
     }
 
+    public static MediaType getMediaType(HttpHeaders httpHeaders) {
```
mrietveld: Isn't this the exact same method as in the `ControllerUtils` class? If not, ignore me! :)

mrietveld:
+1 Once tests complete. 

Up to @mswiderski if he wants to change anything based on my comments: there are things _I_ would change, but they're subjective.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
mswiderski: good point, will change that for xml and json

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
mswiderski: this is quite different as it generates Variant based on given request header - to be more specif Content-Type header instead of using the default so it will match the content with the header - in case Accept is not given at all.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
mswiderski: it is, and it's duplicated to avoid additional dependency (kie-remote-common) just for one or two methods as it would bring quite few transitive deps

mswiderski:
@sutaakar thanks for reminding this, will add a test and fix the issue

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
mswiderski: now I know I didn't add that - again due to dependency - it's kie-server-api so it's not depending on and "rest" stuff so I don't want to add it here just to avoid too much being added

mswiderski:
@sutaakar ok, added tests for accept headers, one important note is that whenever specifying multiple accept formats you need to use quality parameter on them to let server know about preferences, if not given any can be returned as server assumes it does not really matter.

While we still could have then mismatch between actual content and the returned header it's because we use same marshaller type for both request and response.

So the recommendation is use only single accept header or with quality parameter to ensure consistency. How does that sound?

ge0ffrey:
@mswiderski 
the KieSolverAssemblerService.java in optaplanner-core is needed, even though it does nothing itself. It is needed to make the KieContainer based tests work properly, so that ".java" files in the kjar (not ".class" files) actually get found by solver.xml and drl files as .class files.

ge0ffrey:
If that KieSolverAssemblerService isn't there, KieContainerSolverFactoryTest fails.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
mrietveld: Ah, that's a really good reason! :) Thanks for explaining.

mswiderski:
jenkins retest this

mswiderski:
any other comments or we good at merging this?

sutaakar:
looks good for me

PR #441, opened 2016-04-20
JBPM-5088 - KIE-Server process images are not found when stored insid…

mswiderski:
…e packages

@etirelli this could be important for the product too


mswiderski:
jenkins retest this

PR #442, opened 2016-04-20
BZ-1326805 - Serialization problems on websphere as a result of BZ-1299549

mrietveld:
See https://bugzilla.redhat.com/show_bug.cgi?id=1326805

For the reviewer: both this pr and https://github.com/droolsjbpm/kie-wb-distributions/pull/268 are necessary for this. 

The `org.apache.wink.jaxbcontextcache` property (set to `off`) makes sure that JAXB contexts are not cached (otherwise, changes in the deployment would not be "caught" and user-defined serialized class instances _incorrectly_ de/serialized). 

This change makes sure that the (apache wink/websphere) JAX-RS implementation framework gets a real `JAXBContext` instance: it looks like the openwebbeans CDI implementation that websphere uses simply interferes too much for us to use the `DynamicJaxbContext` as the `JAXBContext` instance. 

Lastly, given that the `DynamicJaxbContext` is already caching `JAXBContext` instances, there should be no significant performance penalty by turning off the apache wink `JAXBContext` caching.

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
@@ -125,6 +125,10 @@ public class DynamicJaxbContext extends JAXBContext {
 
     // JAXBContext methods --------------------------------------------------------------------------------------------------------
 
+    JAXBContext getRequestContext() {
+        return requestJaxbContext.get();
+    }
+
     /*
      * (non-Javadoc)
      * @see javax.xml.bind.JAXBContext#createUnmarshaller()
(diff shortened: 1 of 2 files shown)


mrietveld:
@mswiderski or @psiroky, Would you mind reviewing and merging _ASAP_?

mswiderski:
just one question - do we need to distinguish if it's websphere or not?

if we do return that concrete instance on all app servers they will cache that single instance and use it for all deployments or?

mrietveld:
AFAIK, we do ned to distinguish if it's websphere. Basically, my thoughts are this: 

1. The current code (the way it was) works on all the other servers. 
2. I can't remember how the caching worked on other servers, but I'm pretty sure that it is cached -- otherwise, I wouldn't have created this in the first place. It's well known that `JAXBContext` creation is an expensive operation and most JAX-* frameworks try to minimize this -- from what I know. 

Does that make sense?

mswiderski:
yup, thanks for explanation

PR #443, opened 2016-04-20
BZ-1326806 - Serialization problems on websphere as a result of BZ-1310739

mrietveld:
See https://bugzilla.redhat.com/show_bug.cgi?id=1326806

For the reviewer: both this pr and https://github.com/droolsjbpm/kie-wb-distributions/pull/268 are necessary for this. 

The `org.apache.wink.jaxbcontextcache` property (set to `off`) makes sure that JAXB contexts are not cached (otherwise, changes in the deployment would not be "caught" and user-defined serialized class instances _incorrectly_ de/serialized). 

This change makes sure that the (apache wink/websphere) JAX-RS implementation framework gets a real `JAXBContext` instance: it looks like the openwebbeans CDI implementation that websphere uses simply interferes too much for us to use the `DynamicJaxbContext` as the `JAXBContext` instance. 

Lastly, given that the `DynamicJaxbContext` is already caching `JAXBContext` instances, there should be no significant performance penalty by turning off the apache wink `JAXBContext` caching.

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/rest/jaxb/DynamicJaxbContext.java
@@ -93,7 +93,7 @@ public class DynamicJaxbContext extends JAXBContext {
      * create a {@link Marshaller}, {@link Unmarshaller} or {@link Validator} instance.
      * @return The {@link JAXBContext} created or retrieved from cache for the request.
      */
-    private JAXBContext getRequestContext() {
+    JAXBContext getRequestContext() {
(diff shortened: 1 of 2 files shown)


mrietveld:
@mswiderski or @psiroky, would you mind reviewing and merging, _ASAP_?

mswiderski:
one test failed but I believe it's not related to this PR...

mrietveld:
Correct, the failed test is related to https://github.com/droolsjbpm/droolsjbpm-integration/commit/1d6e9f3e20d8a16ee2cabe5aa2c689a6c6c9ed05 which needs to be cherry-picked to 6.4.x.

mswiderski:
you mean to 6.3.x? will merge it now

psiroky:
See https://github.com/droolsjbpm/drools/pull/751 which should fix the test.

PR #444, opened 2016-04-20
Added optaplanner-core to kie-maven-plugin dependencies to avoid clas…

mswiderski:
…s cast exception on assembler when building projects

@mariofusco @psiroky could you please take a look


mariofusco:
+1

psiroky:
Would be possible to add a test for this?

mswiderski:
no idea to be honest... can it be done?? since I don't know planner or maven plugin good enough I don't know if we can do it. And we have test in kie server integration tests :)

psiroky:
OK. Let me merge this and I will take a look at creating the test later (since we need to fix those kie-server tests ASAP).

PR #447, opened 2016-04-20
kie-server-tests: run modules one at a time, even for parallel builds

psiroky:
@sutaakar, @mswiderski please take a look. This is basically hack/workaround to make sure the test modules are not executed in parallel. Ultimately we could fix this by configuring cargo to use random available ports instead of the defaults. But that will take some time to get properly done. This is I think a good workaround for now.


mswiderski:
hmm interesting as long as it works then you have my +1

PR #449, opened 2016-04-20
[6.4.x] JBPM-5026 - Not possible to change UserInfo in KieServer - add test

jakubschwan:


mswiderski:
+1 tests only changes

PR #450, opened 2016-04-21
Remove duplicated dependency

psiroky:
* was introduced by mistake via
   3a92c621a2e400a


mswiderski:
+1

psiroky:
Please keep this one open until the build succeeds. I've switched the PR builds to use WildFly 10 so would like to see if everything works as expected.

mswiderski:
so we now have wildfly 10 for kie server tests?

psiroky:
Yes, I've switched from Tomcat 8 to WildFly 10 on all the places I could think of (in the context of Jenkins). Only on master of course.

mswiderski:
excellent!

PR #455, opened 2016-04-25
kie-server-tests: use Cargo's artifact-installer instead of url-installer

psiroky:
* artifact-installer allows better caching (inside ~./m2)

 * the Tomcat download URLs were also causing issues on JDK 6
   (Could not generate DH keypair: Prime size must be multiple of 64,
   and can only range from 512 to 1024 (inclusive))

@mswiderski, @sutaakar please take a look.

### DIFF
--- a/kie-server-parent/kie-server-tests/pom.xml
+++ b/kie-server-parent/kie-server-tests/pom.xml
@@ -35,10 +35,10 @@
          but for testing e.g. stated builds it is necessary to pass additional repository with the staged artifacts. -->
     <kie.server.testing.kjars.build.settings.xml/>
 
-    <tomcat7x.download.url>https://archive.apache.org/dist/tomcat/tomcat-7/v7.0.55/bin/apache-tomcat-7.0.55.zip</tomcat7x.download.url>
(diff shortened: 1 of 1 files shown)


mswiderski:
+1 once the build is green - assuming all other builds (other containers) will work too - maybe we could run the matrix build to test other containers?

weblogic and websphere remain unchanged, right?

psiroky:
@mswiderski it would be great to run matrix builds. This repo contains more than just kie-server stuff. So running the matrix with every PR would be an overkill. What we could do is to somehow try to detect if kie-server is being changed and then run the matrix job. Or we could add a possibility to trigger the whole matrix job manually (e.g. via comment).

mswiderski:
@psiroky I meant only to run the build on this specific PR or manually on given branch just to make sure we don't have an issue, but I guess it's not as important, thanks

any +1 :)

sutaakar:
+1

psiroky:
OK, let me run the matrix job on this branch. Should be quite easy.

psiroky:
The matrix job finished successfully (all 10 combinations), so we should be fine.

mswiderski:
excellent, go ahead and merge it, thanks a lot @psiroky

PR #458, opened 2016-04-26
JBPM-5083 - JMS Kie server startup test

sutaakar:
@mswiderski Reproducer for JBPM-5083 and JBPM-5098

Maybe cargo-core-uberjar artifact should be moved elsewhere, to have its version managed and synchronized with Cargo maven plugin. Will check what and how.

Test works for Wildfly10 and WebLogic 12.2.
Cargo WebSphere implementation is still missing remote deployer - skipping test for WebSphere temporary.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/pom.xml
@@ -115,7 +115,13 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
-    
+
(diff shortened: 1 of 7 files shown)


sutaakar:
Update: seems that Wildfly 10 turns off queues defined in Kie server when it is undeployed - need to find workaround.

sutaakar:
Update: it should work for Wildfly 10 and WLS again.

sutaakar:
Rebased due to merge conflicts.

mswiderski:
replaced by https://github.com/droolsjbpm/droolsjbpm-integration/pull/461 that includes both the test case and the fix, closing this one

PR #461, opened 2016-04-26
Several KIE Server fixes that are related to each other

mswiderski:
- JBPM-5083
- JBPM-5098
- DROOLS-1122 
- DROOLS-1138

@sutaakar thanks a lot for the tests, great work!

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/ContainerManager.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/ContainerManager.java
@@ -30,6 +30,10 @@ public class ContainerManager {
     private static final Logger logger = LoggerFactory.getLogger(ContainerManager.class);
 
     public void installContainers(KieServerImpl kieServer, Set<KieContainerResource> containers, KieServerState currentState, KieServerSetup kieServerSetup) {
+        installContainersSync(kieServer, containers, currentState, kieServerSetup);
+    }
+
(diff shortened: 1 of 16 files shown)


kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/ContainerManager.java
```diff
@@ -30,6 +30,10 @@
     private static final Logger logger = LoggerFactory.getLogger(ContainerManager.class);
 
     public void installContainers(KieServerImpl kieServer, Set<KieContainerResource> containers, KieServerState currentState, KieServerSetup kieServerSetup) {
+        installContainersSync(kieServer, containers, currentState, kieServerSetup);
+    }
+
+    public void installContainersSync(KieServerImpl kieServer, Set<KieContainerResource> containers, KieServerState currentState, KieServerSetup kieServerSetup) {
```
sutaakar: installContainersSync seems to do same thing as installContainers. Is there any reason to have both methods here?

mswiderski: the difference is here https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/ContainerManagerEJB.java#L35 which is used on app servers do it can be done in async way but still valid for JEE environments

sutaakar: Got it, thanks for info :)

PR #465, opened 2016-05-06
JBPM-5071 - Clean up and remove the process runtimes from the workbench

mswiderski:
this is as part of JBPM-5071 - Clean up and remove the process runtimes from the workbench

to ease collaboration some work that does not affect other components that forked would be good to be placed on master as they are functional as standalone

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/LoadBalancer.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/LoadBalancer.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.client.balancer;
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+
+import org.kie.remote.common.rest.KieRemoteHttpRequest;
+import org.kie.remote.common.rest.KieRemoteHttpRequestException;
+import org.kie.server.client.balancer.impl.RandomBalancerStrategy;
+import org.kie.server.client.balancer.impl.RoundRobinBalancerStrategy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoadBalancer {
+
+    private static final Logger logger = LoggerFactory.getLogger(LoadBalancer.class);
+
+    private ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+
+    private final BalancerStrategy balancerStrategy;
+    private CopyOnWriteArraySet<String> failedEndpoints = new CopyOnWriteArraySet<String>();
+
+    protected LoadBalancer(BalancerStrategy balancerStrategy) {
+        this.balancerStrategy = balancerStrategy;
+    }
+
+    public String getUrl() throws KieRemoteHttpRequestException{
+        String selectedUrl = balancerStrategy.next();
+        logger.debug("Load balancer {} selected url '{}'", balancerStrategy, selectedUrl);
+        return selectedUrl;
+    }
+
+    public void markAsFailed(String url) {
+        failedEndpoints.add(url);
+        balancerStrategy.markAsOffline(url);
+        logger.debug("Url '{}' is marked as failed and will be considered offline by {}", url, balancerStrategy);
+    }
+
+    public void activate(String url) {
+        failedEndpoints.remove(url);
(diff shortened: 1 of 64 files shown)


kie-security/src/main/java/org/kie/security/util/Base64Util.java
```diff
@@ -0,0 +1,212 @@
...
+ * I am placing this code in the Public Domain. Do with it as you will. This software comes with no guarantees or warranties but
+ * with plenty of well-wishing instead! Please visit <a href="http://iharder.net/base64">http://iharder.net/base64</a>
+ * periodically to check for updates or to contribute improvements.
+ * </p>
+ *
+ * @author Robert Harder
+ * @author rob@iharder.net
+ * @version 2.3.7
+ */
+public class Base64Util {
```
psiroky: Could we use the Java's https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html instead?

mswiderski: yes, that's very good point, thanks @psiroky

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/LoadBalancer.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /*
+     * factory methods
+     */
+
+    public static LoadBalancer getDefault(String urls) {
+        String[] endpoints = new String[0];
+        if (urls != null) {
+            endpoints = urls.split("\\|");
```
rsynek: Could we rather define the delimiter char as a constant to prevent its repetition (see `forStrategy()` method)?

kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/LoadBalancerClientTest.java
```diff
@@ -0,0 +1,224 @@
...
+    private WireMockServer wireMockServer2;
+    private WireMockServer wireMockServer3;
+
+    private KieServicesConfiguration config;
+
+    @Before
+    public void startServers() {
+        int port1 = BaseKieServicesClientTest.findFreePort();
+        wireMockServer1 = new WireMockServer(port1);
+        wireMockServer1.stubFor(get(urlEqualTo("/"))
```
rsynek: Would it be possible to move mock creation to a new method? If I am not mistaken, all three mock instances are nearly the same.

mswiderski: done, moved to single method, thanks

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/LoadBalancer.java
mswiderski: done

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/impl/RandomBalancerStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void markAsOnline(String url) {
+        synchronized (availableEndpoints) {
+            availableEndpoints.add(url);
+        }
+    }
+
+    protected int getRandomInt(int min, int max) {
+        return (int) Math.floor(Math.random() * (max - min + 1)) + min;
```
jhrcek: It is generally preferable to use something like `min + random.nextInt(max - min + 1)`. See e.g. this [SO answer](http://stackoverflow.com/questions/738629/math-random-versus-random-nextintint)

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java
```diff
@@ -50,11 +50,31 @@ public String getProcessForm(String containerId, String processId, String langua
             valuesMap.put(RestURI.PROCESS_ID, processId);
 
             return makeHttpGetRequestAndCreateRawResponse(
-                    build(baseURI, FORM_URI + "/" + PROCESS_FORM_GET_URI, valuesMap) + "?lang=" + language);
+                    build(loadBalancer.getUrl(), FORM_URI + "/" + PROCESS_FORM_GET_URI, valuesMap) + "?lang=" + language);
```
sutaakar: Seems that there is missing "filter" parameter.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java
```diff
@@ -70,11 +90,31 @@ public String getTaskForm(String containerId, Long taskId, String language) {
             valuesMap.put(RestURI.TASK_INSTANCE_ID, taskId);
 
             return makeHttpGetRequestAndCreateRawResponse(
-                    build(baseURI, FORM_URI + "/" + TASK_FORM_GET_URI, valuesMap) + "?lang=" + language);
+                    build(loadBalancer.getUrl(), FORM_URI + "/" + TASK_FORM_GET_URI, valuesMap) + "?lang=" + language);
```
sutaakar: Same here, seems that there is missing "filter" parameter.

mswiderski:
jenkins retest this

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/balancer/impl/RandomBalancerStrategy.java
mswiderski: sure we could do this but this strategy is more of a demo one rather than a real production ready strategy. So not sure there is need for performance improvements here.

PR #469, opened 2016-05-13
JBPM-5114 - Nested Map<String, Map<String, Object>[]> objects can not be de/serialized correctly

mrietveld:
https://issues.jboss.org/browse/JBPM-5114

This use case is particular important for migration ( @ibek, Could you add a comment about that? Thanks!) 

List of PRs: 

- https://github.com/droolsjbpm/drools/pull/782
- https://github.com/droolsjbpm/droolsjbpm-integration/pull/469 (contains test)


mswiderski:
+1 to merge

ibek:
@mrietveld after all it's not so urgent - we used Map<String, List<Map<String,String>>> as the alternative though I had to change some code which was operating with the array.

Thank you for the immediate fix.

When a developer designs a form/task which enables to enter list of properties kept in a map, he can encounter the same problems when remote API is used.

PR #470, opened 2016-05-13
Jetty + CXF related fixes after upgrade to IP BOM 7.0.0.CR3

psiroky:
* Jetty was upgraded from 8.x to 9.x and those two
   versions are not 100% compatible.

Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/201

@mrietveld could ouy please take a look?

### DIFF
--- a/kie-remote/kie-remote-common/src/test/java/org/kie/remote/common/rest/KieRemoteHttpRequestTest.java
+++ b/kie-remote/kie-remote-common/src/test/java/org/kie/remote/common/rest/KieRemoteHttpRequestTest.java
@@ -596,11 +596,7 @@ public class KieRemoteHttpRequestTest extends ServerTestCase {
             public void handle( Request request, HttpServletResponse response ) {
                 String auth = request.getHeader("Authorization");
                 auth = auth.substring(auth.indexOf(' ') + 1);
-                try {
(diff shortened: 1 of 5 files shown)


kie-remote/kie-remote-common/src/test/java/org/kie/remote/common/rest/KieRemoteHttpRequestTest.java
```diff
@@ -859,7 +855,7 @@ public void handle( Request request, HttpServletResponse response ) {
             }
         };
         Map<String, List<String>> headers = newRequest(url).get().response().headers();
-        assertEquals(headers.size(), 5);
+        assertEquals(headers.size(), 6);
```
mrietveld: @psiroky Any clue why the number of headers is now 6?

psiroky: Jetty 9.x includes one more default header for some reason, but I forgot which one :) I vaguely remember it was a timestamp of the request (or response?), but can look it up to be sure.

mrietveld:
+1 Thanks!

PR #471, opened 2016-05-16
[6.3.x][BZ-1322171] drools-osgi: upgrade XStream bundle from 1.4.7_1 to 1.4.9_1

psiroky:


psiroky:
@mbiarnes not sure what I did wrong :), but simply upgrading the version seems to work. Could you please try again locally to see if you could reproduce those test failures again?

mbiarnes:
+1 - tested locally and it works. no test failures.

PR #478, opened 2016-05-27
[RHBRMS-2461] adapt the KieContainerResolver to the KieContainer interface

mariofusco:


winklerm:
Tested with the standalone reproducer, the fix works! 

Should the other subclasses of AbstractKieObjectsResolver be updated as well?

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/ea36ac8b28498db56ad8291fc0e11a839d476b86

PR #481, opened 2016-06-02
kie-server-tests: align WildFly client to bom

sutaakar:
wildfly-controller-client is needed by Cargo, used in JmsQueueIntegrationTest.
Updated group Id and aligned to use version defined in WildFly bom.


sutaakar:
Will be updated with other changes.

PR #482, opened 2016-06-02
[6.4.x] RHBRMS-2472 KieServiesClient.completeConversation() won't unset all conversationIds

mswiderski:
back port of required changes for RHBRMS-2472


mswiderski:
merged as required by one off patch

PR #483, opened 2016-06-02
kie-server-tests: on-demand deployment/undeployment updates

sutaakar:
wildfly-controller-client is needed by Cargo, used in JmsQueueIntegrationTest.
Updated group Id and aligned to use version defined in WildFly bom.
Separated path to Kie server WAR file to be possible to define custom Kie server WAR file when running custom container.


mswiderski:
+1
thanks @sutaakar

PR #484, opened 2016-06-06
kie-server-tests: restructured storage property

sutaakar:
To make it configurable externally, also removing duplicities.


mswiderski:
+1

PR #485, opened 2016-06-06
kie-server-tests: updated DocumentServiceIntegrationTest

sutaakar:


sutaakar:
I have a question.
In Kie server API in DocumentInstance there is property "link", it doesn't seem to be tested or used. What is it for? Is it there for some future implementation or just forgotten and should be deleted?

mswiderski:
It mirrors the DocumentImpl from jbpm-document module so it comes with all fields. Though I didn't implement the link at the moment as it might differ if you have multiple kie servers. Instead there is getDocumentLink on the DocumentServicesClient.

Although it cover only the kie server client use case and not direct rest usage. Might be worth including support for link on thr document instance even if it will point to thid given server instance.

mswiderski:
+1

PR #487, opened 2016-06-08
kie-server-tests: release REST connection in finally block

sutaakar:
Also check for OK response status for DELETE requests - WebLogic returns 200 in case of no content in DELETE response.


mswiderski:
+1

PR #489, opened 2016-06-09
JBPM-5071 - Add default process event listeners from workbench

cristianonicolai:
Added InitiatorProviderProcessListener and TaskCleanUpProcessEventListener event listeners

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
@@ -72,9 +72,10 @@ import org.kie.api.task.UserGroupCallback;
 import org.kie.internal.runtime.conf.DeploymentDescriptor;
 import org.kie.internal.runtime.conf.MergeMode;
 import org.kie.internal.runtime.conf.NamedObjectModel;
+import org.kie.internal.runtime.conf.ObjectModel;
+import org.kie.internal.runtime.conf.RuntimeStrategy;
 import org.kie.internal.task.api.UserInfo;
 import org.kie.scanner.KieModuleMetaData;
-import org.kie.internal.runtime.conf.RuntimeStrategy;
 import org.kie.server.api.KieServerConstants;
 import org.kie.server.api.model.KieServerConfig;
 import org.kie.server.services.api.KieContainerCommandService;
@@ -226,8 +227,6 @@ public class JbpmKieServerExtension implements KieServerExtension {
             ((KModuleDeploymentService) deploymentService).setExecutorService(executorService);
         }
 
-
-
         this.kieContainerCommandService = new JBPMKieContainerCommandServiceImpl(context, deploymentService, new DefinitionServiceBase(definitionService),
                 new ProcessServiceBase(processService, definitionService, runtimeDataService, context), new UserTaskServiceBase(userTaskService, context),
                 new RuntimeDataServiceBase(runtimeDataService, context), new ExecutorServiceBase(executorService, context), new QueryDataServiceBase(queryService, context),
@@ -322,6 +321,21 @@ public class JbpmKieServerExtension implements KieServerExtension {
             unit.setKieContainer(kieContainer);
 
             addAsyncHandler(unit, kieContainer);
+
+            if (config.getConfigItemValue(KieServerConstants.CFG_JBPM_TASK_CLEANUP_LISTENER, "true").equalsIgnoreCase("true")) {
+                logger.debug("Registering TaskCleanUpProcessEventListener");
+                addTaskCleanUpProcessListener(unit, kieContainer);
+            }
+
+            if (config.getConfigItemValue(KieServerConstants.CFG_JBPM_TASK_BAM_LISTENER, "true").equalsIgnoreCase("true")) {
+                logger.debug("Registering BAMTaskEventListener");
+                addTaskBAMEventListener(unit, kieContainer);
+            }
+
+            if (config.getConfigItemValue(KieServerConstants.CFG_JBPM_PROCESS_IDENTITY_LISTENER, "true").equalsIgnoreCase("true")) {
+                logger.debug("Registering IdentityProviderAwareProcessListener");
+                addProcessIdentityProcessListener(unit, kieContainer);
+            }
 
             deploymentService.deploy(unit);
             // in case it was deployed successfully pass all known classes to marshallers (jaxb, json etc)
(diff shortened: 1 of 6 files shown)


cristianonicolai:
Depends on https://github.com/droolsjbpm/jbpm/pull/480

mswiderski:
@cristianonicolai this won't work for runtime strategies other than singleton. You need to add them to deployment descriptor similar as it's done for async workitem handler. In addition we could add bam task event listener as well so we have all the default listeners registered by default. Though I would make them configurable so users can simply don't register them - with system property as other configurations

cristianonicolai:
@mswiderski My initial version was using the deployment descriptor, problem I found with that is how would the TaskCleanUpProcessEventListener get a reference to the InternalTaskService. Any ideas for that?

mswiderski:
@cristianonicolai you can follow the same approach as here [1] and use mvel resolver and then simply pass taskService as argument in the constructor. This is named parameter that will be injected into the listener when it is created - by runtime manager code

[1] https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java#L470

cristianonicolai:
@mswiderski thanks. Updated the PR with the items you suggested. 

The identity listener is automatically registered in here: https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java#L209

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -474,6 +478,41 @@ protected void addAsyncHandler(KModuleDeploymentUnit unit, InternalKieContainer
...
     }
 
+    protected void addTaskBAMEventListener(final KModuleDeploymentUnit unit, final InternalKieContainer kieContainer) {
+        final DeploymentDescriptor descriptor = getDeploymentDescriptor(unit, kieContainer);
+        descriptor.getBuilder().addTaskEventListener(
+                new NamedObjectModel(
+                        "mvel",
+                        "taskBAM",
+                        "new org.jbpm.services.task.lifecycle.listeners.BAMTaskEventListener(" +
+                            "org.jbpm.runtime.manager.impl.jpa.EntityManagerFactoryManager.get().getOrCreate(\"" + persistenceUnitName + "\")" +
```
mswiderski: Bam task listener should be without entity manager factory https://github.com/droolsjbpm/jbpm/blob/master/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/lifecycle/listeners/BAMTaskEventListener.java#L71 so it will use one taken from taskcontext, see here https://github.com/droolsjbpm/jbpm/blob/master/jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/persistence/PersistableEventListener.java#L44

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -323,6 +321,16 @@ public void createContainer(String id, KieContainerInstance kieContainerInstance
 
             addAsyncHandler(unit, kieContainer);
 
+            if (config.getConfigItemValue(KieServerConstants.CFG_JBPM_TASK_CLEANUP_LISTENER, "true").equalsIgnoreCase("true")) {
+                logger.info("Registering TaskCleanUpProcessEventListener");
```
mswiderski: Maybe use debug instead?

mswiderski:
Two small comments and can be merged

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -474,6 +483,51 @@ protected void addAsyncHandler(KModuleDeploymentUnit unit, InternalKieContainer
         }
     }
 
+    protected void addTaskBAMEventListener(final KModuleDeploymentUnit unit, final InternalKieContainer kieContainer) {
+        final DeploymentDescriptor descriptor = getDeploymentDescriptor(unit, kieContainer);
+        descriptor.getBuilder().addTaskEventListener(
+                new NamedObjectModel(
```
mswiderski: It's enough to use ObjectModel instead of NamedObjectModel as named is used for objects that needs to be registred with given name like work item handlers ir globals.

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -474,6 +483,51 @@ protected void addAsyncHandler(KModuleDeploymentUnit unit, InternalKieContainer
...
+                )
+        );
+        unit.setDeploymentDescriptor(descriptor);
+    }
+
+    protected void addTaskCleanUpProcessListener(final KModuleDeploymentUnit unit, final InternalKieContainer kieContainer) {
+        final DeploymentDescriptor descriptor = getDeploymentDescriptor(unit, kieContainer);
+        descriptor.getBuilder().addEventListener(
+                new NamedObjectModel(
+                        "mvel",
```
mswiderski: Same here, ObjectModel instead NamedObjectModel

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -474,6 +483,51 @@ protected void addAsyncHandler(KModuleDeploymentUnit unit, InternalKieContainer
...
+                        "new org.jbpm.services.task.admin.listener.TaskCleanUpProcessEventListener(taskService)"
+                )
+        );
+        unit.setDeploymentDescriptor(descriptor);
+    }
+
+    protected void addProcessIdentityProcessListener(final KModuleDeploymentUnit unit, final InternalKieContainer kieContainer) {
+        final DeploymentDescriptor descriptor = getDeploymentDescriptor(unit, kieContainer);
+        descriptor.getBuilder().addEventListener(
+                new NamedObjectModel(
```
mswiderski: Same here, ObjectModel instead NamedObjectModel

cristianonicolai:
@mswiderski done with the updates, please review it once again in case I missed anything. Thanks.

mswiderski:
Excellent work @cristianonicolai +1 to merge

PR #490, opened 2016-06-09
kie-server-tests: custom client-deployment-settings.xml

sutaakar:
Allow external definition of kie-server-testing-client-deployment-settings.xml - for case when we need to use custom private repositories for kjar dependencies.

Side note: using 2 different approaches for creating kjars in tests(by MavenCLI and kie-ci) makes things complicated. I would like to remove one of them in future. What do you think, which one should stay? I think that kie-ci provides more flexibility, could be more useful to keep it.


mswiderski:
+1 to merge

Personally I don't have preferences on which approach is better as long as they do the same we can skip one. I think @psiroky created the test infrastructure so might be best person to advice here

psiroky:
I believe you can't (yet?) build Maven project with kie-ci (like we do now with MavenCLI). I agree that having just one approach would be better if you can make it work. 

It's true that I created the original infrastructure, but things have changed rapidly since then, so it may not even be applicable anymore. @sutaakar go with whatever solution you think is the best.

PR #491, opened 2016-06-09
[6.4.x] kie-server-tests: custom client-deployment-settings.xml

sutaakar:
Allow external definition of kie-server-testing-client-deployment-settings.xml - for case when we need to use custom private repositories for kjar dependencies.


mswiderski:
+1 to merge, affects tests only

PR #492, opened 2016-06-10
kie-remote-jaxb-gen: remove unused plugin config (thanks @baldimir)

psiroky:
* the plugin was bound to non-existing phase inititalize
   (typo, should be initialize). Because of that it was
   never executed and thus is not needed at all


baldimir:
+1

psiroky:
Fixed typo in commit message.

PR #493, opened 2016-06-13
kie-server-tests: use BOM for kjar dependencies

sutaakar:
@psiroky What do you think, is it good to use this platform BOM or I should rather use several specific BOMs for drools/jbpm/optaplanner...


psiroky:
I think it is fine for testing purposes. The only disadvantage is that it depends on several other BOMs, so it will take slightly longer to download everything (which is really not an issue here).

sutaakar:
Fixed compilation warnings in test processes.

mswiderski:
+1 to merge

PR #495, opened 2016-06-15
kie-server: controller - empty KieServerInstanceList fix

jakubschwan:
When is created `KieServerInstanceList` with no `KieServerInstance` and then try get all server instances by `getKieServerInstances()` then KieServerList created with Json marshalling return _empty array_ and  list created with Jaxb marshalling return _null_.
This is caused by content of response: Jaxb ( `<kie-server-instance-list/>` ), JSON (` "kie-server-instance" : [ ] `) so there are used different constructors.
Empty `KieServerInstances` in `KieServerInstanceList` should be empty and not null.


jakubschwan:
jenkins retest this

mswiderski:
+1 to merge

PR #503, opened 2016-06-20
JBPM-5158 - AppScopedEntityManager not clean up by TaskTransactionInt…

mswiderski:
…erceptor


mswiderski:
Merging based on test outcome that proved it fixes the issue

PR #508, opened 2016-06-22
Scores should automatically marshall correctly with XStream - 

ge0ffrey:
With JAXB this is impossible, but we add the package-info anyway so it works for our kie-server-api classes at least.

This PR also points out a TODO related PLANNER-604 that were already there (and aren't being fixed now), but at least it brings that TODO away from hiding.


mswiderski:
+1 to merge

PR #511, opened 2016-06-28
kie-server-tests: add update tests for drools

sutaakar:
Showing problem with updating of container using updateReleaseId() client method.

PR info: test is about updating container from kjar represented by [1] to new kjar. Both kjars are similar, there is just different kie base name and kie session. Both kjars contain Person class which set specific value as their id.

[1] https://github.com/sutaakar/droolsjbpm-integration/tree/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/filtered-resources/kjars-sources/container-isolation-kjar1


mswiderski:
jenkins retest this

sutaakar:
Will be extended in new PR.

sutaakar:
Reopened - there were some GitHub issues.

mswiderski:
incorporated with https://github.com/droolsjbpm/droolsjbpm-integration/pull/544

PR #512, opened 2016-06-29
JBPM-5175: ClassCastException on Tomcat8

sutaakar:
Excluding jboss-transaction-api from webc WAR file.


mswiderski:
+1 to merge

PR #515, opened 2016-06-30
JBPM-5175: update Tomcat controller test war

sutaakar:
Caused by update of resteasy.


kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-servlet-container.xml
```diff
@@ -32,6 +32,10 @@
       <directory>${project.basedir}/src/main/webc-resources</directory>
       <outputDirectory>.</outputDirectory>
     </fileSet>
+    <fileSet>
+      <directory>${project.basedir}/target/classes</directory>
```
psiroky: Minor thing, but you can use `${project.build.outputDirectory}`.

sutaakar: Good catch, thanks. Updated.

mswiderski:
+1 to merge

PR #524, opened 2016-07-05
JBPM-5071 - support for query service to allow filtering by user's as…

mswiderski:
…signed projects in kie server jbpm extension

@cristianonicolai could you please have a look


cristianonicolai:
@mswiderski looks good +1

PR #525, opened 2016-07-05
DROOLS-1224 - Not able to update kie-server container version using R…

mswiderski:
…EAT API

@etirelli do you mind taking a look?

In addition to the fix of the upgrade of container I added new endpoint that allows to get the current state of the server - like the one stored in xml. I believe it makes sense to be able to easily get the current state (especially for unmanaged but for managed could be good as well used for verification). And that was the only way I could test it :)


etirelli:
@mswiderski I agree with you about adding the endpoint. My only concern is that QE might block this into the 6.4.x branch. 

So +1 to merge into 6.5.x and master. Lets talk to QE about 6.4.x.

PR #528, opened 2016-07-06
JBPM-5192 - After the simulation fails the server becomes unresponsiv…

mswiderski:
…e and it does not even respond to shutdown requests


csadilek:
+1

Only comment I have is that we could point out the existence of the system property in the error message: ```Unable to calculate path elements of the process - max size (" + maxElementsSize + ") of elements exceeded. Use -Dorg.jbpm.simulation.max.elements to increase this limit.``` Not sure that's needed or desired though.

mswiderski:
this property is given just in case simulation still works for big (though not complex) processes which would then exceed the default limit. In most of the cases it should not reach that in normal cases but only in those that simulation cannot handle, meaning increasing limit will not help anyway.

mswiderski:
build failure (tests) not related to this one though they are related to JBPM-5191, will update tests accordingly and rerun the build

PR #530, opened 2016-07-06
[6.4.x] JBPM-5192 - After the simulation fails the server becomes unresponsiv…

mswiderski:
…e and it does not even respond to shutdown requests


csadilek:
+1

PR #535, opened 2016-07-07
JBPM-5071 - kie server client load balancer - avoid duplicates, make …

mswiderski:
…sure services clients share same instance of load balancerto keep them always updated

@cristianonicolai could you please take a look?


cristianonicolai:
@mswiderski looks good +1

PR #537, opened 2016-07-08
kie-server - jbpm - avoid npe and provide proper response on get task…

mswiderski:
… if task does not exist

@cristianonicolai one more please :)


cristianonicolai:
+1

PR #539, opened 2016-07-13
kie-server-tests: move container disposing to base test class

sutaakar:
So every test case get clean Kie server out of the box.


mswiderski:
+1 to merge once build completes

thanks a lot @sutaakar!

PR #540, opened 2016-07-18
kie-server-tests: raise timeout for sync with controller

sutaakar:
Raise timeout for synchronization between controller and Kie server.
If controller is integrated as part of workbench then starting of containers is executed asynchronously. This cause delays between returning of response from controller and actual start of container. Low timeout threshold  cause test instability.


mswiderski:
+1

PR #543, opened 2016-07-20
kie-server-tests: update unstable escalation test

sutaakar:
This test become unstable for slower databases, causing false positives. Caused by starting of task more than 2 seconds after process was started - reassignment already triggered.
Removing of this test will save us more than 18 seconds.
I think that this test doesn't bring us so much value.

@mswiderski  What do you think about it?


mswiderski:
in general I am fine with that, but before a merge maybe we could have it "flagged" as just to be executed on the fast dbs instead of removing it completely? Does that make any sense to you (or is it doable)?

triceo:
Let's not remove test coverage unless it's absolutely necessary. In the bigger picture of things, I do not think 18 seconds matter that much.

sutaakar:
Put a flag there for case if it will still cause problems.
Otherwise the problem should hopefully get better, I deleted there one basically redundant check - one less call to container and DB.

mswiderski:
+1

PR #544, opened 2016-07-20
DROOLS-1231 - Container is not upgraded until restarted

mswiderski:


sutaakar:
+1

mswiderski:
merged manually to preserve commits

PR #545, opened 2016-07-20
JBPM-5216 - Adding note about resources

mrietveld:
@lazarotti Also adding a note so that the name doesn't change again.. :)


mswiderski:
+1

PR #546, opened 2016-07-21
[6.5.x] Backport of various fixes from master

mswiderski:
@sutaakar @jakubschwan could you please check it out if I missed/wrongly merged as there were quite few test refactoring on kie server side, thanks


sutaakar:
I noticed one thing.... I didn't find here https://issues.jboss.org/browse/JBPM-5083. Should it be backported too?

mswiderski:
yes, I believe it would make sense to be included

sutaakar:
Found one more fix to be backported: https://issues.jboss.org/browse/JBPM-5098 .

mswiderski:
@sutaakar are we missing anything or we good to merge this one?

sutaakar:
I think we are good to merge. If we will find anything missing we can add it later.

mswiderski:
merged manually

PR #551, opened 2016-07-26
RHBPMS-397 - Allow insecure Remote task operations (not only limited to GetTask* commands)

mrietveld:
@mswiderski Just in case this part of the code remains in 7.x..


mswiderski:
since we have the PR let's merge it, even if we remove that part of code it won't hurt to have it for now. +1 to merge

PR #552, opened 2016-07-27
[DROOLS-969] implement import tag in kie-aries-blueprint integration

mariofusco:


kie-aries-blueprint/pom.xml
```diff
@@ -33,6 +33,12 @@
       <artifactId>org.apache.aries.blueprint.api</artifactId>
       <scope>compile</scope>
     </dependency>
+    <!--<dependency>-->
```
winklerm: I would vote for removing the commented out dependency, unless you have some plans with it?

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/factorybeans/KieImportResolver.java
```diff
@@ -0,0 +1,79 @@
...
+    private synchronized KieContainer registerKieContainer(BlueprintContextHelper context) {
+        if (kieContainer == null) {
+            KieServices ks = KieServices.Factory.get();
+            if ( releaseId == null ) {
+                kieContainer = ks.getKieClasspathContainer();
+            } else {
+                kieContainer = resolveKContainer( releaseId );
+                if (scannerEnabled) {
+                    KieScanner kieScanner = KieServices.Factory.get().newKieScanner( kieContainer );
+                    kieScanner.start(scannerInterval);
```
winklerm: How is the kieScanner stopped? It could be probably done by registering an 'destroy-method' with the bean, or is it done somewhere else?

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/namespace/KieImportElementParser.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        String enableScanner = element.getAttribute("enableScanner");
+        beanMetadata.addArgument(createValue(context, "true".equals( enableScanner )), null, 2);
+
+        String scannerInterval = element.getAttribute("scannerInterval");
+        long interval = -1L;
+        try {
+            interval = Long.parseLong( scannerInterval );
+        } catch (NumberFormatException nfe) { }
```
winklerm: Do we want to swallow the nfe? Maybe failing fast would be safer?

kie-aries-blueprint/src/main/resources/org/kie/aries/blueprint/kie-aries-blueprint.xsd
```diff
@@ -97,6 +97,14 @@
     </xsd:complexType>
   </xsd:element>
 
+  <xsd:element name="import">
+    <xsd:complexType>
+      <xsd:attribute name="releaseId-ref" use="optional" type="xsd:string"/>
```
winklerm: Is the releaseId-ref attribute optional, or should it be mandatory?

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/KieNamespaceHandler.java
```diff
@@ -43,7 +55,6 @@
     private static final String BLUEPRINT_NS = "http://www.osgi.org/xmlns/blueprint/v1.0.0";
 
     public static final String ELEMENT_RELEASE_ID = "releaseId";
-    public static final String ELEMENT_KBASE_REF = "kbase-ref";
```
winklerm: Just to recap: we are removing the _kbase-ref_ element, because the preferred way of using KieBases/KieSessions from a kJar is by _import_ element. In addition, _ksession-ref_ element is preserved because of possible need to inject _TransactionManager_, _UserTransaction_, etc. into it. Is this correct?

kie-aries-blueprint/pom.xml
mariofusco: Agreed, I'll remove this. It is just a leftover when I was experimenting with different solutions.

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/KieNamespaceHandler.java
mariofusco: Correct.

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/factorybeans/KieImportResolver.java
mariofusco: At the moment there's no way to stop it declaratively. The only thing you can is programmatically getting the scanner from the container and stopping it.

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/namespace/KieImportElementParser.java
mariofusco: Agreed, we should fail fast here.

kie-aries-blueprint/src/main/resources/org/kie/aries/blueprint/kie-aries-blueprint.xsd
mariofusco: It's optional and in that case it means that you want to import the classpath kiecontainer.

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/namespace/BlueprintContextHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    private final Object blueprintContext;
+    private final Method addObjectMethod;
+
+    public BlueprintContextHelper(BlueprintContainer blueprintContainer) {
+        this.blueprintContainer = blueprintContainer;
+
+        // The ExecutionContext class is in the blueprint-core module, but we cannot explicitly import it
+        // because there are 2 implementations of it (osgi and no-osgi) with same class names and we
+        // dynamically import one of the 2. For this reason ExecutionContext class can be used only via reflection
+
```
winklerm: I was looking at alternative approaches to avoid reflection but could not find anything suitable at the moment. I presume that this will have to do for now, at least you found a way how to make it work, that's great!

winklerm:
I left some comments, otherwise it looks good to me.

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/factorybeans/KieImportResolver.java
winklerm: I see, but wouldn't it leave behind a thread?

kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/namespace/BlueprintContextHelper.java
mrietveld: The Service/[ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) mechanism could work. 

However, with regards to performance, I'm pretty sure there's no difference between the 2. (Under the hood, ServiceLoader is also just... reflection. :) )

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/20ad89f28ef4f1e72ae7920adf5a50c9edb2dac7

PR #554, opened 2016-08-02
JBPM-5228 - Allow 'fire and forget' feature in runtime Kie Server Rem…

mswiderski:
…ote APIs

@sutaakar @livthomas here comes the fire and forget support for JMS kie server client. In addition there is an async response and transactional support for jms client as well as they all come in the same party so to say.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
@@ -15,6 +15,8 @@
 
 package org.kie.server.client;
 
+import org.kie.server.client.jms.ResponseHandler;
+
 public interface UIServicesClient {
 
     /**
@@ -67,4 +69,11 @@ public interface UIServicesClient {
      * @return svg (xml) representing process image annotated with active (in red) and completed (in grey) nodes
      */
     String getProcessInstanceImage(String containerId, Long processInstanceId);
+
+    /**
+     * Override default response handler to change interaction pattern. Applies only to JMS
+     * based integration.
+     * @param responseHandler
+     */
+    void setResponseHandler(ResponseHandler responseHandler);
 }
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/DocumentServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/DocumentServicesClientImpl.java
@@ -64,6 +64,9 @@ public class DocumentServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<DocumentInstance> response = (ServiceResponse<DocumentInstance>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             result = response.getResult();
         }
 
@@ -86,6 +89,9 @@ public class DocumentServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             result = response.getResult();
         }
         if (result instanceof Wrapped) {
@@ -150,6 +156,9 @@ public class DocumentServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<DocumentInstanceList> response = (ServiceResponse<DocumentInstanceList>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             result = response.getResult();
         }
 
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
@@ -398,7 +398,9 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM", containerId ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
-
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             contentId = deserialize(response.getResult(), Object.class);
         }
 
@@ -426,7 +428,9 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM", containerId ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
-
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             variables = deserialize(response.getResult(), Object.class);
         }
         if (variables instanceof Wrapped) {
@@ -453,6 +457,9 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM", containerId ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             variables = deserialize(response.getResult(), Object.class);
         }
         if (variables instanceof Wrapped) {
@@ -508,7 +515,9 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM", containerId ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
-
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             commentId = deserialize(response.getResult(), Object.class);
         }
 
@@ -556,7 +565,9 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
             ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM", containerId ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
-
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
             commentList = deserialize(response.getResult(), TaskCommentList.class);
         }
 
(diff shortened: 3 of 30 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,260 @@
...
+    @BeforeClass
+    public static void buildAndDeployArtifacts() {
+        KieServerDeployer.buildAndDeployCommonMavenParent();
+        KieServerDeployer.buildAndDeployMavenProject(ClassLoader.class.getResource("/kjars-sources/definition-project").getFile());
+
+        createContainer(CONTAINER_ID, RELEASE_ID);
+    }
+
+    @After
+    public void resetResponseHandlers() {
```
rsynek: Is there missing anything or is it a leftover of previous code?

mswiderski: this needs to be removed as it was attempt to reset the handlers but it turned out it's not needed, thanks for pointing out.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/AsyncResponseHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                if (consumer != null) {
+                    try {
+                        consumer.close();
+                    } catch (JMSException e) {
+                        logger.warn("Error when closing JMS consumer due to {}", e.getMessage());
+                    }
+                }
+                // submit work to executor service to close resources
+                // as they cannot be closed from message listener
+                executorService.submit(new Runnable() {
```
rsynek: Sorry - I don't get the reason why is necessary to dispose Session and Connection from a separate thread instead of using AsyncResponseHandler.dispose(). Could you please clarify?

mswiderski: dispose is invoked before the message arrives as it's async. Imagine situation where you call client.startProcess that uses async handler, this means the message is sent and startProcess method completes - it does not wait for response so if we would close connection and session in dispose that would essentially kill the message listener and it would never receive the response.

does it make sense?

rsynek: Thanks - partially yes, dispose() was a bad idea. Still, I wonder why not closing it in the finally block of the AsyncMessageListener. The response should be received in this moment and just passed to callback, so it should be safe to dispose these resources. Or should not? :-)

mswiderski: I thought so too, and even did implement it as such but then jms provider complained that it cannot be done that way - you can't close connection nor session from within JMS message listener :(

rsynek: Fair enough, could you please add a comment with explanation for next generations?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/DocumentServicesClientImpl.java
```diff
@@ -64,6 +64,9 @@ public DocumentInstance getDocument(String identifier) {
             ServiceResponse<DocumentInstance> response = (ServiceResponse<DocumentInstance>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
```
rsynek: I wonder if this can be rewritten using Optional on master (7.0) - in general, returning null means there is a risk of NPE somewhere else. For 6.5 stream, though, I understand there is no perfect solution due to backward compatibility and support of JDK 6 and 7.

mswiderski: that would break the api rather significantly. Putting optional still does not cover all cases and it requires additional step to get the data for non null cases. So I prefer to stick to this as it easier to deal with and majority of cases will be non null - only async handlers for jms will return null, other (request reply for jms and rest will return always non null value. Taking into consideration that REST is most likely used more than JMS it's better to stick to what we have at the moment.

mswiderski:
@rsynek added comment with reference to actual error code

thanks for review!

livthomas:
@mswiderski I have noticed that `ResponseCallback` has only a single implementation `BlockingResponseCallback`. Why is it done this way? Do you plan to add some more implementations in the future? Or are we just providing a way for users to implement their own callbacks?

mswiderski:
correct, it's for extensibility so users can use whatever callback they need.

sutaakar:
looks good for me

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/ReqReplyResponseHandler.java
```diff
@@ -0,0 +1,97 @@
...
+import org.kie.server.api.model.ServiceResponse;
+import org.kie.server.api.model.ServiceResponsesList;
+import org.kie.server.client.KieServicesClient;
+import org.kie.server.client.KieServicesConfiguration;
+import org.kie.server.client.KieServicesException;
+import org.kie.server.client.impl.KieServicesClientImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReqReplyResponseHandler implements ResponseHandler {
```
livthomas: This is just a detail but why is this called `ReqReplyResponseHandler` and not `RequestReplyResponseHandler`? It would still be shorter than `FireAndForgetResponseHandler`. The current name might be a little bit ambiguous.

mswiderski: not really important reason, I can update it to be RequestReplyResponseHandler

mswiderski: class name updated

mswiderski:
any other comments? I am about to prepare back port for 6.5.x

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
```diff
@@ -67,4 +69,11 @@
      * @return svg (xml) representing process image annotated with active (in red) and completed (in grey) nodes
      */
     String getProcessInstanceImage(String containerId, Long processInstanceId);
+
+    /**
+     * Override default response handler to change interaction pattern. Applies only to JMS
+     * based integration.
+     * @param responseHandler
```
psiroky: I believe we should not use empty `@params` in javadoc. I would either remove it or add some description.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
```diff
@@ -1227,7 +1280,9 @@ public TaskInstance findTaskById(Long taskId) {
             ServiceResponse<TaskSummaryList> response = (ServiceResponse<TaskSummaryList>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
-
+            if (shouldReturnWithNullResponse(response)) {
+                return null;
+            }
```
psiroky: I guess we can't much with this right now, but it seems to me that needing to add the same check to dozens of places is far from ideal.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/AsyncResponseHandler.java
```diff
@@ -0,0 +1,162 @@
...
+                // submit work to executor service to close resources
+                // as they cannot be closed from message listener
+                // due to AMQ129006: It is illegal to call this method (session.close()) from within a Message Listener
+                executorService.submit(new Runnable() {
+                    @Override
+                    public void run() {
+                        try {
+                            if (session != null) {
+                                session.close();
+                                logger.debug("Session closed via separate thread.");
```
psiroky: Could we perhaps add more info about the session in to the log message? From my experience the more specific data the logging message contain the better, it is then much easier to figure our what is going on.

psiroky:
@mswiderski sorry for the delay. I quickly checked the code and generally looks good. Just added few minor comments.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/AsyncResponseHandler.java
mswiderski: not sure I understand the request here... the debug is only to have option to know if session was closed or not to track connection/session leak. Although they will be still closed (with stack trace in logs) by jms provider

psiroky: I was thinking about something like `logger.debug("Session {} closed via....", session.getId())`. But looks like there is actually no such information as part of the session. I guess we can't really provide more info here...please just ignore my previous comment, sorry about the noise.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/BlockingResponseCallback.java
```diff
@@ -0,0 +1,101 @@
...
+ * responses and block
+ * <ul>
+ *     <li>client if there is no message yet</li>
+ *     <li>server if the queue is full</li>
+ * </ul>.
+ */
+public class BlockingResponseCallback implements ResponseCallback {
+
+    private static final Logger logger = LoggerFactory.getLogger(BlockingResponseCallback.class);
+    private BlockingQueue<ServiceResponsesList> responses = new ArrayBlockingQueue<ServiceResponsesList>(10);
```
livthomas: What is the purpose of this `BlockingQueue` with a capacity of 10? Is `BlockingResponseCallback` intended to be used for retrieving only a single response and this queue capacity is just a "random" number? Or do we expect this callback to be reused for multiple requests so the queue will contain multiple responses? If the latter is true, I would recommend to add another constructor with this signature: `public BlockingResponseCallback(Marshaller marshaller, int queueCapacity)` Because it does not make sense for users to implement their own `ResponseCallback` just to be able to handle more than 10 responses.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/BlockingResponseCallback.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private Marshaller marshaller;
+
+    public BlockingResponseCallback(Marshaller marshaller) {
```
livthomas: Do we always need to specify `Marshaller` even though we do not use any custom classes? What about instantiating some default marshaller in non-parametric constructor and adding JavaDoc comment to the one with `Marshaller` parameters explaining when and why you need to use it?

I can see that `null` is passed as marshaller in some tests but this can lead to NPE in some cases, am I right?

mswiderski: no, we don't need always the marshaller in case you want to get raw ServiceResponseList and then deal with reading the response yourself. If you would like to have this done for you then you need to provide your marshaller that is container specific, so it needs to be provided by the one how registers the callback.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/jms/BlockingResponseCallback.java
mswiderski: in general it should not be used for more than single message but at the same time there could be multiple calls initiated from the same client instance. And since the callback is global on the client instance it should support more repossess to be received. 
This is a very simple implementation, almost like a demo one as it will vary between use cases people want to use it for. So main purpose was to make it pluggable and easy to adjust to your needs.

+1 at adding the new constructor.

PR #556, opened 2016-08-04
[6.5.x] kie-remote-jaxb-gen: upgrade jaxb2-maven-plugin from 1.6 to 2.2

psiroky:
* the upgrade should fix random failures recently seen when
   building with JDK8

 * similar upgrade was done on master few months back and everything seems to work there

@mbiarnes, @mrietveld could you please take a look?

### DIFF
--- a/kie-remote/kie-remote-jaxb-gen/pom.xml
+++ b/kie-remote/kie-remote-jaxb-gen/pom.xml
@@ -237,32 +237,9 @@
       </plugin>
 
       <plugin>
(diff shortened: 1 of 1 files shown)


psiroky:
The test failures are not related to this PR, but we need to fix them anayway.

CC @mariofusco `kie-aries-blueprint` + `kie-ci` strikes again :) I'll see if I can implement the same workaround as before.

mbiarnes:
+1 seems good to me

PR #566, opened 2016-08-10
Added test for kie-maven-plugin that reproduces problem with 2 DSL files. 

baldimir:
Please **DO NOT MERGE** now. We need to discuss a solution for this problem with @mariofusco and @etirelli before merging. 

This is a reproducer extracted from issue reported by @tkobayas on support mailinglist. Thanks @tkobayas!

### DIFF
--- a/kie-maven-plugin/src/test/java/org/kie/maven/plugin/TwoDslFilesIntegrationTest.java
+++ b/kie-maven-plugin/src/test/java/org/kie/maven/plugin/TwoDslFilesIntegrationTest.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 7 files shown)


mswiderski:
this is green (including tests) is it waiting for anything or can be merged or closed?

etirelli:
I believe @mariofusco already worked on this and solved it? @mariofusco can you please confirm?

mswiderski:
@mariofusco any update what to do with this PR?

mariofusco:
@etirelli @mswiderski The test is green only for a lucky coincidence. The problem is in the fact that when you have more than one DSL file the order in which you process them matters, but at the moment if those DSL are in the same project you don't have any control on that order. One possible solution would be allowing to declare inside the pom the order in which DSL files have to be processed. However we decided to not do this for now and just suggested to keep all DSL statements belonging to the same area in the same file, so to have control on their order.

PR #568, opened 2016-08-10
minor document related improvements to make it easier to consume

mswiderski:


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/DocumentInstanceList.java
```diff
@@ -29,7 +29,7 @@
 @XmlRootElement(name = "document-instance-list")
 public class DocumentInstanceList implements ItemList<DocumentInstance> {
 
-    @XmlElement(name="document-instance")
+    @XmlElement(name="document-instances")
```
psiroky: Why the change to plural?

mswiderski: because it represents a list/array so it better matches the wrapper when marshalled

psiroky: Hmm, I thought that the @XmlElement defines the name for individual XML tags (not the wrapping collection tag). What wrapper do you mean? (I ma likely just not seeing something, but would like to understand what is going on :))

mswiderski: it is used by both jaxb and json

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/DocumentServiceBase.java
```diff
@@ -117,13 +117,13 @@ public DocumentInstanceList listDocuments(Integer page, Integer pageSize) {
...
         for (Document doc : documents) {
-            list.add(convertDocument(doc));
+            list.add(convertDocument(doc, false));
         }
 
         return list;
     }
 
-    protected DocumentInstance convertDocument(Document document) {
+    protected DocumentInstance convertDocument(Document document, boolean withContent) {
```
psiroky: Tbh, I am not a big fan of these boolean flagging args :) I would prefer enum (even with just two values). But I won't hold the PR for this, since it is largely matter of personal style.

psiroky:
Looks good. +1

PR #571, opened 2016-08-11
JBPM-5220 kie-server: add tests for filtred list of containers

jakubschwan:
Tests scenarios from KieServerContainerCRUDIntegrationTest are included in the new test class.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerContainerCRUDIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerContainerCRUDIntegrationTest.java
@@ -21,13 +21,10 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.kie.server.api.model.KieContainerResource;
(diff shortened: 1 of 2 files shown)


psiroky:
Thanks for the additional tests!

Just one thing I am wondering about: should we add all these as full integration (system) tests? My idea was to add just few "full system" tests and then test the logic more exhaustively via unit tests. But I am open to different views as well.

WDYT? @sutaakar, @mswiderski?

sutaakar:
I think it is good idea to have exhaustive coverage just as unit tests. We don't need there to check all marshalling possibilities, that can be checked by narrow test selection in integration tests.

mswiderski:
in general I think majority of the test for kie server should be as integration tests as we were already bitten by different containers/formats issues so the better coverage there the better in my opinion.

At the same time, we need to use our common sense to not overkill. For things that might not have any impact on this then unit test should be enough - although if there are extensive use of mocks in unit tests then I'd say we need integration tests for it as well.

mswiderski:
so let's vote - I give +1 as it's good piece of test coverage and is already done

psiroky:
+0. I believe we should have less "full system" tests as they are orders of magnitude slower than the unit tests. If we use good abstractions we should be able to test the logic extensively via unit tests and have just few "full system" tests to verify everything was plugged together correctly.

I agree that since the tests are already done, it would be a waste not to merge them.

sutaakar:
+1

PR #576, opened 2016-08-15
karaf-itests: disable debug logging

psiroky:
* logback defaults to DEBUG logging in case
   there is no explicit config. That in turn
   leads to bloated logs.


mswiderski:
jenkins retest this

mswiderski:
+1 to merge

PR #577, opened 2016-08-15
[JBPM-5250] OSGify all needed kie-remote modules

psiroky:
* while the "wrap" hack works on Karaf, it does not work
   on Equinox. In order to support Equinox we need all deps
   properly OSGified.

@mrietveld could you please take a look at this?

### DIFF
--- a/kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
+++ b/kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
@@ -8,9 +8,9 @@
     <version>6.5.0-SNAPSHOT</version>
   </parent>
 
-  <packaging>jar</packaging>
   <artifactId>kie-remote-ws-impl</artifactId>
   <name>KIE Remote Webservices :: Implementation</name>
+  <packaging>bundle</packaging><!-- bundle = jar + OSGi metadata -->
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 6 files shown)


mswiderski:
jenkins retest this

mrietveld:
@psiroky I'm no OSGI expert, but the package names, etc that you use are correct. 

+1?

psiroky:
Thanks @mrietveld. Just wanted second pair of eyes to go over the changes. I am no OSGi expert either, so was just following what we already had in different modules.

kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
```diff
@@ -70,4 +70,26 @@
...
+        <configuration>
+          <instructions>
+            <Bundle.Symbolic-Name>org.kie.remote.service.ws</Bundle.Symbolic-Name>
+            <Export-Package>
+              org.kie.remote.services.ws.*
+            </Export-Package>
+            <Import-Package>
+              *
+            </Import-Package>
+            <Bundle-Activator>org.drools.core.osgi.Activator</Bundle-Activator>
```
winklerm: The activator looks a bit suspicious to me, is it correct?

kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
```diff
@@ -70,4 +70,26 @@
...
+            <_removeheaders>Private-Package</_removeheaders>
```
winklerm: Why is removing the Private-Package header needed? (It might be perfectly all right, I am just curious.)

kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
psiroky: That would be a result of bad copy&paste. Will fix it, thanks!

kie-remote/kie-remote-ws/kie-remote-ws-impl/pom.xml
psiroky: I wish I knew :) I copy pasted the config from another module...I will try to remove it and see what happens (and if the Private-Package is even generated).

psiroky: I removed both the wrong activator and Private-Package configs and everything seems to work correctly).

psiroky:
I believe this one is now ready to merge.

PR #586, opened 2016-08-22
Allows user to specify additional Maven repositories for pax-exam tests.

jiripetrlik:
Allows user to specify additional Maven repositories for pax-exam tests. This can be useful when running pax-exam tests against binaries which are not available in public Maven repositories.


psiroky:
+1 to merge

PR #588, opened 2016-08-22
[RHBRMS-2541] Add reproducer.

jiripetrlik:
[RHBRMS-2541] Add reproducer.


drools-osgi/drools-karaf-itests/src/it/drools-karaf-itests-kjar/pom.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.drools</groupId>
+    <artifactId>drools-osgi</artifactId>
+    <version>7.0.0-SNAPSHOT</version>
```
psiroky: We will need to use property here, e.g. `${version.org.kie}`. Otherwise the version won't get changed automatically by Maven versions plugin.

drools-osgi/drools-karaf-itests-kjar/pom.xml
```diff
@@ -18,4 +18,17 @@
...
   </dependencies>
 
+  <packaging>kjar</packaging>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-maven-plugin</artifactId>
+        <version>7.0.0-SNAPSHOT</version>
```
psiroky: We should use ${version.org.kie} here.

psiroky: Actually, looking at this again, I am afraid this won't work, because plugin will be in the same reactor build as this module. We had to exclude the kie-maven-plugin-example from the build for the same reason.

drools-osgi/drools-karaf-itests-kjar/src/main/java/org/dummypackage/DummyClass.java
```diff
@@ -0,0 +1,10 @@
+package org.dummypackage;
+
+/**
+ * Created by jpetrlik on 8/25/16.
+ */
```
psiroky: Nit-picking, but we should not be using author tags (see https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/README.md#faq for explanation).

The class is also missing licence header.

mariofusco:
@jiripetrlik I fixed RHBRMS-2542 in drools with this commit https://github.com/droolsjbpm/drools/commit/21b2d5b08

Please let me know if it is all ok now and if so I'll squash and merge this branch.

mariofusco:
Merged by https://github.com/droolsjbpm/droolsjbpm-integration/commit/e3a976103940be20de48e2b19f60e8fcc184b3b6

PR #589, opened 2016-08-22
[6.5.x][JBPM-5282,RHBPMS-4107] recognize GetTaskByWorkItemIdCommand when pr…

psiroky:
…ocessing JAXB responses

@mrietveld could you please review this?

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/jaxb/JaxbCommandsResponse.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/jaxb/JaxbCommandsResponse.java
@@ -35,6 +35,7 @@ import org.jbpm.process.audit.ProcessInstanceLog;
 import org.jbpm.process.audit.VariableInstanceLog;
 import org.jbpm.process.audit.event.AuditEvent;
 import org.jbpm.services.task.commands.GetContentMapForUserCommand;
+import org.jbpm.services.task.commands.GetTaskByWorkItemIdCommand;
 import org.jbpm.services.task.commands.GetTaskContentCommand;
 import org.jbpm.services.task.impl.model.xml.JaxbContent;
 import org.jbpm.services.task.impl.model.xml.JaxbTask;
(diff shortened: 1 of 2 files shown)


kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/jaxb/JaxbCommandsResponse.java
```diff
@@ -212,8 +218,10 @@ public void addResult(Object result, int i, Command<?> cmd) {
             this.responses.add(new JaxbVariableInstanceLog((VariableInstanceLog) result));
         } else if( result instanceof DefaultFactHandle ) { 
            this.responses.add(new JaxbOtherResponse(result, i, cmd));
-        } else if( cmd instanceof GetTaskContentCommand || cmd instanceof GetContentMapForUserCommand ) { 
-           this.responses.add(new JaxbTaskContentResponse((Map<String, Object>) result, i, cmd));
+        } else if( cmd instanceof GetTaskContentCommand || cmd instanceof GetContentMapForUserCommand ) {
+            this.responses.add(new JaxbTaskContentResponse((Map<String, Object>) result, i, cmd));
+        } else if ( cmd instanceof GetTaskByWorkItemIdCommand ) {
+            this.responses.add(new JaxbTaskResponse((Task)result, i,  cmd));
```
mrietveld: @psiroky I think it might be better to modify line 165 so that it gets added there? 

Also, I would make sure that the logger message shows the entire name, instead of just the simple name of the class.

psiroky: Thanks, I fixed both. Please check the result.

I had to change `this.responses.add(new JaxbTaskResponse((JaxbTask)result, i, cmd));` to `this.responses.add(new JaxbTaskResponse((Task)result, i, cmd));` because the result is TaskImpl in case of the `GetTaskByWorkItemCommand`

mrietveld:
+1 Thanks!

PR #592, opened 2016-08-23
[RHBRMS-2541] implement import-session blueprint element

mariofusco:


mariofusco:
Replaced by https://github.com/droolsjbpm/droolsjbpm-integration/pull/596

PR #593, opened 2016-08-24
RHBRMS-2547 Add exclusion for springwork.osgi since the transitive de…

ryanzhang:
…ps are invalid

This is fix is needed to avoid failing download the invalid maven artifacts. 
Those artifacts isn't defined correctly and can't be found in maven central.

PR #595, opened 2016-08-25
RHBRMS-2547 Add exclusion for springwork.osgi since the transitive de…

ryanzhang:
…ps are invalid

This patch has been merged to jboss-integration/droolsjbpm-integration.(See https://github.com/droolsjbpm/droolsjbpm-integration/pull/593 )

This should also added into droolsjbpm/droolsjbpm-integration project.


psiroky:
ok to test

mswiderski:
how is this different from https://github.com/droolsjbpm/droolsjbpm-integration/pull/593

psiroky:
@mswiderski good question :) I originally thought that the target branches are different, but now seeing that both target 6.5.x.

ryanzhang:
Ah, my bad!
 I originally thought the #593 has been merged to jboss-integration/droolsjbpm-integration.
And this one is to PR to droolsjbpm/droolsjbpm-integration.

But you are right.  There is no difference.
I am closing this. Sorry for inconvinence.

PR #596, opened 2016-08-25
[RHBRMS-2541] implement blueprint elements to import ksessions, kbases and kscanners

mariofusco:


winklerm:
Hi @mariofusco, @jiripetrlik, I am wondering if there should be also _import-container_ element, or not. With the current implementation, if I understand it right, I am not able to configure a _KieScanner_ for a _KieContainer_ in Blueprint. I can use _kcontainer-ref_ element, but I cannot configure a _KieScanner_ for it, right?

On the other hand, kie-spring's implementation of kie:import does the same, therefore the above is probably a topic for another JIRA.

PR #597, opened 2016-08-25
[6.5.x] [RHBRMS-2546] Kie Server - Creating container has releaseId null

jakubschwan:


mswiderski:
shouldn't there be any test for it or is it already covered?

jakubschwan:
This issue was detected by [Controller tests](https://github.com/droolsjbpm/droolsjbpm-integration/blob/6.5.x/kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/KieControllerManagementIntegrationTest.java#L292) after implementation of filtering on list of containers ([JBPM-5220](https://issues.jboss.org/browse/JBPM-5220)). Creating of container on kie server take a moment so it's hard to reproduce.
Creating of the container is well tested and filtering of containers list too. There are tests for start and stop container via controller where this issue was found. This is enough in my opinion.

sutaakar:
General question: should container in CREATING state already contain releaseId?
If yes, how it should behave in case release id needs to be resolved (like in case it has version value LATEST)?

psiroky:
@sutaakar great question :) Exactly what I wanted to ask as well. I guess it should already have `releaseId` (e.g. with the LATEST), but it may not have (or can't have?) the `resolvedReleaseId`. WDYT?

sutaakar:
@psiroky I think it is good idea to put there `releaseId` as we know it already when creating container.

mswiderski:
is this ready to be merged?

psiroky:
OK, let's merge this. I will close #614 as duplicate.

mswiderski:
merged, I guess we need one for master then?

psiroky:
Yes, for sure. @jakubschwan could you please create PR for master?

jakubschwan:
Yes, of course

PR #598, opened 2016-08-25
[6.5.x] JBPM-5228 - Add more tests for 'fire and forget' feature

livthomas:
I have added more tests for 'fire and forget' feature. I tried to cover various methods from different `*ServicesClient` implementations.

Some tests are failing and I think those are potential bugs. Please, look at failed tests when the build is finished and tell me your opinion.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
@@ -17,8 +17,6 @@ package org.kie.server.client.impl;
 
 import org.kie.api.command.Command;
 import org.kie.server.api.model.KieContainerResourceFilter;
-import org.kie.server.api.model.KieContainerStatusFilter;
-import org.kie.server.api.model.ReleaseIdFilter;
 import org.kie.server.api.commands.CallContainerCommand;
 import org.kie.server.api.commands.CommandScript;
 import org.kie.server.api.commands.CreateContainerCommand;
@@ -44,6 +42,8 @@ import org.kie.server.client.KieServicesConfiguration;
 import org.kie.server.client.KieServicesException;
 import org.kie.server.client.RuleServicesClient;
 import org.kie.server.client.helper.KieServicesClientBuilder;
+import org.kie.server.client.jms.RequestReplyResponseHandler;
+import org.kie.server.client.jms.ResponseHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -56,7 +56,7 @@ import java.util.ServiceLoader;
 
 public class KieServicesClientImpl extends AbstractKieServicesClientImpl implements KieServicesClient {
 
-    private static Logger logger = LoggerFactory.getLogger( KieServicesClientImpl.class );
+    private static Logger logger = LoggerFactory.getLogger(KieServicesClientImpl.class);
 
     private static final ServiceLoader<KieServicesClientBuilder> clientBuilders = ServiceLoader.load(KieServicesClientBuilder.class, KieServicesClientImpl.class.getClassLoader());
     private static List<KieServicesClientBuilder> loadedClientBuilders = loadClientBuilders();   // load it only once to make sure it's thread safe
@@ -81,11 +81,7 @@ public class KieServicesClientImpl extends AbstractKieServicesClientImpl impleme
         List<String> serverCapabilities = config.getCapabilities();
 
         if (serverCapabilities == null) {
-            // no explicit capabilities configuration given fetch them from kieServer
-            kieServerInfo = getServerInfo().getResult();
-            logger.debug("KieServicesClient connected to: {} version {}", kieServerInfo.getServerId(), kieServerInfo.getVersion());
-            serverCapabilities = kieServerInfo.getCapabilities();
-            logger.debug("Supported capabilities by the server: {}", serverCapabilities);
+            serverCapabilities = getCapabilitiesFromServer();
         }
         if (serverCapabilities != null && !serverCapabilities.isEmpty()) {
             // process available client builders
@@ -123,6 +119,26 @@ public class KieServicesClientImpl extends AbstractKieServicesClientImpl impleme
 
     }
 
+    private List<String> getCapabilitiesFromServer() {
+        ResponseHandler responseHandler = getResponseHandler();
+        if (config.isJms() && !(responseHandler instanceof RequestReplyResponseHandler)) {
+            setResponseHandler(new RequestReplyResponseHandler());
+            kieServerInfo = getServerInfo().getResult();
+            setResponseHandler(responseHandler);
+        } else {
+            kieServerInfo = getServerInfo().getResult();
+        }
+
+        logger.debug("KieServicesClient connected to: {} version {}", kieServerInfo.getServerId(), kieServerInfo.getVersion());
+        List<String> capabilities = kieServerInfo.getCapabilities();
+        logger.debug("Supported capabilities by the server: {}", capabilities);
+
+        return capabilities;
+    }
+
+    private KieServerInfo getServerInfoInTransaction() {
+        throw new IllegalStateException("Cannot get capabilities from server in a transaction");
+    }
 
     @Override
     public <T> T getServicesClient(Class<T> serviceClient) {
@@ -136,11 +152,12 @@ public class KieServicesClientImpl extends AbstractKieServicesClientImpl impleme
 
     @Override
     public ServiceResponse<KieServerInfo> getServerInfo() {
-        if( config.isRest() ) {
-            return makeHttpGetRequestAndCreateServiceResponse( baseURI, KieServerInfo.class );
+        if (config.isRest()) {
+            return makeHttpGetRequestAndCreateServiceResponse(baseURI, KieServerInfo.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new GetServerInfoCommand() ) );
-            return (ServiceResponse<KieServerInfo>) executeJmsCommand( script ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new GetServerInfoCommand()));
+            ServiceResponse<KieServerInfo> response = (ServiceResponse<KieServerInfo>) executeJmsCommand(script).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
@@ -156,91 +173,103 @@ public class KieServicesClientImpl extends AbstractKieServicesClientImpl impleme
             String uri = baseURI + "/containers" + (queryParams.isEmpty() ? "" : "?" + queryParams);
             return makeHttpGetRequestAndCreateServiceResponse(uri, KieContainerResourceList.class);
         } else {
-            CommandScript script = new CommandScript(Collections.singletonList( (KieServerCommand) new ListContainersCommand(containerFilter)));
-            return (ServiceResponse<KieContainerResourceList>) executeJmsCommand(script).getResponses().get(0);
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new ListContainersCommand(containerFilter)));
+            ServiceResponse<KieContainerResourceList> response = (ServiceResponse<KieContainerResourceList>) executeJmsCommand(script).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<KieContainerResource> createContainer(String id, KieContainerResource resource) {
-        if( config.isRest() ) {
-            return makeHttpPutRequestAndCreateServiceResponse( baseURI + "/containers/" + id, resource, KieContainerResource.class );
+        if (config.isRest()) {
+            return makeHttpPutRequestAndCreateServiceResponse(baseURI + "/containers/" + id, resource, KieContainerResource.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new CreateContainerCommand( resource ) ) );
-            return (ServiceResponse<KieContainerResource>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new CreateContainerCommand(resource)));
+            ServiceResponse<KieContainerResource> response = (ServiceResponse<KieContainerResource>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<KieContainerResource> getContainerInfo(String id) {
-        if( config.isRest() ) {
-            return makeHttpGetRequestAndCreateServiceResponse( baseURI + "/containers/" + id, KieContainerResource.class );
+        if (config.isRest()) {
+            return makeHttpGetRequestAndCreateServiceResponse(baseURI + "/containers/" + id, KieContainerResource.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new GetContainerInfoCommand( id ) ) );
-            return (ServiceResponse<KieContainerResource>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new GetContainerInfoCommand(id)));
+            ServiceResponse<KieContainerResource> response = (ServiceResponse<KieContainerResource>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<Void> disposeContainer(String id) {
-        if( config.isRest() ) {
-            return makeHttpDeleteRequestAndCreateServiceResponse( baseURI + "/containers/" + id, Void.class );
+        if (config.isRest()) {
+            return makeHttpDeleteRequestAndCreateServiceResponse(baseURI + "/containers/" + id, Void.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new DisposeContainerCommand( id ) ) );
-            return (ServiceResponse<Void>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new DisposeContainerCommand(id)));
+            ServiceResponse<Void> response = (ServiceResponse<Void>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponsesList executeScript(CommandScript script) {
-        if( config.isRest() ) {
-            return makeHttpPostRequestAndCreateCustomResponse( baseURI + "/config", script, ServiceResponsesList.class );
+        if (config.isRest()) {
+            return makeHttpPostRequestAndCreateCustomResponse(baseURI + "/config", script, ServiceResponsesList.class);
         } else {
-            return executeJmsCommand( script );
+            ServiceResponsesList responsesList = executeJmsCommand(script);
+            if (!responsesList.getResponses().isEmpty() && shouldReturnWithNullResponse(responsesList.getResponses().get(0))) {
+                return null;
+            }
+            return responsesList;
         }
     }
 
     @Override
     public ServiceResponse<KieScannerResource> getScannerInfo(String id) {
-        if( config.isRest() ) {
-            return makeHttpGetRequestAndCreateServiceResponse( baseURI + "/containers/" + id + "/scanner", KieScannerResource.class );
+        if (config.isRest()) {
+            return makeHttpGetRequestAndCreateServiceResponse(baseURI + "/containers/" + id + "/scanner", KieScannerResource.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new GetScannerInfoCommand( id ) ) );
-            return (ServiceResponse<KieScannerResource>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new GetScannerInfoCommand(id)));
+            ServiceResponse<KieScannerResource> response = (ServiceResponse<KieScannerResource>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<KieScannerResource> updateScanner(String id, KieScannerResource resource) {
-        if( config.isRest() ) {
+        if (config.isRest()) {
             return makeHttpPostRequestAndCreateServiceResponse(
                     baseURI + "/containers/" + id + "/scanner", resource,
-                    KieScannerResource.class );
+                    KieScannerResource.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new UpdateScannerCommand( id, resource ) ) );
-            return (ServiceResponse<KieScannerResource>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new UpdateScannerCommand(id, resource)));
+            ServiceResponse<KieScannerResource> response = (ServiceResponse<KieScannerResource>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<ReleaseId> updateReleaseId(String id, ReleaseId releaseId) {
-        if( config.isRest() ) {
+        if (config.isRest()) {
             return makeHttpPostRequestAndCreateServiceResponse(
                     baseURI + "/containers/" + id + "/release-id", releaseId,
-                    ReleaseId.class );
+                    ReleaseId.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new UpdateReleaseIdCommand( id, releaseId ) ) );
-            return (ServiceResponse<ReleaseId>) executeJmsCommand( script, null, null, id ).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new UpdateReleaseIdCommand(id, releaseId)));
+            ServiceResponse<ReleaseId> response = (ServiceResponse<ReleaseId>) executeJmsCommand(script, null, null, id).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
     @Override
     public ServiceResponse<KieServerStateInfo> getServerState() {
-        if( config.isRest() ) {
+        if (config.isRest()) {
             return makeHttpGetRequestAndCreateServiceResponse(baseURI + "/state", KieServerStateInfo.class);
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new GetServerStateCommand() ) );
-            return (ServiceResponse<KieServerStateInfo>) executeJmsCommand( script).getResponses().get( 0 );
+            CommandScript script = new CommandScript(Collections.singletonList((KieServerCommand) new GetServerStateCommand()));
+            ServiceResponse<KieServerStateInfo> response = (ServiceResponse<KieServerStateInfo>) executeJmsCommand(script).getResponses().get(0);
+            return getResponseOrNullIfNoResponse(response);
         }
     }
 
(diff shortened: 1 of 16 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        commands.add(commandsFactory.newSetGlobal(LIST_NAME, new ArrayList<String>(), LIST_OUTPUT_NAME));
+        commands.add(commandsFactory.newStartProcess(PROCESS_ID));
+        commands.add(commandsFactory.newFireAllRules());
+
+        BatchExecutionCommand batchExecution = commandsFactory.newBatchExecution(commands, KIE_SESSION);
+
+        ruleClient.setResponseHandler(new FireAndForgetResponseHandler());
+        assertNoResponse(ruleClient.executeCommandsWithResults(CONTAINER_ID, batchExecution));
+
+        Thread.sleep(5000); // TODO use something else
```
rsynek: There are other tests using delay() method - it could be parametrized and reused. Out of curiosity - why exactly 5 sec?

livthomas: I forgot about this one. There is no particular reason why it is 5 seconds. It just needs to wait until the command is executed because `FireAndForgetResponseHandler` does not wait for the response. We should use some of `KieServerSynchronization` methods here instead of `Thread.sleep()`.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        ruleClient.setResponseHandler(asyncResponseHandler);
+
+        List<Command<?>> commands = new ArrayList<Command<?>>();
+        commands.add(commandsFactory.newSetGlobal(LIST_NAME, new ArrayList<String>(), LIST_OUTPUT_NAME));
+        commands.add(commandsFactory.newStartProcess(PROCESS_ID));
+        commands.add(commandsFactory.newFireAllRules());
+        commands.add(commandsFactory.newGetGlobal(LIST_NAME, LIST_OUTPUT_NAME));
+        BatchExecutionCommand batchExecution = commandsFactory.newBatchExecution(commands, KIE_SESSION);
+
+        assertNoResponse(ruleClient.executeCommandsWithResults(CONTAINER_ID, batchExecution));
```
mswiderski: this assertNoResponse is wrongly assuming that there will be response - the idea for async/fire and forget is it always returns null and then use callback to get the response. Even though the response handler internally returns NO_RESPONSE response it is then changed by the client to return null. You can see this in jbpm tests as well.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+        ExecutionResults result = responseCallback.get(ExecutionResults.class);
```
mswiderski: don't use interface to unmarshall data, use ExecutionResultImpl instead

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
livthomas: Yes, but the difference between `ProcessServicesClient` and `RuleServicesClient` methods is that the former one unwraps `ServiceResponse` and returns the object inside while the latter one  returns the result wrapped in `ServiceResponse`. That is why I think it should return this object with **NO_RESPONSE** response type. The same is true for `SolverServicesClient`.

The different behavior is in `KieServicesClient` where it returns `ServiceResponse` object with **NO_RESPONSE**. I think this is the right behavior so we can distinguish between async and sync calls if it is possible (it is not in `ProcessServicesClient` methods). What do you think?

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
livthomas: Thanks a lot @mswiderski. This test was failing with JSON marshaller because of this problem.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
mswiderski: the general approach is that when you use async response handler you should not expect any return value as it's simply not there thus I decided to always return null. If the KieServicesClient does return the ServiceResponse that is an overlook and should be fixed instead.

livthomas: What is then purpose of `NO_RESPONSE` response type? It is just our internal interpretation that should never get to the end user?

mswiderski: at the moment this is to allow internal components distinguish between different types of responses so various components can interpret it as needed - in this case return null as response of method call.

livthomas: Ok, I'll update the tests and try to fix `KieServicesClient`.

mswiderski: excellent, thank you @livthomas

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerSynchronization.java
```diff
@@ -115,6 +121,71 @@ public boolean conditionPassed() {
         });
     }
 
+    public static void waitForContainerWithReleaseId(final KieServicesClient client, final ReleaseId releaseId) throws Exception {
+        waitForCondition(new WaitingCondition() {
+            @Override
+            public boolean conditionPassed() {
+                ServiceResponse<KieContainerResourceList> containersList = client.listContainers();
```
sutaakar: You may use here listContainers(KieContainerResourceFilter containerFilter) to retrieve just containers with specific release id.

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/jms/ContainerJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,164 @@
+package org.kie.server.integrationtests.common.jms;
```
sutaakar: Missing copyright header.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,124 @@
+package org.kie.server.integrationtests.drools.jms;
```
sutaakar: Missing copyright header.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/jms/OptaPlannerJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
+package org.kie.server.integrationtests.optaplanner.jms;
```
sutaakar: Missing copyright header.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
```diff
@@ -138,15 +143,33 @@ public void testGetTasksUseOfAsyncResponseHandlerWithMarshaller() throws Excepti
...
+    }
+
+    @Test
+    public void testStartAndCompleteTaskUseOfFireAndForgetResponseHandler() {
+        testStartAndCompleteTask(new FireAndForgetResponseHandler());
+    }
+
+    @Test
+    public void testStartAndCompleteTaskUseOfAsyncResponseHandler() {
+        ResponseCallback callback = new BlockingResponseCallback(null);
```
sutaakar: You may also check responses from callback.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
```diff
@@ -138,15 +143,33 @@ public void testGetTasksUseOfAsyncResponseHandlerWithMarshaller() throws Excepti
...
+    public void testQueryRegistrationUseOfFireAndForgetReponseHandler() {
+        testQueryRegistration(new FireAndForgetResponseHandler());
+    }
 
-        List<TaskSummary> tasks = taskSummaryList.getItems();
-        assertNotNull(tasks);
-        assertEquals(1, tasks.size());
+    @Test
+    public void testQueryRegistrationUseOfAsyncReponseHandler() {
+        ResponseCallback callback = new BlockingResponseCallback(null);
```
sutaakar: You may also check responses from callback.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/jms/OptaPlannerJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+        assertThat(solver.getSolverId()).isEqualTo(SOLVER_1_ID);
+        assertThat(solver.getStatus()).isEqualTo(SolverInstance.SolverStatus.SOLVING);
+
+        assertNoResponse(solverClient.getSolverState(CONTAINER_1_ID, SOLVER_1_ID));
+        solver = responseCallback.get(SolverInstance.class);
+        assertThat(solver.getSolverId()).isEqualTo(SOLVER_1_ID);
+        assertThat(solver.getStatus()).isEqualTo(SolverInstance.SolverStatus.SOLVING);
+
+        solverClient.setResponseHandler(new RequestReplyResponseHandler());
+        KieServerSynchronization.waitForSolverStatus(solverClient, CONTAINER_1_ID, SOLVER_1_ID, SolverInstance.SolverStatus.NOT_SOLVING);
```
sutaakar: Just a question, is there any specific reason why do you wait for solver to finish?
Solver can be disposed even while running.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/jms/DroolsJmsResponseHandlerIntegrationTest.java
```diff
@@ -0,0 +1,124 @@
...
+        commands.add(commandsFactory.newStartProcess(PROCESS_ID));
+        commands.add(commandsFactory.newFireAllRules());
+
+        BatchExecutionCommand batchExecution = commandsFactory.newBatchExecution(commands, KIE_SESSION);
+
+        ruleClient.setResponseHandler(new FireAndForgetResponseHandler());
+        assertNoResponse(ruleClient.executeCommandsWithResults(CONTAINER_ID, batchExecution));
+
+        // wait for commands to be executed
+        Thread.sleep(500);
```
sutaakar: We cannot rely on Thread.sleep as it can fail on slower machines or application servers. It could be updated to waiting condition as is in KieServerSynchronization, for example as anonymous class.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
```diff
@@ -210,35 +232,114 @@ private void testGetProcessInstancesResponseHandler(ResponseHandler responseHand
...
+
+        List<TaskSummary> tasks = taskClient.findTasksAssignedAsPotentialOwner(USER_YODA, 0, 10);
+        assertThat(tasks).hasSize(1);
+
+        Long taskId = tasks.get(0).getId();
+        taskClient.setResponseHandler(responseHandler);
+        taskClient.startTask(CONTAINER_ID, taskId, USER_YODA);
+        taskClient.completeTask(CONTAINER_ID, taskId, USER_YODA, new HashMap<String, Object>());
+
+        delay();
```
sutaakar: Same issue as is with previous Thread.sleep.

livthomas:
@mswiderski, I have found one problem when trying to rewrite `KieServicesClientImpl` methods to return _null_ in case of **NO_RESPONSE**. There is a [method](https://github.com/droolsjbpm/droolsjbpm-integration/blob/6.5.x/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java#L195) which returns `ServiceResponsesList`. What should we do in this case?

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerSynchronization.java
livthomas: Thanks, I have rewritten it this way.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
livthomas: I think there is no point in doing this since all the operations executed in this test that use `AsyncResponseHandler` return `null`.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
livthomas: I think there is no point in doing this since all the operations executed in this test that use `AsyncResponseHandler` return `null`.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/jms/OptaPlannerJmsResponseHandlerIntegrationTest.java
livthomas: I think it should wait until the solving is finished in order to dispose the solver.  Do you think this assumption is wrong and we can skip it in order to save some time?

sutaakar: I think we can skip it and terminate solver while running. It will save us quite some time. It is already done in tests, see OptaplannerIntegrationTest.testTerminateEarly().

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
sutaakar: Right, good point.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
sutaakar: Right, good point.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsResponseHandlerIntegrationTest.java
livthomas: Sorry, they do not return `null` but their return type is `void` so there is no point in checking the response.

mswiderski:
@livthomas I'd say same thing here - should return nothing as there is nothing to return. So again, all returns from async handlers are empty so does not make sense to return any "fake" objects any way.

livthomas:
@sutaakar, thanks for your comments. I have applied some changes and added JMS transaction tests. Now there should be only one failing test method `testTransactionCommitWithoutCapabilities()` in `JmsTransactionsIntegrationTest`. We need to figure out how to retrieve server capabilities in case of JMS transactions are turned on because we cannot use `RequestReplyResponseHandler`. One possible solution would be to pass `UserTransaction` to `KieServicesConfiguration` and do it with `AsyncResponseHandler` in a transaction.

@mswiderski, when it comes to `ServiceResponsesList` return type of that method, I think it is not possible to return `null` because you do not know response types of those requests. Or should we look at the first one and act according to its value?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
```diff
@@ -123,6 +119,25 @@ private void init() {
...
+        ResponseHandler responseHandler = null;
+        if (config.getTransport() == KieServicesConfiguration.Transport.JMS) {
+            responseHandler = getResponseHandler();
+            setResponseHandler(new RequestReplyResponseHandler());
+        }
+
+        kieServerInfo = getServerInfo().getResult();
+        logger.debug("KieServicesClient connected to: {} version {}", kieServerInfo.getServerId(), kieServerInfo.getVersion());
+        List<String> capabilities = kieServerInfo.getCapabilities();
+        System.out.println(capabilities);
```
sutaakar: Accidental leftover?

livthomas: Yes, definitely. Thanks.

mswiderski:
@livthomas ServiceResponseList is a return type of single call to the server thus it will be handled by single responseHandler and in case of async it should still return null. Does that make sense? Or there is another case...

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
```diff
@@ -196,7 +215,11 @@ public ServiceResponsesList executeScript(CommandScript script) {
         if( config.isRest() ) {
             return makeHttpPostRequestAndCreateCustomResponse( baseURI + "/config", script, ServiceResponsesList.class );
         } else {
-            return executeJmsCommand( script );
+            ServiceResponsesList responsesList = executeJmsCommand( script );
+            if (!responsesList.getResponses().isEmpty() && shouldReturnWithNullResponse(responsesList.getResponses().get(0))) {
+                return null;
+            }
+            return responsesList;
```
livthomas: @mswiderski, did you mean something like this?

livthomas:
Jenkins retest this. It seems there has been some change made in jBPM repo recently which causes build failure.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
mswiderski: looks good to me

... and now I know what you meant :)

livthomas:
Since the problem with jBPM compilation has been fixed, Jenkins can retest this.

livthomas:
Jenkins retest this and do not go down this time :)

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/KieServerXAConnectionFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public XAConnection createXAConnection() throws JMSException {
+        return createXAConnection(TestConfig.getUsername(), TestConfig.getPassword());
+    }
+
+    @Override
+    public XAConnection createXAConnection(String userName, String password) throws JMSException {
+        try {
+            Method method = connectionFactory.getClass().getMethod("createXAConnection", String.class, String.class);
+            return (XAConnection) method.invoke(connectionFactory, userName, password);
```
livthomas: @sutaakar, this is an ugly workaround because ActiveMQ Artemis and HornetQ connection factories do not implement the same interface for creating XA connections but they have methods with the same signature. Please, try how this behaves on WebSphere and WebLogic. I am not sure about their JMS client implementations.

livthomas:
@mswiderski, look at the test failures now. Please, ignore those caused by not aborted processes and see the results of `JmsResponseHandlerIntegrationTest` and `JmsTransactionsIntegrationTest`.

1. `testGlobalConfigurationOfAsyncResponseHandler` fails because it is not possible to unmarshall `Long` value when using JSON and XStream. However, it works with JAXB.
2. `testTransactionCommitWithoutSettingCapabilities` fails because the client cannot retrieve a list of capabilities using `RequestReplyResponseHandler` when the transactions are turned on.This is expected since the transaction is never committed and we need to find the solution to this problem.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/KieServerXAConnectionFactory.java
sutaakar: WebLogic returns sympathetic error: "[JMSClientExceptions:055059]getXAResource can be called only from the server."
If you would like to test it locally you may use maven command: "mvn clean install -Dit.test=JmsTransactionsIntegrationTest -Poracle-wls-12 -Dweblogic.home=xxx"

mswiderski: why do you create new implementation of XAConnectionFactory? You should simply look it up from servers JNDI and no need to deal with any implementation specifics...

mswiderski:
@livthomas second error is expected as request reply should never be used with transactional client as that simply leads to a deadlock.

I'll look later today/tomorrow at the other failure

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
```diff
@@ -123,6 +119,30 @@ private void init() {
 
     }
 
+    private List<String> getCapabilitiesFromServer() {
+        if (config.isJms()) {
+            if (config.isJmsTransactional()) {
+                kieServerInfo = getServerInfoInTransaction();
```
livthomas: @mswiderski, what I mean is this case. The client simply cannot get the capabilities from the server unless it has an access to a transaction manager. So one solution might be to add a reference to `UserTransaction` to the configuration and get server info in a transaction.

mswiderski: the use case for transactional jms is to send message to server only when transaction completes and in async way as there won't be response until tx completes.
Moreover for async execution expected server capabilities needs to be provided as they can't be loaded automatically since the response will arrive later on. 

KieServicesClients need to be tx dependencies free as it might or might not run within tx context so it does not need to do any manual look ups itself. Thus the setting to make the jms delivery transactional so when ConnectionFactory is XA/JTA then it can deliver message only if transaction commits.

mswiderski:
@livthomas 
```
testGlobalConfigurationOfAsyncResponseHandler fails because it is not possible to unmarshall Long value when using JSON and XStream. However, it works with JAXB.
```

I believe you need to use JaxbLong.class instead

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/JmsTransactionsIntegrationTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        transaction.rollback();
+
+        List<ProcessInstance> processInstances = queryClient.findProcessInstances(0, 10);
+        assertThat(processInstances).isNotNull().isEmpty();
+    }
+
+    @Test
+    @Ignore
+    public void testTransactionCommitWithoutSettingCapabilities() throws Exception {
```
mswiderski: I would remove this test as it does not provide much of a value. Usually use case is that KieServicesClinet (that provides management capabilities) should be used in default request/reply way why the additional handlers are more for rules, processes etc.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/KieServerXAConnectionFactory.java
mswiderski: would recommend to try to look up remote JTA aware connection factory but if that becomes too difficult I'd say it does not bring much value to be tested on other containers than EAPI/wildfly. or at least not at the moment - mainly because this is standard JMS api and thus was tested multiple times on other places.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesClientImpl.java
```diff
@@ -123,6 +119,30 @@ private void init() {
 
     }
 
+    private List<String> getCapabilitiesFromServer() {
+        if (config.isJms()) {
+            if (config.isJmsTransactional()) {
```
mswiderski: I'd recommend at removing this as it does not provide much value and should not be used with async handlers on kie services client.

mswiderski:
@livthomas I added few comments, do you mind taking a look and let me know what you think?

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/jms/KieServerXAConnectionFactory.java
livthomas: Neither container provides XAConnectionFactory by default. We would have to configure it on each one of them. I will change the transaction tests so they will only be run on Wildfly/EAP (the way they are implemented right now) and will be skipped on other containers.

livthomas:
I have fixed all the things we discussed. Hopefully, the tests will finally pass.

livthomas:
Jenkins retest this (jBPM compilation problems should be fixed now)

mswiderski:
@livthomas #613 was just merged, could you please rebase this one so it gets tested?

livthomas:
Jenkins retest this

mswiderski:
now we're talking :)

+1 to merge, thanks @livthomas great job!

PR #607, opened 2016-08-30
Skip deploy for testing kjar

psiroky:
* there is no need to deploy
   the kjar as it's just used for
   testing

This should fix the deploy we are seeing. However, it's just quick&dirty fix. We need to come with a better solution.

cc @mariofusco, @baldimir


baldimir:
+1 for merge. Agree on that we need to find a better solution. Ideally the kjar shouldn't be part of the Maven reactor build and should be built and handled by the test that uses it.

PR #612, opened 2016-08-31
kie-server-tests: improve findProcessInstances tests

sutaakar:
findProcessInstances tests took into account number of completed process instances - assuming that there is no process instance finished. This made tests influenced by other tests. This problem was spotted in https://github.com/droolsjbpm/droolsjbpm-integration/pull/598.


mswiderski:
+1 to merge, test failures not related to this PR

PR #618, opened 2016-09-01
Configure kie.maven.settings.custom fro OSGi itests

psiroky:


psiroky:
So that did not work :( Let's try something else

mswiderski:
still valid or can be closed?

psiroky:
I am closing this as it did not fix the issue as expected. And we need better fix anyway.

PR #626, opened 2016-09-02
[JBPM-5288] process scanner config when creating new kieContainer

psiroky:


mswiderski:
+1 to merge

PR #627, opened 2016-09-05
JBPM-4650: Fix NPE when listing case definitions and instances

cristianonicolai:
Moved common invokeCaseOperation to abstract class.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-case-mgmt/pom.xml
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-case-mgmt/pom.xml
@@ -45,5 +45,10 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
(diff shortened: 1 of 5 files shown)


mswiderski:
@cristianonicolai is the NPE on building conversation header or somewhere else?

Just wondering why this [1] test does not fail at the moment...

[1] https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/jbpm/cases/CaseRuntimeDataServiceIntegrationTest.java#L171

cristianonicolai:
@mswiderski good point. I'm not quite sure why it doesn't fail there. The reason of NPE is a null ContainerId in here[1]. Which is always the case for both getCaseInstances and getCaseDefinitions.

[1] - https://github.com/droolsjbpm/droolsjbpm-integration/blob/2d791924c265edf1bba78add9f3fbccf332d8431/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java#L186

mswiderski:
+1 thanks @cristianonicolai

PR #629, opened 2016-09-05
RHBPMS - 2449 - XStream marshaller can't be configured with NameCoder, RHBPMS-2569 - addJaxbClasses method should be deprecated

mswiderski:
@etirelli in case you find a bit of time would be good to have your opinion on these two (skip the optaplanner as that is only to fix the tests... an attempt)

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
@@ -117,8 +117,13 @@ import org.optaplanner.core.api.score.buildin.simple.SimpleScore;
 import org.optaplanner.core.api.score.buildin.simplebigdecimal.SimpleBigDecimalScore;
 import org.optaplanner.core.api.score.buildin.simpledouble.SimpleDoubleScore;
 import org.optaplanner.core.api.score.buildin.simplelong.SimpleLongScore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class JaxbMarshaller implements Marshaller {
+
+    private static final Logger logger = LoggerFactory.getLogger(JaxbMarshaller.class);
+
     public static final Class<?>[] KIE_SERVER_JAXB_CLASSES;
 
     static {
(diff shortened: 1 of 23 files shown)


mswiderski:
@psiroky @ge0ffrey could you please take a look at this commit as it seems it fixed the issue with optaplanner tests failing recently in kie server https://github.com/droolsjbpm/droolsjbpm-integration/pull/629/commits/bb6692fd38682dab02f7c666836d531af729806f

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
```diff
@@ -277,26 +282,38 @@
...
 
             allClasses.addAll(Arrays.asList(KIE_SERVER_JAXB_CLASSES));
             if (classes != null) {
                 allClasses.addAll(classes);
             }
-
+            logger.debug("All classes for JAXB context are {}", allClasses);
             this.jaxbContext = JAXBContext.newInstance( allClasses.toArray(new Class[allClasses.size()]) );
         } catch ( JAXBException e ) {
+            logger.error("Error while creating JAXB Marshaller due to {}", e.getMessage(), e);
```
ge0ffrey: this will cause the same error to be logged twice: once here with this new log message and once by the default exception handler which should have the obligation to log all errors.

mswiderski: what do you mean by default exception handler?

ge0ffrey:
Looks ok - I presume that the XStream implementation still registers the Score Converters out-of-the-box? Score is an optaplanner class, much like String and Joda's LocalDate, etc. Users shouldn't need to add annotations to make it xstream nicely through the specialized ScoreXStreamConverters.

mswiderski:
@ge0ffrey all stays as it was, it was just split into separate methods so the can be overloaded by custom implementation.

Thanks for review @ge0ffrey

ge0ffrey:
+1 - the custom implementation will need to register those ScoreXStreamConverters itself (and it might register alternative implementations).

etirelli:
+1, I like the approach.

mswiderski:
thanks @etirelli @ge0ffrey merged manually

PR #630, opened 2016-09-05
[6.5.x] RHBPMS - 2449 - XStream marshaller can't be configured with NameCoder, RHBPMS-2569 - addJaxbClasses method should be deprecated 

mswiderski:
and here is the back port of #629


mswiderski:
jenkins retest this

mswiderski:
merged manually

PR #631, opened 2016-09-05
[DROOLS-898] rename drools-osgi* to kie-osgi*

psiroky:
* this is generally not backwards compatible and will
   be documented in the 6.x -> 7.x upgrade recipe. That being
   said there are only couple of classes which users may have
   directly used: mostly the OsgiKModuleBeanFactoryPostProcessor.
   Of course the GAV of the -features file changes as well, so
   that is something users will need to fix.

All related PRs:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/274
https://github.com/droolsjbpm/drools/pull/902
https://github.com/droolsjbpm/droolsjbpm-integration/pull/631

@mariofusco, @mswiderski, @winklerm, @jiripetrlik please take a look


mswiderski:
looks ok to me, though I let better osgi experts to comment as well

mariofusco:
+1

PR #633, opened 2016-09-06
JBPM-5304 - Process Instances: Parallel gateway stays active on merge…

mswiderski:
… in process diagram

@cristianonicolai do you mind taking a quick look?

the change here is to use different methods - two instead of one - to collect active and completed nodes separately as they provide better fit for this use case. They ensure only one entry for given node instance id is returned instead of multiple as it was done by the previously used method.

Tests are covering this in jbpm and those that we have in kieserver should be sufficient.


cristianonicolai:
+1

PR #634, opened 2016-09-06
[6.5.x] kie-server-tests: allow external configuration of excluded groups

sutaakar:
Needed for cases when external environment doesn't support specific functionality.


mswiderski:
+1 to merge

PR #635, opened 2016-09-06
[6.4.x] JBPM-5277 - getTasksAssignedAsPotentialOwner doesn't allow groups searching, only actorid

livthomas:
This fixes the problem with the list of accepted commands which was not updated before and causes fails of all requests with `GetTasksByGroupCommand`.

### DIFF
--- a/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/jaxb/AcceptedClientCommands.java
+++ b/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/jaxb/AcceptedClientCommands.java
@@ -86,6 +86,7 @@ public class AcceptedClientCommands {
         acceptedCommands.add(GetTaskAssignedAsPotentialOwnerCommand.class);
         acceptedCommands.add(GetTaskByWorkItemIdCommand.class);
         acceptedCommands.add(GetTaskCommand.class);
+        acceptedCommands.add(GetTasksByGroupCommand.class);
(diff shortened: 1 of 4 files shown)


livthomas:
It seems that this fixes the problem only partially.

mswiderski:
+1 to merge

could you please back port this to 6.5.x and master?

mswiderski:
@livthomas reading your comment - are you still working on the other part or need help with it? I'll wait with merge for now...

mswiderski:
@livthomas I believe you need to add it here as well https://github.com/livthomas/droolsjbpm-integration/blob/e86ee5803da4ba95ba95ee14af96cd0632ae35b9/kie-remote/kie-remote-client/src/main/java/org/kie/remote/client/jaxb/JaxbCommandsRequest.java#L129

triceo:
Although it would not be trivial, a test could be written to make sure that the two places contain the same elements. This seems to me like a major potential source of bugs.

livthomas:
@mswiderski I have added it to some other places where it should be but I need to test it in Business Central in order to verify it really works.

@triceo Actually, it is much more complicated as you can see from the latest fix.

livthomas:
I have just tested it in Business Central and it works.

mswiderski:
@livthomas could you please create PRs for upper branches?

livthomas:
@mswiderski Sure, I am working on it.

PR #641, opened 2016-09-07
[JBPM-5288,RHBPMS-4087] process scanner config when creating new kieC…

psiroky:
…ontainer

@jakubschwan this is needed in order to completely fix RHBPMS-4087. Without this the scanner config is not actually read during the startup (it is stored in the XML config, but ignored).


jakubschwan:
thanks for quick fix
+1

PR #642, opened 2016-09-07
[JBPM-5288,RHBPMS-4086] process scanner config when creating new kieC…

psiroky:
…ontainer (#626)


mswiderski:
+1

PR #650, opened 2016-09-13
removed useless dependency from drools-karaf-itests-kjar pom

mariofusco:


psiroky:
Although this is not needed at build time, I believe it is needed at runtime as the rules are importing classes from that model. WDYT?

mariofusco:
@psiroky I ran the tests and it worked for me. Let's wait for jenkins to finish building this.

mariofusco:
@psiroky as I thought all test passed. Let me know if I can merge this. Thanks

psiroky:
If the tests work, then it can be merged. But I don't really understand why the tests are passing :) I'll figure that out tomorrow.

+1

mariofusco:
I think that the test works because you don't need the dependency at compile time while it is resolved and downloaded from the maven repository by kie-ci at runtime.

PR #653, opened 2016-09-15
[DROOLS-1294] fix kjar root path for both spring and blueprint integration when running on Windows

mariofusco:


kie-aries-blueprint/src/main/java/org/kie/aries/blueprint/namespace/KieObjectsInjector.java
```diff
@@ -180,12 +180,12 @@ private InternalKieModule createOsgiKModule(KieModuleModel kieProject) {
 
     protected void addKieModuleToRepo(KieModuleModel kieModuleModel) {
         String rootPath = configFilePath;
-        if ( rootPath.lastIndexOf( ':' ) > 0 ) {
+        if ( rootPath.lastIndexOf( ':' ) >= 2 ) {
```
psiroky: I think it would be worth noting why we `2` here.

mariofusco: Do you mean that I could just add a comment or were you thinking to something else?

psiroky: Yes, comment would be great.

PR #655, opened 2016-09-20
[6.5.x] kie-server-tests: fix response handler tests on slow databases

livthomas:
I have fixed some tests that caused problems on slower databases.


livthomas:
@sutaakar, if you see test failures in our test suite, it is because of these tests.

sutaakar:
Thanks for info.
Looks good.

mswiderski:
@livthomas thanks, can we have this for master too?

livthomas:
@mswiderski, yes, I will send a pull-request as soon as I run fixed tests on various databases and verify that there are no more failures.

PR #657, opened 2016-09-20
[JBPM-5330] xstream marshalling: use values set by default constructo…

psiroky:
…r instead of nulls

 * XStream by default overrides the default values set by constructor
   with nulls - in case the XML string did not contain any date
   about the particular field. In this specific case, if the container
   filter was not set, it was set to null (instead of using the default
   constructor to supply the ACCEPT_ALL filter)

 * this should also align the marshalling more closely with the JAXB

This could in theory break a backwards compatibility, but I could not think of such case.

@mswiderski could you please quickly review this?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -16,11 +16,11 @@
 package org.kie.server.api.marshalling.xstream;
 
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
 import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
 import org.drools.core.runtime.help.impl.XStreamXML;
 import org.kie.server.api.commands.CallContainerCommand;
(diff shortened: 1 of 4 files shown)


psiroky:
It seems the test failure is _not_ related to these changes. The same test fails also for this PR: https://github.com/droolsjbpm/droolsjbpm-integration/pull/656

mswiderski:
this might help I believe https://github.com/droolsjbpm/droolsjbpm-integration/pull/655 when forward ported to master

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -16,11 +16,11 @@
 package org.kie.server.api.marshalling.xstream;
 
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
 import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
```
psiroky: Sorry, I am not sure I understand...the field is final, so it needs to the value needs to be set via constructor. 

The other option would be to have a custom XStream converter. But that would also mean that we would need to keep an eye on what needs a converter and what doesn't. This seems very error prone to me.

I am open to other suggestions for sure - the `PureJavaReflectionProvider` I can think of right now.

mswiderski: alright, that's fine just was wondering if we could initialize that differently than final/constructor way. But if you feel that's the best way then I am good with this. Assuming that this PureJavaReflectionProvider is public api so to say.

psiroky: I believe immutability is good :) This way, we can also fix everything at one place (in the xstream object), instead of having to add custom code to multiple classes.

I haven't found any mention on the PureJavaReflectionProvider being private API, so I assumed its fine to use it.

mswiderski:
fair enough @psiroky 

+1

PR #658, opened 2016-09-21
[6.4.x] [JBPM-5288,RHBPMS-4087] kie-server: scanner status not retained after…

jakubschwan:
… restart in container list

After restart of server kie scanner is not in container list. There is kie scanner after get container by id.

@psiroky wdyt?


psiroky:
We should add a test for this. I'll look into creating one.

psiroky:
I've created https://github.com/droolsjbpm/droolsjbpm-integration/pull/665 which should fix the issue in a slightly different way. Let me know what you think.

mswiderski:
+1

PR #659, opened 2016-09-21
kie-server-tests: fix response handler tests on slow databases

livthomas:
Here is the same fix for `master` branch as the one introduced in pull-request #655 for `6.5.x`.


mswiderski:
+1 to merge

PR #660, opened 2016-09-21
[JBPM-5330,RHBPMS-4279] xstream marshalling: use values set by defaul…

psiroky:
…t constructor instead of nulls

 * XStream by default overrides the default values set by constructor
   with nulls - in case the XML string did not contain any date
   about the particular field. In this specific case, if the container
   filter was not set, it was set to null (instead of using the default
   constructor to supply the ACCEPT_ALL filter)

 * this should also align the marshalling more closely with the JAXB

Backport of #657.


mswiderski:
+1

PR #663, opened 2016-09-21
[JBPM-5327,RHBPMS-4273] fix tests on JDK6 - use correct jaxb impl ver…

psiroky:
…sion

 * the version 2.2.11 should be used everywhere. This is the one we
   support. In case no jaxb impl is available on classpath, JDK will
   use the one available together with the JDK installation (which is
   slightly different for each JDK version). So using own version makes
   sure we always test and run against the exactly same libs

### DIFF
--- a/kie-server-parent/kie-server-api/pom.xml
+++ b/kie-server-parent/kie-server-api/pom.xml
@@ -72,6 +72,17 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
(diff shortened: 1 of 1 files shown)


mswiderski:
+1 to merge

@psiroky do we need this on master as well?

psiroky:
Technically we don't need this on master, because we no longer support JDK 6 there. However, I believe it would be more correct to test with these jaxb impl jars as well (since those will eventually used by the kie-server war as well). I'll prepare PR for master.

PR #665, opened 2016-09-22
[JBPM-5288,RHBPMS-4087] manage scanner lifecycle directly by KieContainerInstanceImpl

psiroky:
* the refactoring also makes sure the scanner status is always
   consistent between InternalKieScanner and KieScannerResource
   (both as fields of KieContainerInstanceImpl)

This is an alternative solution to https://github.com/droolsjbpm/droolsjbpm-integration/pull/658 (with more refactorings).

@jakubschwan, @mswiderski, @etirelli please let me know what you think.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieContainerInstanceImpl.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieContainerInstanceImpl.java
@@ -17,12 +17,15 @@ package org.kie.server.services.impl;
 
 import org.drools.compiler.kie.builder.impl.InternalKieContainer;
 import org.drools.compiler.kie.builder.impl.InternalKieScanner;
-import org.kie.server.api.model.KieContainerResource;
-import org.kie.server.api.model.KieContainerStatus;
-import org.kie.server.api.model.ReleaseId;
+import org.kie.api.KieServices;
 import org.kie.server.api.marshalling.Marshaller;
 import org.kie.server.api.marshalling.MarshallerFactory;
 import org.kie.server.api.marshalling.MarshallingFormat;
+import org.kie.server.api.model.KieContainerResource;
+import org.kie.server.api.model.KieContainerStatus;
+import org.kie.server.api.model.KieScannerResource;
+import org.kie.server.api.model.KieScannerStatus;
+import org.kie.server.api.model.ReleaseId;
 import org.kie.server.services.api.KieContainerInstance;
 
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieContainerInstanceImpl.java
```diff
@@ -49,11 +51,39 @@ public KieContainerInstanceImpl(String containerId, KieContainerStatus status, I
         super();
         this.kieContainer = kieContainer;
         this.resource = new KieContainerResource( containerId, null, status );
+        // set the default scanner state to DISPOSED (which is the actual default state)
+        // this way we don't need to do null checks all around for the scanner resource
```
psiroky: Here I am not 100% sure this is true. Is there any difference between `DISPOSED` and null scanner? Do we care if the scanner was ever set before (and thus is not null), and if it has not been touched at all (and thus is null)?

psiroky:
jenkins retest this

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieContainerInstanceImpl.java
etirelli: Makes sense

etirelli:
+1 to merge, looks good

mswiderski:
@psiroky can you resolve conflicts?

psiroky:
@mswiderski sure, I'll do that. This PR was meant as a replacement for #658 (which got merged), but I forgot to clarify that...

mswiderski:
so shall this still be merged or?

psiroky:
From my point of view, yes. I've tried to refactor a code a bit to make the scanner state a bit more consistent. If you could take a quick and let me know what you think, that would be great.

mswiderski:
it looks good to me

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieContainerInstanceImpl.java
```diff
@@ -84,16 +114,49 @@ public KieContainerResource getResource() {
...
+        this.getResource().getScanner().setStatus(KieScannerStatus.STOPPED);
+    }
+
+    public void disposeScanner() {
+        if (this.scanner == null) {
+            throw new IllegalStateException("Can not dispose non-existing (null) scanner!");
+        }
+        this.scanner.shutdown();
+        this.scanner = null;
+        this.getResource().setScanner(new KieScannerResource(KieScannerStatus.DISPOSED));
```
jakubschwan: What about to add here also scanNow and createScanner method?? Then in KieServerImpl class would be a similar use of scanner in private methods.

psiroky: Yes, that should be possible. I added just those needed to fix the sync issues, but this would make it more consistent, so I'll add those as well. I'll also see if we could remove the getScanner() method to avoid any possible mutation by 3rd party objects.

psiroky: This should be fixed now

jakubschwan: +1

psiroky:
This should be ready to merge now.

mswiderski:
+1

psiroky:
Thanks @mswiderski, I'll prepare PRs for 6.5.x and master.

PR #667, opened 2016-09-26
[6.5.x][JBPM-5288,RHBPMS-4086] manage scanner lifecycle directly by KieConta…

psiroky:
…inerInstanceImpl

 * the refactoring also makes sure the scanner status is always
   consistent between InternalKieScanner and KieScannerResource
   (both as fields of KieContainerInstanceImpl)

Forward-port of #665

PR #668, opened 2016-09-26
[JBPM-5288] manage scanner lifecycle directly by KieContainerInstanceImpl

psiroky:
* the refactoring also makes sure the scanner status is always
   consistent between InternalKieScanner and KieScannerResource
   (both as fields of KieContainerInstanceImpl)

Forward port of #667


etirelli:
+1

PR #671, opened 2016-09-29
[6.5.x] RHBPMS-4317 - AsyncWorkItemHandler fails to complete after timer fire…

mswiderski:
…s in Spring environment

@krisv would do the honors?


krisv:
+1

PR #672, opened 2016-09-29
RHBPMS-4317 - AsyncWorkItemHandler fails to complete after timer fire…

mswiderski:
…s in Spring environment


mswiderski:
merging as it's corresponding PR was already merged into 6.5.x

PR #677, opened 2016-10-06
[6.4.x] DROOLS-987 - Errors in Phreak under heavy and multi threaded load - f…

mswiderski:
…ixed tests as rule node fires rules directly

(cherry picked from commit 4ec6f14438709211b0cc347a91cecf24c5b1a460)


mswiderski:
merging as requested for one off patch

PR #678, opened 2016-10-07
JBPM-4646 - Case management - support for reopening the case

mswiderski:
depends on https://github.com/droolsjbpm/jbpm/pull/629

@sutaakar as we talked about it the other day :) can you take a look?

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseManagementServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseManagementServiceBase.java
@@ -160,6 +160,22 @@ public class CaseManagementServiceBase {
         }
     }
 
+    public void reopenCase(String caseId, String containerId, String caseDefinitionId, String payload, String marshallingType) {
+        containerId = context.getContainer(containerId, new ByCaseIdContainerLocator(caseId)).getContainerId();
+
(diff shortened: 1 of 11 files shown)


kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseManagementServiceBase.java
```diff
@@ -160,6 +160,22 @@ public void cancelCaseInstance(String containerId, String caseId, boolean destro
         }
     }
 
+    public void reopenCase(String caseId, String containerId, String caseDefinitionId, String payload, String marshallingType) {
+        containerId = context.getContainer(containerId, LatestContainerLocator.get()).getContainerId();
```
sutaakar: Just a question, shouldn't be here rather ByCaseIdContainerLocator? As case file may contain objects specific to container version.

sutaakar:
Looks good :)
Just had one minor question.

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseManagementServiceBase.java
mswiderski: I was thinking about this and though it would make sense to always start process from latest deployment but after giving it a second thought (and your comment) I think you're right as it would be safer to stay in same container as the case itself. Will update it

mswiderski:
jenkins retest this

PR #679, opened 2016-10-07
kie-server-tests: manage javassist dependency

sutaakar:
To allow setting custom javassist version for product tests.


psiroky:
@sutaakar just out of the curiosity - why do you need to set custom javassist version for product tests?

sutaakar:
Because javassist is brought to dependency tree by org.reflections:reflections, which is brought by optaplanner-core. In usual pom structure javassist version is defined by jboss-integration-platform-parent, but since our kjars don't have this parent then version is defined in org.reflections:reflections. This version isn't included in in product repos, causing product builds to fail.

psiroky:
Thanks for the detailed comment @sutaakar!

PR #683, opened 2016-10-10
[6.5.x] don't fail on kie scanner when not set and requested to stop

mswiderski:
@psiroky could you please double check as this most likely will be a blocker for 6.5

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
@@ -566,8 +566,8 @@ public class KieServerImpl {
         List<Message> messages = getMessagesForContainer(id);
         messages.clear();
         if (kci.getScanner() == null) {
-            return new ServiceResponse<KieScannerResource>(ServiceResponse.ResponseType.FAILURE,
-                    "Invalid call. Scanner is not instantiated. ",
+            return new ServiceResponse<KieScannerResource>(ResponseType.SUCCESS,
+                    "Scanner is not started.",
                     getScannerResource(kci));
         }
         if (KieScannerStatus.STARTED.equals(mapScannerStatus(kci)) ||


psiroky:
This is a valid fix if we want to be more user-friendly, e.g. if user calls stop scanner and the scanner does not really exist, then we don't have to fail.

+1 for the change even if it changes the current behavior. I think there may be other places similar to this one where we could more relaxed checks.

One thing I don't understand is how the call to stopScanner() happens when creating the container...

mswiderski:
@psiroky stopScanner is invoked because when you create container from workbench by default it has scanner stopped as it was not requested to be started. So when you createContainer in kie server it first checks if the scanner resource is given as part of request:
https://github.com/mswiderski/droolsjbpm-integration/blob/9235824a270b10b82e6de2dc894ccf4e3bb249a1/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java#L258
then it looks what status the resource has:
https://github.com/mswiderski/droolsjbpm-integration/blob/9235824a270b10b82e6de2dc894ccf4e3bb249a1/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java#L485
and since it is stopped status it then attempts to stop it here but it assumes it's started so actual scanner instance in KieContainerInstance should be there, though it's not because container is just being created:
https://github.com/mswiderski/droolsjbpm-integration/blob/9235824a270b10b82e6de2dc894ccf4e3bb249a1/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java#L568

psiroky:
Thanks for explanation @mswiderski. It seems that either we should fix the workbench code to not set stopped state for new containers, or we need to refine the scanner lifecycle (e.g. the stopped state can be set anytime regardless of the current state of the scanner).

mswiderski:
@psiroky I think it is perfectly valid that workbench starts container with scanner being set to stopped as that is default - start container but not scanner. So kie server should simply skip that as scanner for newly created container is anyway stopped.

psiroky:
My understanding was that default state is disposed, so that's why I did not consider the stopped as an option for the new container. 

That being said, I agree that we should relax the checks on the inputs where it makes sense. So stopping non-existing container is something we should allow as it does really change state, and would not lead to any inconsistencies.

PR #685, opened 2016-10-13
[JBPM-5366] remove no longer used kie-remote modules

psiroky:
* kie-remote-common is being used by kie-server,
   so can not be removed at this point

@mswiderski could you please take a look?


mswiderski:
@psiroky I see you left kie-remote-common in same place, is there any reason to do so? Maybe we can move these classes to kie server client as most likely that is the only place that requires it, wdyt?

I think that we need to remove the kie-remote-* from distribution: 
https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/droolsjbpm-bpms-distribution/src/main/assembly/pre-bin.xml#L55-L58
https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/droolsjbpm-bpms-distribution/src/main/assembly/pre-bin.xml#L112-L115

not on vpn so can't see why the build failed...

psiroky:
Ah, right, those distributions are only enabled for full profile, that's why I missed it. Will fix that.

I was not sure what to do with kie-remote-common. It is being used (or at least being specified in pom) by multiple kie-server modules, e.g. `kie-server-client`, `kie-server-rest-common`, `kie-server-services-case-mgmt`. I believe we could move it under the kie-server directory as a `kie-server-common` module?

mswiderski:
@psiroky good point to move it kie server common instead so it then belongs to kie server which would then own that module.

psiroky:
@mswiderski I've moved the module under `kie-server-parent/kie-server-common` + changed the package names from `org.kie.remote` to `org.kie.server`. I'll double check the downstream repositories, but none of them should actually depend on `kie-remote-common` (if they do, we need to fix it).

mswiderski:
excellent work @psiroky 
+1

is it ok to be merged or you still working on anything related?

psiroky:
I've created three more related PRs in other repos:
https://github.com/droolsjbpm/drools-wb/pull/317
https://github.com/droolsjbpm/kie-docs/pull/179
https://github.com/droolsjbpm/kie-docs/pull/179

mswiderski:
drools-wb is now merged so merging this as well, thanks a lot @psiroky

PR #689, opened 2016-10-20
JBPM-4646 - Case management - assign correlation key for each process…

mswiderski:
… instance - fixed test assertion

@psiroky here is the fix for the failing tests


psiroky:
@mswiderski great, thanks! Let's wait for the Jenkins.

PR #694, opened 2016-10-24
JBPM-5293 - Remove Lists of group names from 'getGroupsForUser( ... )…

mswiderski:
…' method signature in UserGroupCallback interface

depends on https://github.com/droolsjbpm/jbpm/pull/652

@krisv @psiroky  and the last one from this work.


krisv:
+1

PR #695, opened 2016-10-25
PLANNER-661 PLANNER-604 Execution Server: SolverInstance.score doesn't use the ScoreJaxbXmlAdapaters?

mcimbora:
Introduce ScoreWrapper to make sure the unmarshaller has enough information to convert string representation of the score into object representation.

Provide toScore() method with the support of out-of-box score implementations.

This PR also includes removal of deprecated Solution interface occurrences.

@ge0ffrey Could you please have a look at this?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
@@ -1,0 +1,65 @@
+package org.kie.server.api.model.instance;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlValue;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import com.thoughtworks.xstream.annotations.XStreamConverter;
+import com.thoughtworks.xstream.converters.extended.ToAttributedValueConverter;
+import org.optaplanner.core.api.score.Score;
+import org.optaplanner.core.impl.score.ScoreUtils;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XStreamConverter(value = ToAttributedValueConverter.class, strings = { "scoreString" })
+public class ScoreWrapper {
+
+    @XmlAttribute(name = "scoreClass")
+    @XStreamAlias(value = "scoreClass")
+    private Class<? extends Score> scoreClass;
+
+    @XmlValue
+    private String scoreString;
+
+    // Define default constructor to enable class marshalling/unmarshalling
+    private ScoreWrapper() {
+    }
+
+    public ScoreWrapper( Score score ) {
+        this.scoreClass = score == null ? null : score.getClass();
+        this.scoreString = score == null ? null : score.toString();
+    }
+
+    public Class<? extends Score> getScoreClass() {
+        return scoreClass;
+    }
+
+    public String getScoreString() {
+        return scoreString;
+    }
+
+    /**
+     * Returns score representation of the object.
+     *
+     * @return Score representation of the object. Returns null if the score has not been assigned by the solver yet.
+     * @throws IllegalArgumentException If <code>scoreClass</code> is not one of the out-of-box score implementations. In this case
+     *                                       clients may implement their own way to extract the score object.
+     */
+    public Score toScore() {
+        if ( scoreClass == null ) {
+            return null;
+        }
+
+        return ScoreUtils.parseScore( scoreClass, scoreString );
+    }
+
+    @Override
+    public String toString() {
+        return "ScoreWrapper{" +
+                "scoreClass='" + scoreClass + '\'' +
+                ", scoreString='" + scoreString + '\'' +
+                '}';
+    }
+
+}
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
@@ -20,10 +20,10 @@ import org.junit.Test;
 import org.kie.api.KieServices;
 import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.ServiceResponse;
+import org.kie.server.api.model.instance.ScoreWrapper;
 import org.kie.server.api.model.instance.SolverInstance;
 import org.kie.server.api.model.instance.SolverInstanceList;
 import org.kie.server.client.KieServicesException;
-import org.optaplanner.core.api.domain.solution.Solution;
 import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
 
 import java.lang.Thread;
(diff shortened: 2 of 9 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
```diff
@@ -0,0 +1,107 @@
...
+import org.optaplanner.core.api.score.buildin.simplebigdecimal.SimpleBigDecimalScore;
+import org.optaplanner.core.api.score.buildin.simpledouble.SimpleDoubleScore;
+import org.optaplanner.core.api.score.buildin.simplelong.SimpleLongScore;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XStreamConverter(value = ToAttributedValueConverter.class, strings = { "scoreClass" })
+public class ScoreWrapper {
+
+    @XmlAttribute(name = "scoreClass")
+    @XStreamAlias(value = "scoreClass")
```
ge0ffrey: I'd prefer not to mention the name/value, so it's always the same as the fieldName (which it is). But maybe the rest of KIE server does do this like this style. Either way it's a soft suggestion.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return Score representation of the object.
+     * @throws UnsupportedOperationException If <code>scoreClass</code> is not one of the out-of-box score implementations. In this case
+     *                                       clients may implement their own way to extract the score object.
+     */
+    public Score toScore() {
+        if ( scoreClass == null ) {
+            return null;
+        }
+
+        if ( BendableScore.class.equals( scoreClass ) ) {
```
ge0ffrey: Order of these if's should be the same order as
  https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/config/score/definition/ScoreDefinitionType.java#L19
so it's easier to spot if one is missing.

ge0ffrey: Note that one might become missing once I add an new one upstream and forgot it here. At some point we'll add a HardMediumSoft one as there's one missing.

We should probably push these code into optaplanner-core. I am thinking a static method on ScoreDefinition or ScoreUtil or so.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -140,7 +140,7 @@ public void testGetSolverState() throws Exception {
         assertEquals( SOLVER_1_ID, returnedInstance.getSolverId() );
         assertEquals( SolverInstance.getSolverInstanceKey( CONTAINER_1_ID, SOLVER_1_ID ), returnedInstance.getSolverInstanceKey());
         assertEquals( SolverInstance.SolverStatus.NOT_SOLVING, returnedInstance.getStatus());
-        assertNull( returnedInstance.getScore() );
+        assertNull( returnedInstance.getScoreWrapper() );
```
ge0ffrey: also assert that returnedInstance.getScoreWrapper().getScore() is not null.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
mcimbora: @ge0ffrey Correct, I followed the style that's used kie-server-wide.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
mcimbora: @ge0ffrey This assertion checks whether the score wrapper is null. This happens in cases where the score is null, as there's a little point in returning an empty score wrapper instance.

mcimbora:
@ge0ffrey I updated the PR, thanks for the review :-)

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
ge0ffrey: Looking at it from a client perspective, I 'd argue that the ScoreWrapper instance should be null if and only if the method I just called doesn't fill it in because it's not part of its contract (think javadoc). If it did fill it in, but the score is null because the solver hasn't assigned it yet, then the ScoreWrapper should be empty.

Why? Think of this code:
response = clientApi.callService(...);
Score score = response.getSolverInstance().getScoreWrapper().toScore(); // NPE otherwise!
if (score == null) {
   print("No score assigned yet");
}

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Returns score representation of the object.
+     *
+     * @return Score representation of the object.
+     * @throws UnsupportedOperationException If <code>scoreClass</code> is not one of the out-of-box score implementations. In this case
+     *                                       clients may implement their own way to extract the score object.
+     */
+    public Score toScore() {
+        if ( scoreClass == null ) {
+            return null;
```
ge0ffrey: This should fail fast, because the constructor guarantees that scoreClass is never null.

So when it is null, that means JAXB or XStream messed up, and will want to fail fast here, as soon as possible. Otherwise this null will leak and potentially end up as garbage in the user's database for months in production before they notice this bug.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
mcimbora: @ge0ffrey Makes more sense now seeing it from a different angle (not just from XML perspective). The score wrapper is now always assigned, no matter if the score is null or not.
For this purpose I removed the null check from `ScoreWrapper` constructor, as it's valid to construct an instance which encapsulates null score.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/ScoreWrapper.java
mcimbora: @ge0ffrey With current changes, it's valid to encapsulate a null score.

ge0ffrey:
+1 to merge, waiting for Jenkins CI

mcimbora:
The build error is unrelated.

mcimbora:
Lets rerun the tests once https://github.com/droolsjbpm/droolsjbpm-integration/pull/725 is integrated.

mbiarnes:
Jenkins retest this please

mcimbora:
Jenkins retest this

PR #704, opened 2016-10-27
improved searching for svg file within kjar for displaying process image

mswiderski:
@krisv as we discussed to improve the searching mechanism for svg files.


krisv:
+1

PR #705, opened 2016-10-28
JBPM-5336 - Support for rolling updates in kie server - improved supp…

mswiderski:
…ort for pluggable ContainerLocator impls

@etirelli here comes some improvements in handling container alias to make it pluggable + some fixes and tests for drools part of dealing with alias.


etirelli:
+1 looks good @mswiderski

PR #706, opened 2016-10-28
JBPM-5388 - Policy support in KIE Server

mswiderski:
@etirelli whenever you have a chance, please have a look. it should support automatic dispose of containers and much more by additional policies implementations which are completely pluggable. 

Will do an article about the concepts and features next week.

@sutaakar the tests might be unstable as they are difficult to properly assert and the policy evaluation is time based... if you have any ideas on how to improve them I'd be more than happy to apply :)

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.services.impl.policy;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.kie.server.api.model.ServiceResponse;
+import org.kie.server.services.api.KieServer;
+import org.kie.server.services.api.KieServerRegistry;
+import org.kie.server.services.api.Policy;
+import org.kie.server.services.impl.KieContainerInstanceImpl;
+import org.kie.server.services.impl.controller.DefaultRestControllerImpl;
+import org.kie.server.services.impl.locator.LatestContainerLocator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Policy that will dispose older container and leave only latest one - latest according to version of ReleaseId.
+ *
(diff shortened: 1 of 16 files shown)


kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
```diff
@@ -0,0 +1,131 @@
...
+                logger.debug("Containers for alias {} are already on expected level (number of containers is {})", alias, containerInstances.size());
+                continue;
+            }
+
+            String latestContainerId = LatestContainerLocator.get().locateContainer(alias, containerInstances);
+            final Map<String, String> report = new HashMap<>();
+            containerInstances.stream()
+                    .filter(kci -> !kci.getContainerId().equals(latestContainerId))
+                    .forEach(kci -> {
+                        ServiceResponse<Void> response = kieServer.disposeContainer(kci.getContainerId());
```
etirelli: Is this disposeContainer() call safe? Will it allow ongoing requests to finish executing before disposing?

etirelli:
@mswiderski great job, looks good! +1

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
etirelli: No, you are correct. We have to make the dispose() call safe (in case it is not) and not put the burden on the policy. +1

mswiderski:
jenkins retest this

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public void start() {
+        logger.debug("Starting KeepLatestContainerOnlyPolicy policy...");
+        TimeUnit timeUnit = TimeUnit.valueOf(System.getProperty(INTERVAL_TIME_UNIT, TimeUnit.MILLISECONDS.toString()));
+        long givenInterval = Long.parseLong(System.getProperty(INTERVAL_VALUE, "0"));
+
+        if (givenInterval > 0) {
+            timeUnit.convert(givenInterval, timeUnit);
```
sutaakar: Is this line used for anything? Interval is computed in next line.

sutaakar:
@mswiderski Tests looks ok, they should be hopefully stable. I tried them to run several times and it passed every time. In case of any instability we can fix them later.

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
mswiderski: Very good point, that needs to be removed

mswiderski:
Thanks @sutaakar, what I don't really like is that the policy config is set globally and not scoped to test. At the moment (where we have only one) that's not big issue but when more will come into picture that will become difficult to test in isolation. Any idea how we could improve it? I guess starting/stopping app or server is not good approach...

sutaakar:
@mswiderski I think that for these cases it would be better to use unit tests, probably on kie-server-services level. It is faster and simpler solution than trying to test it on whole app server. We can keep KeepLatestOnly policy there to check that all integration works in app servers.

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

PR #707, opened 2016-11-01
RHBPMS-4374: FormURLGenerator class contains old reference to kie-wb.…

pefernan:
Changed reference to wb home page.


pefernan:
@mswiderski yes, it can be merged

PR #708, opened 2016-11-02
[6.5.x] kie-server-tests: test improvement for Oracle RAC

jakubschwan:
Some tests failed on Oracle RAC database due id ordering from db.


mswiderski:
jenkins retest this

jhrcek:
@jakubschwan using conditional logic in tests, like `if (processInstanceId < processInstanceId2)` is considered bad practice. If the order of `TaskSummary` in the resulting Lists is not important it would be better to use different way to assert presence based on processInstanceId. E.g. using hamcrest matchers (included in JUnit4) or some stuff from AssertJ (since this module already has assertJ as dependency).

Something along the lines of 
`assertThat(tasks).exracting("processInstanceId").hasSize(2).contains(id1, id2);`

jakubschwan:
@jhrcek thanks, that's a good point. But these tests results are sorted by ID, so order in resulting Lists is important.

mswiderski:
@jakubschwan @jhrcek are there any other planned changes for these tests or this is good to be merged?

jhrcek:
@mswiderski not from my side

jakubschwan:
@mswiderski can be merged

PR #710, opened 2016-11-03
RHBPMS-4435 - Add test to reproduce issue

jakubschwan:
@mswiderski when I tried to use bigger poll-interval (10000) then this test pass

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerIntegrationTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.kie.server.api.KieServerEnvironment;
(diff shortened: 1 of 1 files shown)


mswiderski:
this interval is in milliseconds so giving it 20 ms is not really good when it comes to remote interaction as the scanner will constantly scan when being stopped. So in my eyes that's not really meaningful test as it's kind of abuse the scanner use case. @mariofusco wdyt?

jakubschwan:
Sure, set scanner to this small interval is not good. But community docs say, that interval is specified in seconds (https://docs.jboss.org/drools/release/latest/drools-docs/html/ch22.html#d0e25010) and also 20 ms interval is used in REST example of post command to scanner (https://docs.jboss.org/drools/release/latest/drools-docs/html/ch22.html#d0e25325)

mariofusco:
@mswiderski @jakubschwan I fixed docs on both master ( https://github.com/droolsjbpm/kie-docs/commit/a571343f41bef5d7e8771f310f25d60a34157216 ) and 6.5.x branch ( https://github.com/droolsjbpm/kie-docs/commit/d4c7f42816b70a278956a4224ee8427092b40bd3 ).

I agree with @mswiderski that setting the timeout to 20ms is not a good idea for that use case, but I'm also not sure if that test should fail (I don't think so). I'll give it a run asap.

mswiderski:
this is going to be fixed once the https://github.com/droolsjbpm/droolsjbpm-integration/pull/723 is merged, just tried locally

@psiroky let me know if I should help with changes mentioned in https://github.com/droolsjbpm/droolsjbpm-integration/pull/723 or you'll take care of this

psiroky:
@mswiderski I am just looking at that other PR and will take care of it.

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

psiroky:
jenkins retest this

PR #713, opened 2016-11-03
[do not merge yet] JBPM-4646: extend Case test coverage

sutaakar:
Some more info:
- changes contains group of //TODO marks, these needs to be clarified first. One widespread question: if Kie server gets request containing wrong container id or case id what it should return? Should it return error or empty result? For example see testGetActiveProcessInstancesNotExistingCase()
- See also sorting extensions for Case client - is sorting implemented for all methods which should have it? Is it missing somewhere else? Is sorting pointless for some methods where it is implemented?

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/CaseServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/CaseServicesClient.java
@@ -31,6 +31,15 @@ import org.kie.server.api.model.instance.ProcessInstance;
 
 public interface CaseServicesClient {
 
+    public static final String SORT_BY_CASE_DEFINITION_ID = "CaseId";
+    public static final String SORT_BY_CASE_DEFINITION_NAME = "CaseName";
+    public static final String SORT_BY_CASE_DEFINITION_DEPLOYMENT_ID = "Project";
+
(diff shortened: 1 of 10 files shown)


mswiderski:
> One widespread question: if Kie server gets request containing wrong container id or case id what it should return? 

if there is no case then an error should be thrown - CaseNotFoundException - 404

> See also sorting extensions for Case client - is sorting implemented for all methods which should have it? Is it missing somewhere else? Is sorting pointless for some methods where it is implemented?

sorting only makes sense for potential big results where the benefit of doing server side sorting makes sense.

I'll go over all tests/remarks and fix what is missing/broken. Thanks @sutaakar

mswiderski:
> // TODO: how this should work? Looks like destroy case/reopen doesn't work as expected.
it's not possible to reopen case after it was destroyed 

> // TODO what does mandatory mean?
it's a meta data to indicate its importance though not really enforced by the engine atm

> // TODO: should it return also case id, maybe also container id? So we will know where fragments belong.
no, I don't think so as that would be duplication of data, as to be able to get adhoc fragments you need to provide case id and container so sending it again does not make much sense in my opinion

> // TODO which should be correct?
completionMessage should be empty string in case it's not set

mswiderski:
this is replaced by https://github.com/droolsjbpm/droolsjbpm-integration/pull/716

PR #717, opened 2016-11-07
[do not merge - reproducer] JBPM-5398: reproducer

sutaakar:
Steps to reproduce (reproducer against postgresql 9.3 DB and EAP6.4):
- download EAP6.4, prepare postgre DB
- fill datasource URL, username and password according to your configuration
- execute reproducer from kie-server-integ-tests-jbpm module:  mvn clean install -Peap64x -Deap64x.download.url=<EAP6.4 download URL> -Dit.test=ProcessServiceIntegrationTest
 
I cleaned up ProcessServiceIntegrationTest so there is just testProcessInstanceWithTimer() left, which is failing. Raised time duration to be sure that timeout doesn't clash with any other access on slow DBs.


mswiderski:
closing as this was used to find and analyze issue but it's not going to be merged, thanks @sutaakar

PR #718, opened 2016-11-08
JBPM-4646 - Case management - extended comment test coverage and fixes

sutaakar:


sutaakar:
jenkins retest this

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-case-mgmt/src/main/java/org/kie/server/remote/rest/casemgmt/Messages.java
```diff
@@ -24,5 +24,7 @@
 
     public static final String CASE_INSTANCE_ACTIVE = "Case with id \"{0}\" is already started/reopened";
 
+    public static final String CASE_COMMENT_NOT_FOUND = "Could not find case comment with id \"{0}\" in container \"{1}\"";
```
mswiderski: I don't think container is relevant here, better to include case id rather than container

sutaakar: Good point, updated.

PR #719, opened 2016-11-09
JBPM-5403: work item for per process instance test coverage

sutaakar:
Contains partial fix - fixed getWorkItem() functionality.
Minor refactoring to avoid duplicities.


mswiderski:
@sutaakar I think the other case - when per process instance throws the SessionNotFound for context error we should transform it to ProcessInstanceNotFound exception

let me know if you'd like to take care of that too in jbpm services

sutaakar:
@mswiderski I will try to take care about that

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
@sutaakar I believe we are good with this to be merged after your improvements in jbpm that were already merged, correct?

sutaakar:
@mswiderski Right.

PR #723, opened 2016-11-11
[DROOLS-1357] kie-server: always return container resource with refre…

psiroky:
…shed releaseIds

@mswiderski, @etirelli one thing I am wondering about is the existence of the method `KieContainerInstanceImpl#getResource` since it may return stale data. I think we should just rename the `getRefreshedResource()` to `getResource()` and make sure to always return refreshed resource. WDYT?

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/api/KieContainerInstance.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/api/KieContainerInstance.java
@@ -34,8 +34,6 @@ public interface KieContainerInstance {
 
     KieContainerResource getResource();
 
-    KieContainerResource getRefreshedResource();
-
     KieScanner getScanner();
 
     Marshaller getMarshaller(MarshallingFormat format);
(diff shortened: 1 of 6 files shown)


mswiderski:
just to clarify, initially there was issue with the getResource that did dispose the marshallers etc so that's why the refreshed version was added, but later on it was further refactored to call the dispose marshallers explicitly when needed (update version etc)

so we still have the issue if scanner updates the kie container, the marshallers should be refreshed as well, though right this won't happen. So I believe we would have to have some kind of event listener support in kie scanner to be notified whenever scanner did update the version.

Why we could do until we have the listener in place is to check if the version of the resource in KieContainerInstanceImpl is same as one returned from KieCOntainer and if not, dispose marshallers. That's not a nice solution but at least will keep it up to date.

Anyway, @psiroky agree, these two methods are now redundant and as you said one of them might return out of date information. Could you please remove getRefreshedResource method and move the logic into getResource?

psiroky:
Thanks for the details @mswiderski, I was not aware of the issue with marshallers. I'll remove the `getRefreshedResource` method and then also look at disposing the marshallers when needed.

psiroky:
@mswiderski I did the changes. Could you please check if this is what you had in mind?

psiroky:
jenkins retest this

mswiderski:
@psiroky looks good to me, +1 once the build is green

@sutaakar we need to rerun the tests for decreasing performance after this is merged - RHBRMS-2608

sutaakar:
@mswiderski Sure, will keep that in mind

sutaakar:
Performance test revealed a bug in the implementation, fixed by https://github.com/droolsjbpm/droolsjbpm-integration/pull/732

PR #724, opened 2016-11-11
[DROOLS-1353] fix build: remove dependency on drools-simulator which …

psiroky:
…is not currently build


psiroky:
Closing as it needs further fixes in -distribution which bundles the examples (which are now disabled as well).

PR #725, opened 2016-11-14
[DROOLS-1353] temporarily remove examples from distribution assembly

mariofusco:


psiroky:
I guess it would be good to add a comment saying why the stuff is commented out.

Also, should we at least include the other binaries? With most of the assembly file commented out, the resulting zip will just contain sh|bat scripts, for examples, which are not there. I think we should either completely disable the distribution or at least add back the other binaries. WDYT?

PR #727, opened 2016-11-15
[6.4.x] RHBRMS-2621 - [GSS] (6.3.z) Errors in Phreak under heavy and multi th…

mswiderski:
…readed load

depends on https://github.com/droolsjbpm/jbpm/pull/675
@krisv and additional test updates after chasing default behavior for rule node


krisv:
+1

PR #729, opened 2016-11-15
[DROOLS-1357, RHBPMS-4438] kie-server: always return container resour…

psiroky:
…ce with refreshed releaseIds

Backport of #723.

@mswiderski I am not sure if it is safe to remove the `getRefreshedResource()` in patch release. It seems it should be just internal interface, but is there a chance users could be using it as well?

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
@@ -126,6 +126,17 @@
     </dependency>
 
     <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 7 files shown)


mswiderski:
@psiroky I don't think this is used by anyone so in my opinion it's completely ok to be merged

I guess we need the same for 6.5.x branch

psiroky:
We will need to wait with 6.5.x branch, as it should be part of the first patch release, not the GA. I'll be tracking this and create the PR once it is "safe".

sutaakar:
@psiroky please incorporate here https://github.com/droolsjbpm/droolsjbpm-integration/pull/732

psiroky:
@sutaakar good catch, thanks. I guess we could merge this and use your PR for the other fix?

sutaakar:
@psiroky sure

mswiderski:
@psiroky I am not sure we should add things to lower branches that does not exist in higher. So we should either at to both or to none of them.

psiroky:
@mswiderski I've created https://github.com/droolsjbpm/droolsjbpm-integration/pull/737 to make I don't forget. It should be merged after the 6.4 GA release as we need only need it for 6.4.x

mswiderski:
does this include the fix done by @sutaakar?

psiroky:
@mswiderski now it does. Thanks for spotting that!

PR #731, opened 2016-11-16
removed version that does not upgrade automatically

mbiarnes:


drools-osgi/drools-karaf-itests-kjar/pom.xml
```diff
@@ -14,7 +14,6 @@
   <groupId>org.drools</groupId>
```
psiroky: While we are at it, the groupId could be removed as well. But its not strictly necessary.

mbiarnes: will do this asap

mswiderski:
jenkins retest this

PR #733, opened 2016-11-16
kie-osgi tests: use single Karaf instance per test class to speed up …

psiroky:
…the build

I got about 50% improvement when running the tests locally.

@mariofusco, @jiripetrlik, @winklerm WDYT?


jiripetrlik:
+1

PR #737, opened 2016-11-21
[DROOLS-1357,RHBPMS-4437] kie-server: always return container resourc…

psiroky:
…e with refreshed releaseIds

Backport of #723. Should be merged for 6.4.1.


psiroky:
I believe this should be ready to merge now.

PR #739, opened 2016-11-21
broker example session configuration to use rule node as wait state

mswiderski:
@tarilabs here is configuration change for the broker example. could you give it a try when you have a sec?


tarilabs:
Before:
image

After:
image

Thank you!

PR #740, opened 2016-11-22
JBPM-5418: Signal event does not work on deployment on per process in…

stathisp:
…stance runtime strategy

### DIFF
--- a/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/cdi/ProcessRequestBean.java
+++ b/kie-remote/kie-remote-services/src/main/java/org/kie/remote/services/cdi/ProcessRequestBean.java
@@ -325,11 +325,11 @@ public class ProcessRequestBean {
         // process instance
         if( processInstanceId == null && cmd instanceof ProcessInstanceIdCommand ) {
             processInstanceId = ((ProcessInstanceIdCommand) cmd).getProcessInstanceId();
-            if( processInstanceId != null ) {
+            if( processInstanceId != null && processInstanceId != -1L) {
                return ProcessInstanceIdContext.get(processInstanceId);
             }
         }
-        if( processInstanceId != null ) {
+        if( processInstanceId != null && processInstanceId != -1L) {
             return ProcessInstanceIdContext.get(processInstanceId);
         }
 


mswiderski:
ok to test

mswiderski:
@livthomas do we have tests for kie remote in community already? or it's outside...

livthomas:
@mswiderski, no, the tests are in private QE repository. But I can manually build `kie-remote-services` with this fix and let you know if the problem is really gone.

Anyway, we need to create a product Jira and have it approved for the next patch release before merging this. I can take care of that too.

mswiderski:
@livthomas excellent, please do that so we can merge it once approved as it makes sense to include it

livthomas:
@stathisp What exactly is your use case? Do you have any simple reproducer for this issue?

Although I was able to reproduce the bug, your commit seems not to fix it.

stathisp:
@livthomas  I've just used ksession.signalEvent(signalRef, payLoadDataObject) to a process that contains start signal event using KIE Remote Java API. In, this case the processInstanceId in the attached class is -1 and now it returns EmptyContext instead of ProcessInstanceContext. You still get 'No context found for session -1'?

mswiderski:
the exception is gone, but was the signal received by anything?

stathisp:
I had deployed my process at Deploy->Process Deployments with PER_PROCESS_INSTANCE strategy. My process has a start signal event with name 'customerStartsApplication' and a simple request object. From the client, I send:

2016-11-23 21:52:19 TRACE org.kie.services.client.api.command.AbstractRemoteCommandObject     - Sending SignalEventCommand via POST to http://localhost:8080/business-central/rest/execute
2016-11-23 21:52:19 TRACE org.kie.services.client.api.command.AbstractRemoteCommandObject     - Serialized JaxbCommandsRequest:
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?><command-request><deployment-id>drive:RemedialBPMN:1.2</deployment-id><ver>6.4.0.1</ver><user>drools</user><signal-event process-instance-id="-1" event-type="customerStartsApplication"><event xsi:type="request" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><requestId>90</requestId><requestedAmount>0.0</requestedAmount></event></signal-event></command-request>
2016-11-23 21:52:19 DEBUG org.kie.services.client.api.command.AbstractRemoteCommandObject     - Sending POST request with SignalEventCommand to http://localhost:8080/business-central/rest/execute

and my process receives the signal and starts. This works in 6.4.0.

stathisp:
2016-11-23 22:05:34 TRACE org.kie.services.client.api.command.AbstractRemoteCommandObject     - Sending SignalEventCommand via POST to http://localhost:8080/business-central/rest/execute
2016-11-23 22:05:34 TRACE org.kie.services.client.api.command.AbstractRemoteCommandObject     - Serialized JaxbCommandsRequest:
 &lt;?xml version="1.0" encoding="UTF-8" standalone="yes"?&gt;&lt;command-request&gt;&lt;deployment-id&gt;drive:RemedialBPMN:1.2&lt;/deployment-id&gt;&lt;ver&gt;6.4.0.1&lt;/ver&gt;&lt;user&gt;drools&lt;/user&gt;&lt;signal-event process-instance-id="-1" event-type="customerStartsApplication"&gt;&lt;event xsi:type="request" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"&gt;&lt;requestId&gt;90&lt;/requestId&gt;&lt;requestedAmount&gt;0.0&lt;/requestedAmount&gt;&lt;/event&gt;&lt;/signal-event&gt;&lt;/command-request&gt;
2016-11-23 22:05:34 DEBUG org.kie.services.client.api.command.AbstractRemoteCommandObject     - Sending POST request with SignalEventCommand to http://localhost:8080/business-central/rest/execute

livthomas:
I can confirm that this commit fixes the problem with start signal event. But at the same time it hides the problem with intermediate signal event which is not fixed. I would suggest that we further investigate this and provide a more complex fix for this issue.

mswiderski:
I'd say we should then separate these two issues - this one to be renamed and fixed as in this PR to solve problem when starting process by signal regardless of the strategy, and another one - which will be much more complex and might not be worth addressing for 6.x anymore. wdyt?

livthomas:
+1 We would need this commit anyway. However, we should also fix the problem with intermediate signal events for 6.x but let's discuss it somewhere else.

PR #741, opened 2016-11-23
kie-server-tests: improve Case dynamic task coverage

sutaakar:
Rest tests are failing:
- testAddDynamicWorkItemTaskNoTaskSpec fails because CaseManagementServiceBase suppose that unmarshalled taskSpecificationMap is not null. May be good to return Response.Status.BAD_REQUEST in that case with description what is missing there.
- testAddDynamicWorkItemTaskNotExistingContainer fails because Kie server service returns IllegalArgumentException for missing container which isn't handled in case resource.

@mswiderski What do you think about it?

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/filtered-resources/kjars-sources/case-insurance/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/filtered-resources/kjars-sources/case-insurance/pom.xml
@@ -13,6 +13,15 @@
   <version>1.0.0.Final</version>
   <name>case-insurance</name>
 
+  <dependencies>
+    <dependency>
+     <groupId>org.kie</groupId>
+     <artifactId>kie-api</artifactId>
+     <version>${version.org.kie}</version>
+     <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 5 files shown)


mswiderski:
> testAddDynamicWorkItemTaskNoTaskSpec fails because CaseManagementServiceBase suppose that unmarshalled taskSpecificationMap is not null. May be good to return Response.Status.BAD_REQUEST in that case with description what is missing there.

+1, we should respond with bad request in that situation

> testAddDynamicWorkItemTaskNotExistingContainer fails because Kie server service returns IllegalArgumentException for missing container which isn't handled in case resource.

is this only to handle the exception?

sutaakar:
Was thinking about exception handling in Kie server a little bit.....
How we should distinguish between different failures in Kie server?

For example for case represented by testAddDynamicWorkItemTaskNoTaskSpec we can check unmarshalled taskSpecificationMap and if we find that there is missing parameter we can throw IllegalArgumentException which will be converted to HTTP response code 400 in REST endpoint. 

Then there is for example testAddDynamicWorkItemTaskNotExistingContainer, where we throw same IllegalArgumentException when we try to locate proper container [1]. In this case it will be more appropriate to return HTTP response code 404 as case id (which wasn't found) is part of request URL.

I think that Kie server could use some specific exceptions to detect fail states in own code like missing container or case.
What do you think about it?

[1] https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/locator/ByCaseIdContainerLocator.java#L67

mswiderski:
incorporated in https://github.com/droolsjbpm/droolsjbpm-integration/pull/753

PR #743, opened 2016-11-29
kie-server-tests: improve case stage coverage

sutaakar:
One question about active nodes included - to be clarified.

Fail in testGetCaseStages - stage, which should be completed, has Active status. Caused by https://github.com/droolsjbpm/jbpm/blob/master/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java#L489 where new instance of CaseStageInstanceImpl has status set as Active by default .

Fail in testCompleteCaseStageAndAbort - task is available even after case is cancelled - seems to happen when case was in different stage than user task while closing.
Side question - what should happen to user tasks if they aren't started/completed and case is closed or moved to another stage?


mswiderski:
incorporated in https://github.com/droolsjbpm/droolsjbpm-integration/pull/753

PR #744, opened 2016-11-30
JBPM-5435: JaxbMap modifies collection passed as constructor argument

livthomas:
This fixes problems described in [JBPM-5435](https://issues.jboss.org/browse/JBPM-5435).

### DIFF
--- a/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/model/type/JaxbMapTest.java
+++ b/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/model/type/JaxbMapTest.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.api.model.type;
+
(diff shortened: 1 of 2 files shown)


kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/model/type/JaxbMapTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class JaxbMapTest {
+
+    private static final String KEY = "variable1";
+    private static final Object VALUE = new Date();
+
+    @Test
+    public void testImmutableMap() {
+        Map<String, Object> entries = Collections.singletonMap(KEY, VALUE);
+        JaxbMap jaxbMap = new JaxbMap(entries);
+        assertThat(jaxbMap.getEntries()).containsKey(KEY);
```
livthomas: This test fails with `UnsupportedOperationException` on line 35 if you use the previous implementation of `JaxbMap`. This is because `JaxbMap` tried to modify immutable map.

triceo: Thanks for explaining - would be nice if the test case already came with such a comment. My other comment is probably irrelevant now.

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/model/type/JaxbMapTest.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Test
+    public void testEntryNotModified() {
+        Map<String, Object> entries = new HashMap<>();
+        entries.put(KEY, VALUE);
+        JaxbMap jaxbMap = new JaxbMap(entries);
+        assertThat(jaxbMap.getEntries()).containsKey(KEY);
+        assertThat(entries).containsEntry(KEY, VALUE);
```
livthomas: Although the two tests are quite similar, they show different problems which you could have experienced with the previous implementation of `JaxbMap`. While the first one shows an exception thrown in case immutable map is passed as an argument to `JaxbMap` constructor, the second one shows that if you pass modifiable map, `JaxbMap` will change its content.

PR #747, opened 2016-12-02
New exception type for kie server http related errors

cristianonicolai:
To allow clients to better handle the different type of errors that could be thrown while executing http requests to the Kie Server.


cristianonicolai:
@mswiderski, created this to allow me to handle the different errors on the UI like unauthorized, etc. That way I can display specific messages based on the different error codes.

mswiderski:
jenkins retest this

PR #748, opened 2016-12-02
Fixes after JPA 2.0 to 2.1 upgrade

psiroky:
* custom settings.xml needed for builds
   that use "-Dmaven.repo.local", as otherwise
   KIE Maven support looks into the default ~./m2
   repo where the artifacts are not present

 * the mapping-file "META-INF/ProcessInstanceInfo.hbm.xml"
   is not needed (likely just copy&paste error). Hibernate
   was complaining it can not find the file and would refuse
   to initialize

Please don't merge yet. All the IP BOM 7.0.0.CR5 PRs need to be merged together.

### DIFF
--- a/kie-aries-blueprint/pom.xml
+++ b/kie-aries-blueprint/pom.xml
@@ -96,7 +96,7 @@
 
     <dependency>
       <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
       <optional>true</optional>
     </dependency>
     <dependency>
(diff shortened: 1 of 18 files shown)


mswiderski:
Jenkins retest this

mswiderski:
@psiroky these test failures seems to be either connected to dependency that does not support spa 2.1 or some misconfiguration. Any ideas?

psiroky:
@mswiderski I am not sure, will take a closer look.

psiroky:
It seems the failures are caused by an old aries-blueprint-jpa bundle version which does not support JPA 2.1 (only 2.0). Not really how to fix that, we may have to upgrade to newer Karaf version to fix that.

mswiderski:
or fir the time being downgrade persistence.xml version for the tests that fail so we can move on with upgrade to hibernate 5

psiroky:
I temporarily ignored those two failing tests as Aries Blueprint does not support JPA 2.1. I created https://issues.jboss.org/browse/DROOLS-1380 to track this.

psiroky:
@mswiderski ok, let me try that. That would could a better solution

psiroky:
@mswiderski I've tried to downgrade the JAP version in that one `persistence.xml`, but that did not work. I am not exactly sure why though. For now I've ignored the tests and created JIRA to investigate these failures. Could you please approve this PR is this OK (for the record)?

PR #754, opened 2016-12-12
Align reactivity support for Collection interfaces.

tarilabs:
Depends on droolsjbpm/drools#1033 .

### DIFF
--- a/kie-maven-plugin/src/main/java/org/kie/maven/plugin/BytecodeInjectReactive.java
+++ b/kie-maven-plugin/src/main/java/org/kie/maven/plugin/BytecodeInjectReactive.java
@@ -240,15 +240,6 @@ public class BytecodeInjectReactive {
         LOG.debug("buildWriteInterceptionBodyFragment: {} {}", field.getType().getClass(), field.getType());
         
         if ( isCtFieldACollection(field) ) {
-            // if ctField is a collection we may need to wrap..
-            if ( field.getType().subtypeOf(cp.get(ReactiveCollection.class.getName())) ) {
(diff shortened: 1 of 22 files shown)


tarilabs:
```
Test Result (1 failure / +1)
org.kie.server.integrationtests.common.rest.RestMalformedRequestIntegrationTest.testCreateContainerNonExistingGAV2[JAXB]
```

Jenkins, test again :8ball:

psiroky:
IMO we really should not merge PRs until they are "green". Without successful build, we can't be sure if the PR won't break the build.

mswiderski:
+1 for this, without build results we can't asses the impact as well.

tarilabs:
I agree but in this case the modification of all commits in the PR are affecting just the `kie-maven-plugin`
for bytecode instrumentation (with optional parameter to be set in order to "trigger" the feature) while the failed test:
```
org.kie.server.integrationtests.common.rest.RestMalformedRequestIntegrationTest.testCreateContainerNonExistingGAV2[JAXB]
```
is about the kie server.

The earlier commit a8f7ee3 is ticked green and subsequent d9688d1 is just a micro optimization pertaining the same scope (of the `kie-maven-plugin`) hence we agreed with @mariofusco to just merge anyway.

psiroky:
I disagree. If the test is failing, then it needs to be fixed regardless if this PR is causing the fail or not. If you don't know how to fix it, then contact someone who can. If it is randomly failing then it needs to be stabilized. If wee keep doing this, the tests will never get stabilized.

PR #750, opened 2016-12-06
RHBPMS-4514 - Null Pointer exception occurs for Business Process with…

tsurdilo:
… empty Embedded SubProcess

PR #755, opened 2016-12-13
RHBPMS-4514 - Null Pointer exception occurs for Business Process with…

tsurdilo:
… empty Embedded SubProcess (#750)

### DIFF
--- a/jbpm-simulation/src/main/java/org/jbpm/simulation/PathContext.java
+++ b/jbpm-simulation/src/main/java/org/jbpm/simulation/PathContext.java
@@ -19,7 +19,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
-
 import org.eclipse.bpmn2.FlowElement;
 import org.eclipse.bpmn2.SequenceFlow;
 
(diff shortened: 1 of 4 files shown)


jbpm-simulation/src/main/java/org/jbpm/simulation/PathContext.java
```diff
@@ -15,10 +15,7 @@
 
 package org.jbpm.simulation;
 
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
psiroky: I believe we should not be using star imports.

tsurdilo:
@psiroky sorry will update. ide change it.

tsurdilo:
@psiroky amended

jbpm-simulation/src/main/java/org/jbpm/simulation/PathContext.java
```diff
@@ -15,14 +15,15 @@
 
 package org.jbpm.simulation;
 
+
+import org.eclipse.bpmn2.FlowElement;
+import org.eclipse.bpmn2.SequenceFlow;
```
psiroky: Nit-picking, but this is still wrong. This does not follow the `java.*` + `javax.*` first (which is what we should use), nor the alphabetical order (which may be colliding with what we use).

tsurdilo:
@psiroky ok. amended too.

tsurdilo:
jenkins retest this

PR #760, opened 2016-12-16
[6.5.x][RHBPMS-4534] enable REST deploy without request body

psiroky:


psiroky:
@mswiderski could you please take a look at this?

mswiderski:
only comment is about these two classes that says don't change it will be generated by maven plugin... shall this be committed that way?

so now we have option to deploy with and without descriptor, right?

+1 in general, we could ask our friends from QE to check on WAS and WLS so it won't break anything there...

psiroky:
Those two classes should not really depend on some maven plugin to change the code, but I figured its not worth the time fixing this properly since we don't have this code on master anymore. Basically every time you do a `mvn clean install` those files get updated and git notices that. This should be the last change there as we won't be upgrading the version on this branch anymore.

With the new method added, we will allow EAP7 users to deploy without descriptor. This was working for some reason with EAP6 (likely because of RestEasy 2.x and 3.x differences).

Yes, It would be indeed great to check this on WAS and WLS as well. @mswiderski do you know whom to ask directly?

mswiderski:
@psiroky I think last time it was @livthomas who dealt with WLS/WAS testing of workbench. Though that might have changed too...

livthomas:
I have not tested the fix provided by @psiroky but having there `deploy` method twice caused some problems on WebLogic in the past. See [RHBPMS-644](https://issues.jboss.org/browse/RHBPMS-644) and the [commit](https://github.com/droolsjbpm/droolsjbpm-integration/commit/b96a1205) that fixed that issue.

psiroky:
Thanks @livthomas! Seems to me like walking in circles :) So either it won't work on EAP 7 (without this fix) or it will won't work on WebLogic (with this fix). I will take another look to see if there is any other option which would work on both....

mswiderski:
@psiroky are you going to work on this more? Meaning it should not be merged yet....

psiroky:
@mswiderski if I understand correctly, we can not merge this as it would break WebLogic. I would first like to finish different JIRA (https://issues.jboss.org/browse/RHBPMS-4435), then I can look if there is any other option...

psiroky:
I currently don't see an easy way to make this work on both EAP7 and WebLogic.

We would probably need two different classes, one for EAP7 and the second one for WebLogic (and then we would need to test the other containers carefully and pick the right class for those as well). That seems like a huge maintenance burden to me, and I am not sure if it is worth it.

mswiderski:
@psiroky what do you think about having two endpoints /deploy
- POST that will not expect to have deployment descriptor
- PUT that will require deployment descriptor
I believe most of the time people would use POST anyway without deployment descriptor and that way we make it as simple as possible for most of users. Then if you'd like to override deployment settings you would use the PUT version. Does it make any sense?

psiroky:
That sounds good, considering the current situation. My only worry is that this will likely break the backwards compatibility? I have no idea though how many users may be using the POST + custom deployment descriptor.

mswiderski:
I'd say that's the only thing we can do and even if someone is using it right now, the only change people would have to do is change the HTTP method which is trivial in any language possibly being used for this.

livthomas:
@mswiderski What about having it the other way round? I think it makes more sense to keep the current endpoint with deployment descriptor as it is and just add a new one without deployment descriptor parameter using PUT method. Therefore, we would not break backward compatibility and EAP7 users will be able to use this new endpoint. What do you think?

mswiderski:
the reason why I recommended to use POST without deployment descriptor is because most of people don't even know it's there and no one is sending deployment descriptor. But either way is fine with me, maybe GSS can comment on this as well giving their preference.

livthomas:
@mswiderski Oh, I see. In that case the solution proposed by you is probably better since it would be the same for all containers. Moreover, deployment descriptor parameter is not mentioned in the documentation so I would assume nobody uses it.

psiroky:
@mswiderski, @livthomas thanks for the insights! I will go with the original solution Maciej proposed - removing the deployment descr. parameter from the POST method and adding new PUT method with the param.

psiroky:
jenkins retest this

psiroky:
@mswiderski could you please take one more look at this? I've basically removed the param from the POST method and added new PUT method which consumes the parameter.

psiroky:
The test failures seem to be related to the POI upgrade, but let's wait with merging this until the issue is resolved and all the tests are passing.

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
Jenkins retest this

psiroky:
@mswiderski I was trying to follow the Swagger annotations which mention which formats are supported. If that's not correct, we need to fix it there as well.

psiroky:
Jenkins retest this

psiroky:
According to https://docs.jboss.org/jbpm/release/6.5.0.Final/jbpm-docs/html/ch17.html#remote.additional all  REST operations, except `execute`, should support both XML and JSON.

mswiderski:
excellent, thanks @psiroky

PR #761, opened 2016-12-19
[6.5.x] Fix KIE Server output for not registered extensions

jakubschwan:
Cherry pick from master of PR #758

jBPM-UI KIE Server extension can't be registered without a jBPM extension.


mswiderski:
we need to have RHBPMS jira for this and most likely customer case that requires this to be included in 6.5.x. Personally I don't think we should add it here as it changes the KieServerExtension interface.

triceo:
+1 on what @mswiderski said - it is far too late to introduce such changes to 6.5.x.

PR #764, opened 2016-12-21
[DROOLS-1384] Objects in a List is not correctly unmarshalled by JSON…

tkobayas:


mswiderski:
incorporated in https://github.com/droolsjbpm/droolsjbpm-integration/pull/765

@tkobayas if you find some time please give it a try, I do believe it should now work with both versions - where all custom types are wrapped or only the top level. Though priority for this to work is work where all types are wrapped.

PR #765, opened 2016-12-23
DROOLS-1384 - Objects in a List is not correctly unmarshalled by JSONMarshaller

mswiderski:


mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

tkobayas:
Thank you @mswiderski , I confirmed that it works for "having type information only on top level object" case as well. FYI, this is the test case I confirmed. You can add it to MarshallingRoundTripCustomClassListTest if you want.
https://gist.github.com/tkobayas/1fac568d9544fe894c919b558f07421f

mswiderski:
Excellent, thanks for confirming. Will include the test as well

PR #771, opened 2016-12-30
RHBPMS-4557: NPE when searching nonexistent task using the method Use…

jesuino:
…rTaskServicesClient.findTaskById


mswiderski:
Could you please add a test case for this?

PR #776, opened 2017-01-02
RHBPMS-4557: NPE when searching nonexistent task using the method Use…

jesuino:
…rTaskServicesClient.findTaskById

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
@@ -286,7 +286,10 @@ public class ConvertUtils {
     }
 
     public static TaskInstance convertToTask(UserTaskInstanceDesc userTask) {
-
+    	if (userTask == null) {
+    		return null;
+    	}
+    	
         TaskInstance instance = TaskInstance.builder()
                 .id(userTask.getTaskId())
                 .name(userTask.getName())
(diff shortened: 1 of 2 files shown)


psiroky:
Do we really need full integration test for this? Integration tests are quite time consuming, so we should try to create unit tests for simple fixes like this. WDYT?

sutaakar:
I think it is suitable for this case. There is also checked that endpoint will correctly process not existing task and return proper response and client will handle the response accordingly.

mswiderski:
agreed that it does make sense to have this as integration test. Actually I think that cause we have this as integration test we got these failures captured that are not handled same for JMS, @jesuino would you mind taking a look?

mswiderski:
jenkins retest this

MarianMacik:
After a little bit of research I have found that in JMS there is no exception thrown on the server side so the response is marked as [SUCCESS](https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JBPMKieContainerCommandServiceImpl.java#L145) since null is returned. On the client side, since it is a SUCCESS, it is simply returned. To have the same behaviour I would suggest to throw an exception like with [UserTaskServiceBase](https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java#L450) so it is handled as a failure for both REST and JMS.

So what do you think?

Marian

MarianMacik:
After a discussion with @psiroky we agreed that I will create a new PR for this with the same handling as in the case of client's getTaskInstance method (handled via exceptions).

jesuino:
I noticed that it would require changes on JMS side and how it deals with exceptions so I was supposed to talk to you about it, but I see @MarianMacik took over this issue, thanks, Marin!

mswiderski:
replaced (and already merged) by https://github.com/droolsjbpm/droolsjbpm-integration/pull/797

PR #781, opened 2017-01-04
[6.5.x] RHBPMS-4548 - RuntimeDataService 'getTaskAssignedAsPotentialOwner' ru…

mswiderski:
…ns single threaded and therefore doesn't scale

depends on https://github.com/droolsjbpm/jbpm/pull/711


sutaakar:
looks good

PR #783, opened 2017-01-05
JBPM-4646 - Case management - filter by task name

mswiderski:


tsurdilo:
jenkins retest this

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
```diff
@@ -171,11 +173,26 @@ public static QueryFilter buildQueryFilter(Integer page, Integer pageSize) {
     }
 
     public static QueryFilter buildQueryFilter(Integer page, Integer pageSize, String orderBy, boolean asc) {
+        return buildQueryFilter(page, pageSize, orderBy, asc, null);
+    }
+
+    public static QueryFilter buildQueryFilter(Integer page, Integer pageSize, String orderBy, boolean asc, String filter) {
```
sutaakar: I think that "general" query builder shouldn't contain specific implementation like this task name filtering. I think that this method should be renamed to something like buildTaskQueryFilter, retrieve query filter from general method and apply task specific stuff here.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
```diff
@@ -870,6 +870,37 @@ public TaskInstance findTaskById(Long taskId) {
     }
 
     @Override
+    public List<TaskSummary> findTasksAssignedAsPotentialOwner(String userId, String filter, List<String> status, Integer page, Integer pageSize) {
```
sutaakar: This is unnecessary duplication, it would be better to call the other findTasksAssignedAsPotentialOwner with default sorting.

mswiderski: it can't. Be reused as it includes additional parameter, so the choice is between duplicate a bit or including additional if statements to check if filter is given or not. This is exspecially important for jms part that relies on number of arguments to select proper method server side

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
mswiderski: Valid point, will rename the method

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
sutaakar: I meant to reuse findTasksAssignedAsPotentialOwner(String userId, String filter, List<String> status, Integer page, Integer pageSize, String sort, boolean sortOrder) - the other method added by this PR . In this method there is sent same JMS command, just with default sorting values.

mswiderski: Ah, misunderstood firstly. Then it's a volid point too, thanks

mswiderski:
@sutaakar changes applied

mswiderski:
any other comments?

sutaakar:
No, looks good to me

PR #784, opened 2017-01-06
JBPM-5504: Missing getReleaseId method in KieServicesClient

livthomas:
I have added missing `getReleaseId` method to `KieServicesClient`.


sutaakar:
Looks good, +1

PR #787, opened 2017-01-09
KIE Server bugfixes : DROOLS-1350, DROOLS-1381, JBPM-5466, RHBPMS-4503, RHBPMS-4400

mswiderski:

PR #788, opened 2017-01-11
JBPM-5534: catch and handle IllegalArgumentException when aborting pr…

sutaakar:
…ocess


mswiderski:
@sutaakar this has been just fixed in jbpm see here https://github.com/droolsjbpm/jbpm/pull/722/files#diff-dd00f53a93585355731113247f2cf9adR153 though I moved your test class to #787 PR so it get merged when the jbpm pr is merged. So I believe we can close this one, don't you think?

sutaakar:
Right, thanks.

PR #789, opened 2017-01-11
GUVNOR-2845: Data modeller: Add new Java 8 Local Date/Time types

mcimbora:


mcimbora:
Verify java.time.LocalDate, java.time.LocalDateTime, java.time.LocalTime types can be marshalled/unmarshalled correctly.

Part of:
#198
droolsjbpm/droolsjbpm-integration#789
droolsjbpm/kie-wb-common#631

@ge0ffrey Can you please have a look?

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/JAXBMarshallerTest.java
```diff
@@ -0,0 +1,62 @@
...
+                "    <localDateTime>2017-01-01T10:10:10</localDateTime>\n" +
+                "    <localTime>10:10:10</localTime>\n" +
+                "</date-object>\n";
+
+        Set<Class<?>> extraClasses = new HashSet<Class<?>>();
+        extraClasses.add(DateObject.class);
+        Marshaller marshaller = MarshallerFactory.getMarshaller( extraClasses, MarshallingFormat.JAXB, getClass().getClassLoader() );
+
+        DateObject dateObject = new DateObject();
+        dateObject.setLocalDate( LocalDate.parse( "2017-01-01" ) );
```
ge0ffrey: Instead of parse(...) use of(...):
  LocalDate.of(2017, 1, 1)

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/JAXBMarshallerTest.java
```diff
@@ -0,0 +1,62 @@
...
+                "</date-object>\n";
+
+        Set<Class<?>> extraClasses = new HashSet<Class<?>>();
+        extraClasses.add(DateObject.class);
+        Marshaller marshaller = MarshallerFactory.getMarshaller( extraClasses, MarshallingFormat.JAXB, getClass().getClassLoader() );
+
+        DateObject dateObject = marshaller.unmarshall( expectedString, DateObject.class );
+        assertNotNull( dateObject );
+
+        assertEquals( "2017-01-01", dateObject.getLocalDate().toString() );
```
ge0ffrey: I might avoid the toString() (not that it's localized so it's not platform dependent iirc) althoug it's just a matter of taste:
assertEquals(LocalDate.of(2017, 1, 1), dateObject.getLocalDate());

ge0ffrey:
Can we also have a json (so Jackson) test?

mswiderski:
not sure I understand how this is solving the issue as it "hides" the java.time classes in another type so you actually checking if you can marshall and unmarshall properly DateObject. Would it sill work if you use the LocalDate directly?

```
String dateObjectString = marshaller.marshall( LocalDate.of(2017, 1, 1) );
```

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/objects/DateObject.java
```diff
@@ -0,0 +1,62 @@
...
+import java.time.LocalTime;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import com.thoughtworks.xstream.annotations.XStreamConverter;
+import org.kie.internal.jaxb.LocalDateTimeXmlAdapter;
```
ge0ffrey: This is bad (I hadn't noticed that before).
We shouldn't be writing our own JAXB adapters and XStream converters for LocalDate and other java.time.*  classes.
Those already exist, hopefully written and maintained by the JAXB and XStream guys themselves. But they might be in a extension jar (= separate dependency) because of JDK 6/7 compatibility issues.

mcimbora: Had they existed, I wouldn't have to come up with the PR :-) I did some research, however couldn't find any.

mcimbora:
@mswiderski @ge0ffrey The point is to enforce the usage of the XStream Converter and JAXB XmlAdapter (see https://github.com/droolsjbpm/droolsjbpm-integration/pull/789/files#diff-fe0c55124474c5b6691c8fd4f79b4fa9R26) I introduced in the other PR. This is how the real use case will look like anyway from OptaPlanner perspective. We have a domain object (e.g. DateObject) that may include date attributes.

As to your question,

`String dateObjectString = marshaller.marshall( LocalDate.of(2017, 1, 1) );`

works for XStream, however it produces the following output:

```
<java.time.LocalDate resolves-to="java.time.Ser">
  <byte>3</byte>
  <int>2017</int>
  <byte>1</byte>
  <byte>1</byte>
</java.time.LocalDate>
```

I'd rather give the users option to construct their requests in a more user friendly way. As to JAXB this doesn't even work as the unmarshaller doesn't know how to construct such instance.

mswiderski:
@mcimbora you have most likely did that but just to double check - you have added java.time.LocalDate to extra classes so JAXB context is aware of it? The same would apply for JSON. 

Though since XStream didn't work that is an issue - usually it's XStream that does work :)

mcimbora:
@mswiderski Correct, to be concrete, the marshaller complains about missing `@XmlRootElement` annotation.

`[com.sun.istack.SAXException2: unable to marshal type "java.time.LocalDate" as an element because it is missing an @XmlRootElement annotation]`

ge0ffrey:
Ideally, the LocalDate output follows the XSD date formats (which are based on the ISO formats IIRC).
http://www.w3schools.com/xml/schema_dtypes_date.asp
So not the ugly no-converter XStream format of above (which is btw not guaranteed to be backwards compatible in java 9 because those are private fields being reflected over).

LocaDate's. Java 8's DateTimeFormatter has predefined configs for all ISO formats, so reading(parsing)/writing(formatting) those exactly correct shouldn't be an issue.

mcimbora:
@ge0ffrey PR updated (JSON test + OffsetDateTime field (un)marshalling).

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/objects/DateObject.java
```diff
@@ -0,0 +1,77 @@
...
+import org.kie.internal.xstream.LocalTimeXStreamConverter;
+import org.kie.internal.xstream.OffsetDateTimeXStreamConverter;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlRootElement(name = "date-object")
+@XStreamAlias( "date-object" )
+public class DateObject {
+
+    @XmlJavaTypeAdapter(LocalDateXmlAdapter.class)
+    @XStreamConverter(LocalDateXStreamConverter.class)
```
ge0ffrey: Please add a TODO comment with the url to the xstream PR or issue that will replace our version of the LocalDateXStreamConverter

PR #790, opened 2017-01-12
Exclude geronimo-jta_1.1_spec artifact from Kie server

sutaakar:
This artifact is brought by Hibernate 5. Cause issues with Tomcat.
Kie server should use jboss-transaction-api artifact.
@psiroky FYI


psiroky:
Thanks @sutaakar. +1, we should also remove the outdated exclusion of `jboss-transaction-api_1.1_spec`, but that can be done separately.

PR #791, opened 2017-01-12
[Do not merge] JBPM-5551 - reproducer

sutaakar:
Just run TimerIntegrationTest against WildFly10 and check console output.


mswiderski:
replaced with https://github.com/droolsjbpm/droolsjbpm-integration/pull/809

PR #792, opened 2017-01-16
PLANNER-598 Reuse registerScoreConverters() (also adds missing score …

ge0ffrey:
…converter of hardmediumsoftbigdecimal)

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -49,6 +49,7 @@ import org.kie.server.api.model.ReleaseIdFilter;
 import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.api.model.ServiceResponsesList;
 import org.kie.server.api.model.instance.SolverInstance;
+import org.optaplanner.persistence.xstream.api.score.AbstractScoreXStreamConverter;
 import org.optaplanner.persistence.xstream.api.score.buildin.bendable.BendableScoreXStreamConverter;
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -49,6 +49,7 @@
 import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.api.model.ServiceResponsesList;
 import org.kie.server.api.model.instance.SolverInstance;
+import org.optaplanner.persistence.xstream.api.score.AbstractScoreXStreamConverter;
 import org.optaplanner.persistence.xstream.api.score.buildin.bendable.BendableScoreXStreamConverter;
```
mswiderski: a minor thing, aren't these unused imports since they were removed...

mswiderski:
@ge0ffrey is this waiting for someone else review or should it be merged?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
ge0ffrey: yes, they are - but we do "optimize imports" just before releasing CR1 normally (not all module owners currently do this, but I think we should).
Unlike other style changes, I tend to completely ignore unused imports as they are autofixed later anyway, as I think they're a waste of time to fix manually.

Just before CR1, I also tend to check findbugs, intellij inspect, etc - always worth it.

ge0ffrey:
Ready to merge

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
mswiderski: fair enough, thanks

PR #799, opened 2017-01-25
[6.5.x] RHBPMS-4445: add reproducer of found regression

livthomas:
I have added a reproducer of regression found after the fix for [RHBPMS-4445](https://issues.jboss.org/browse/RHBPMS-4445) was merged.

The problem is that the filtering of results does not work correctly now. Yes, it does not return older variable values (unless `all` parameter is given) but it returns processes in which the queried variable value is not the latest value (just a history record).


tsurdilo:
the test i being added in https://github.com/droolsjbpm/droolsjbpm-integration/pull/801 so this one can be closed

PR #801, opened 2017-01-25
RHBPMS-4445 - fix for regression + added regression test

tsurdilo:


mswiderski:
this looks good to me

@livthomas would you mind taking a look/try some tests before we merge this?

livthomas:
I have run REST Query API tests from QE's test suite with Business Central containing this fix and there is still one problem. Although it now works for the long parameter (`varvalue`), it fails for the short one (`vv`). See [REST Query API documentation](https://github.com/mrietveld/droolsjbpm-integration/wiki/jBPM-REST-Query-API) for more details.

livthomas:
I can see that the performance fix is not applied when using the variable value parameter. Just out of curiosity, would not it be possible to filter it later based on this parameter? I mean somewhere in `filterLastVarInstanceLogs` method.

tsurdilo:
@livthomas @mswiderski - thanks guys! amended.

livthomas:
REST Query API tests from our test suite pass now.

PR #809, opened 2017-02-02
JBPM-5551 - Timer start event triggered every second cycle

mswiderski:
depends on https://github.com/droolsjbpm/jbpm/pull/739

@sutaakar do you mind having a look?


mswiderski:
jenkins retest this

sutaakar:
@mswiderski Looks good. Just one idea, shouldn't it be better to place such test to Engine in container test suite - https://github.com/droolsjbpm/jbpm/tree/master/jbpm-container-test/jbpm-in-container-test - as this is engine issue?
What do you think?

mswiderski:
main reason I put it here is that we test it on all containers (wildfly, eap, was, wls) and not sure we do that for jbpm-in-container-test

sutaakar:
AFAIK jbpm-in-container-tests should be executed on all containers too. @MarianMacik Is it so?

mswiderski:
if that's the case then it's surely makes sense to have it there

mswiderski:
@sutaakar what do you think at merging this and later on add tests to jbpm if we confirm it does what we need?

sutaakar:
I agree.

PR #811, opened 2017-02-06
exclude jax-rs-api from ee6 ad ee7 assemblies of kie server

mswiderski:
@sutaakar here it is as we discussed...


sutaakar:
+1

sutaakar:
Jenkins retest this

PR #814, opened 2017-02-08
[DROOLS-936] clean-up dependencies to fix duplicated classes issue

psiroky:
brms-distribution and bpms-distribution cleaned-up in the process
as well. They did contain some jars which really should not be
part of the engine distributiosn (kie-maven-plugin, kie-ci-osgi).


psiroky:
Needs to be merged together with https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/383

mswiderski:
looks good to me +1

PR #816, opened 2017-02-09
[DROOLS-1426] remove CDI extension from drools-compiler

mariofusco:


mswiderski:
I think we should add the new module to https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/droolsjbpm-brms-distribution/src/main/assembly/pre-bin.xml and https://github.com/droolsjbpm/droolsjbpm-integration/blob/master/droolsjbpm-bpms-distribution/src/main/assembly/pre-bin.xml

mswiderski:
jenkins retest this

mariofusco:
ok to test

PR #817, opened 2017-02-09
JBPM-4646 - Case management - added support for taking completed node…

mswiderski:
…s for case instance, improved filtering for case nodes to exclude less meaningful nodes

depends on https://github.com/droolsjbpm/jbpm/pull/745


mswiderski:
@nmirasch here is comes :)

PR #818, opened 2017-02-10
[6.4.x] Workaround for dependency issues in OptaPlanner kie-server test

psiroky:
Importing the kie-platform-bom somehow confuses the KIE Maven builder
as it is not able to find the droolsjbpm-integration.pom during the
test execution. Since this branch (6.4.x) is no longer being supported,
it does not make sense to invest more time into proper solution.

@ksuta could you please take a look? I know this is not ideal solution, but its IMO good enough considering this branch is no longer actively maintained (+ the tests are working on 6.5.x and master).


sutaakar:
+1

PR #819, opened 2017-02-14
JBPM-5698 - Starting process with new Kie server node

mswiderski:
@sutaakar do you mind to take a quick look?


sutaakar:
Looks good, +1

PR #824, opened 2017-02-17
[reproducer] JBPM-5704: reproducer

sutaakar:
I was able to reproduce it just with this custom process, process already used in tests passed with expected error.


mswiderski:
replaced by https://github.com/droolsjbpm/droolsjbpm-integration/pull/825

PR #830, opened 2017-02-22
JBPM-5673 - Adding containers for Kie server router server template, …

mswiderski:
…JBPM-5670 - Kie server synchronization with router

@sutaakar could you have a look?


sutaakar:
Looks great, +1

PR #831, opened 2017-02-23
kie-server-tests: Case Id generator test

sutaakar:


mswiderski:
@sutaakar do we need to copy all the car insurance case stuff into the new module? Maybe we could just use simplest possible case definition instead as we don't really need anything advanced there to verify case id generators. Wdyt?

sutaakar:
Right, it could be better, will change it.

sutaakar:
Updated

mswiderski:
thanks @sutaakar

PR #835, opened 2017-03-02
[DROOLS-1452] fixes after Maven 3.2.5 -> 3.3.9 upgrade

psiroky:
Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/401

### DIFF
--- a/kie-eap-integration/kie-eap-distribution/pom.xml
+++ b/kie-eap-integration/kie-eap-distribution/pom.xml
@@ -316,6 +316,10 @@
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
+      <artifactId>maven-builder-support</artifactId>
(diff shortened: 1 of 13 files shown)


psiroky:
@mswiderski or @sutaakar could you please quickly review the kie-server changes?

mswiderski:
looks ok to me, though would be worth trying this on WAS as I remember there were few issues with guide in there but that might have been related to workbench only.

psiroky:
@mswiderski I am afraid I won't be able to try this on WAS.

mswiderski:
@psiroky no worries, I just thought about this and soon we need to try this (kie server) out on all containers anyway so we fix anything related at that time

mswiderski:
+1

PR #842, opened 2017-03-08
kie-server-tests: Specify directory for test remote repo

sutaakar:
Currently Maven CLI downloads dependencies for built test kjars into default repo (i.e. /home/"user"/.m2/repository). That happens even when settings.xml provided for kjar builds contains defined localRepository element - not sure why is it so.

This behaviour cause troubles for example when we want to use mirror server in settings.xml. As default repository can contain artifacts from different Nexus servers it can lead to classloader issues like:

"ERROR org.apache.maven.cli.MavenCli - Failed to execute goal org.kie:kie-maven-plugin:7.0.0-SNAPSHOT:build (default-build) on project stateless-session-kjar: Execution default-build of goal org.kie:kie-maven-plugin:7.0.0-SNAPSHOT:build failed: A required class was missing while executing org.kie:kie-maven-plugin:7.0.0-SNAPSHOT:build: org/drools/compiler/kproject/xml/PomModelGenerator"

@mswiderski @psiroky What do you think about this?

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/config/TestConfig.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/config/TestConfig.java
@@ -67,6 +67,7 @@ public class TestConfig {
 
     private static final StringTestParameter KJARS_BUILD_SETTINGS_XML = new StringTestParameter("kie.server.testing.kjars.build.settings.xml");
     private static final StringTestParameter KIE_CLIENT_DEPLOYMENT_SETTINGS = new StringTestParameter("kie.server.client.deployment.settings.xml");
+    private static final StringTestParameter REMOTE_REPO_DIR = new StringTestParameter("kie.server.testing.remote.repo.dir");
 
     private static final StringTestParameter CONTAINER_ID = new StringTestParameter("cargo.container.id");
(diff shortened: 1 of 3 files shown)


psiroky:
I am slowly getting lost in all these maven related configs (and it's largely my fault and I introduced lots of these!).

Not directly related to this PR, but could we start relying on the default maven repo (~/.m2/repository)? That way we could hopefully remove some of the settings.xml files. This is no longer an issue in CI as all the builds are starting from a clean state. No stale artifacts. It could still be an issue locally sometimes, but that's we we have to live with - people would have to clean their repos in those (hopefully) rare cases. WDYT?

sutaakar:
@psiroky I thought that these separated repos/settings were introduced to simulate real conditions as much as possible. That it serves as verification that Kie server can take care of download of kjar and its dependencies.

psiroky:
@sutaakar yeah, that definitely was one of the ideas (together with making sure we have stable CI builds). It just seems it got bit our of hand :) That being said, I am fine with adding these changes. Just throwing around things to possibly think about...

rsynek:
@psiroky I agree that running tests got complicated, especially community tests from QE environment. As using the default .m2 repository could at least partially simplify it, I am afraid we are not ready yet to rely on this option - e.g. windows slaves are not refreshed from image before every test job.

sutaakar:
jenkins retest this

sutaakar:
It seems to have some performance impact on tests. I guess we should keep current solution for now.

PR #843, opened 2017-03-08
JBPM-5750: Cannot disable registration of process event listeners

livthomas:
I think the listeners registration should be configurable only globally and so they should be retrieved directly from the system properties. Correct me if I am wrong.


mswiderski:
@livthomas initial idea was that it would be good to have it per container thus it was using the configItems from KieContainerInstance. Although as seen now it was over optimised which actually made it not work at all. 
Ideally I'd say we should have:
- improve the config items handling in kie server to deal with hierarchy - container -> context > system properties
- extend the UI of server management -> Process configuration to allow to mark if the listener should be enabled or not - at least the clean up and BAM

Though depending on urgency of this to be fixed we might decide now to marge this PR and try to fix this/improve later on. Wdyt?

livthomas:
@mswiderski I like your idea of creating some configuration hierarchy. Right now, different sets of properties in `KieServerConstants` are treated very differently and so it is hard to see how they can be used.

So +1 on merging this now so we will be able to configure listeners and can start working on a more complex solution.

mswiderski:
agreed, thanks @livthomas

PR #847, opened 2017-03-09
RHBPMS-4636 - MigrationManager API throws NPE when used in Kie-Spring  environment

tsurdilo:
@mswiderski thanks for all your help on this!

### DIFF
--- a/kie-spring/src/main/java/org/kie/spring/persistence/KieSpringTransactionManagerFactory.java
+++ b/kie-spring/src/main/java/org/kie/spring/persistence/KieSpringTransactionManagerFactory.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


tsurdilo:
jenkins retest this

kie-spring/pom.xml
```diff
@@ -258,6 +258,7 @@
           <systemPropertyVariables>
             <kie.maven.settings.custom>${project.build.testOutputDirectory}/kie-custom-settings.xml</kie.maven.settings.custom>
           </systemPropertyVariables>
+          <forkMode>always</forkMode>
```
mswiderski: @psiroky would it be possible to have a fork for just single test? I am bit afraid this might be too heavy...

psiroky: I don't think that's possible using just surefire config.

I agree that `<forkMode>always</forkMode>` is quite heavy. It usually suggests that the tests are not properly isolated or that they are not cleaning-up after themselves.

mswiderski: the reason why we have it here is that there are changes in transactionmanagerfactory - change of the impl. So we might be better of to add logic to reset the instance in that https://github.com/droolsjbpm/drools/blob/master/drools-persistence-jpa/src/main/java/org/drools/persistence/TransactionManagerFactory.java

tsurdilo:
@mswiderski amended per your suggestion - thanks. Depends on https://github.com/droolsjbpm/drools/pull/1139

will ask to retests once that pr is merged.

mswiderski:
Nice work @tsurdilo 

I'll leave it open for other to take a look and then merge on Monday

mswiderski:
jenkins retest this

PR #848, opened 2017-03-10
[DROOLS-1473] camel component

mariofusco:


kie-camel/src/test/java/org/kie/camel/KieComponentIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+                                                      "  <kie-server-info>\n" +
+                                                      "    <version>1.2.3</version>\n" +
+                                                      "  </kie-server-info>\n" +
+                                                      "</response>")));
+
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:start")
+                        .to("kie:" + mockServerBaseUri + "?username=admin&password=admin")
```
etirelli: @mariofusco probably need to discuss with the camel guys what to do in cases where authentication is necessary.

kie-camel/src/main/java/org/kie/camel/KieProducer.java
```diff
@@ -0,0 +1,229 @@
...
+        super(endpoint);
+        client = KieServicesFactory.newKieServicesClient( endpoint.getConfiguration() );
+    }
+
+    @Override
+    public void process( Exchange exchange ) throws Exception {
+        writeResponse( exchange, getProducer(exchange).execute(exchange) );
+    }
+
+    private void writeResponse( Exchange exchange, Object response ) {
```
mswiderski: wouldn't it be better to move that method (or it's responsibility) into producer? then we would not have to deal with service response if/else and the process method above would simply call:

```
getProducer(exchange).execute(exchange)
```

kie-camel/src/test/java/org/kie/camel/KieComponentIntegrationTest.java
mariofusco: @etirelli I discussed this with Nicola and he suggested to add the authentication mechanism that you can now see in the KieConfiguration object.

kie-camel/src/main/java/org/kie/camel/KieProducer.java
mariofusco: @mswiderski Thanks for the suggestion. Done.

tarilabs:
```
09:19:15 [INFO] --- maven-enforcer-plugin:1.4.1:enforce (ban-blacklisted-dependencies) @ kie-camel ---
09:19:15 [WARNING] Rule 0: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
09:19:15 Found Banned Dependency: commons-logging:commons-logging:jar:1.1.1
09:19:15 Use 'mvn dependency:tree' to locate the source of the banned dependencies.
```

triceo:
@jiripetrlik FYI.

mariofusco:
@mswiderski I asked @nicolaferraro to also review this PR and I addressed all his remarks except one with my last commit. What is not clear with the automatic reflective approach is which argument should be taken from the header and which one from the body. At the moment I'm only considering the header but this is clearly wrong. I'm seeing that on kie-server you're using the jrapidoc-rest-plugin to automatically generate documentation for rest's calls, so I was wondering:

1. Does that plugin also know what would be the content of the http body request? If so how?
2. Do you think I could use a similar mechanism even in the camel component to infer which argument should be taken from the exchange body?

mariofusco:
ok to test

mswiderski:
@mariofusco I am not camel expert but how I see it in general is that headers are like kind of control parameters while the body is data. But that might not be applicable for camel use cases.

I don't think that jrapidoc-rest-plugin would know it. Main issue is that we explicitly have on REST api string that represents the payload. This is to move the responsibility of the marshalling to our code base as data sent might be unknown to REST provider - classes from kjar etc. So the plugin will only know there is payload expected by not know what exactly it is. That's a drawback of moving the marshaller responsibility to our resource classes.

mariofusco:
ok to test

PR #854, opened 2017-03-13
JBPM-5775: KieServerRegistry now allows registration of custom classes for serve…

DuncanDoyle:
…r-wide marshallers.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/marshal/MarshallerHelper.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/marshal/MarshallerHelper.java
@@ -74,7 +74,11 @@ public class MarshallerHelper {
 
         Marshaller marshaller = serverMarshallers.get(format);
         if (marshaller == null) {
-        	marshaller = MarshallerFactory.getMarshaller(registry.getExtraClasses(), format, this.getClass().getClassLoader());
+        	if (registry != null) {
+        		marshaller = MarshallerFactory.getMarshaller(registry.getExtraClasses(), format, this.getClass().getClassLoader());
+        	} else {
+        		marshaller = MarshallerFactory.getMarshaller(format, this.getClass().getClassLoader());
+        	}
             serverMarshallers.put(format, marshaller);
         }
 
@@ -119,7 +123,11 @@ public class MarshallerHelper {
 
         Marshaller marshaller = serverMarshallers.get(format);
         if (marshaller == null) {
-            marshaller = MarshallerFactory.getMarshaller(format, this.getClass().getClassLoader());
+        	if (registry != null) {
+        		marshaller = MarshallerFactory.getMarshaller(registry.getExtraClasses(), format, this.getClass().getClassLoader());
+        	} else {
+        		marshaller = MarshallerFactory.getMarshaller(format, this.getClass().getClassLoader());
+        	}
             serverMarshallers.put(format, marshaller);
         }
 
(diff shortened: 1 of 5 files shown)


mswiderski:
looks like the test failures are related, do you mind double check @DuncanDoyle?

kie-server-parent/kie-server-services/kie-server-services-common/src/test/java/org/kie/server/services/impl/marshal/MarshallerHelperTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+		QueryFilterSpec expectedQueryFilterSpec = new QueryFilterSpecBuilder().get();
+
+		String marshalledQFS = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>" + "<query-filter-spec>"
+				+ "<order-asc>false</order-asc>" + "</query-filter-spec>";
+
+		QueryFilterSpec unmarshalledQFS = helper.unmarshal(marshalledQFS, MarshallingFormat.JAXB.toString(), QueryFilterSpec.class);
+
+		// QueryFilterSpec does not implement equals method .....
+		// assertEquals(expectedQueryFilterSpec, unmarshalledQFS);
```
mswiderski: commented out assert? is this on purpose... if so it does not test match then ;)

kie-server-parent/kie-server-services/kie-server-services-common/src/test/java/org/kie/server/services/impl/marshal/MarshallerHelperTest.java
```diff
@@ -0,0 +1,183 @@
...
```
mswiderski: same here

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/marshal/MarshallerHelper.java
```diff
@@ -120,7 +123,11 @@ public String marshal(String marshallingFormat, Object entity) {
 
         Marshaller marshaller = serverMarshallers.get(format);
         if (marshaller == null) {
-            marshaller = MarshallerFactory.getMarshaller(format, this.getClass().getClassLoader());
+        	if (registry != null) {
+        		marshaller = MarshallerFactory.getMarshaller(registry.getExtraClasses(), format, this.getClass().getClassLoader());
+        	} else {
+        		marshaller = MarshallerFactory.getMarshaller(format, this.getClass().getClassLoader());
+        	}
```
mswiderski: maybe this if/else to be replaced with:
```
marshaller = MarshallerFactory.getMarshaller(getExtraClasses(registry), format, this.getClass().getClassLoader());
```

and the getExtraClasses(registry) can do the check and if no registry return empty set. wdyt?

DuncanDoyle: Yes, that should be a bit cleaner.

kie-server-parent/kie-server-services/kie-server-services-common/src/test/java/org/kie/server/services/impl/marshal/MarshallerHelperTest.java
DuncanDoyle: The problem is that that QueryFilterSpec does not implement an "equals" method. So either we:
- implement an "equals" method on that class.
- implement the equals manually
- remove the assert

Atm, I don't really have a preference. The thing is that we're testing MarshallerHelper, not Marshaller. The check whether the marshalled and unmarshalled content is correct should actually be in a Marshaller test.

Anyway, let me know what you prefer and I'll change the impl.

mswiderski:
jenkins retest this

DuncanDoyle:
The problem was that I returned an Empty Set when there was no registry configured on the MarshallerHelper. This works fine with the JAXB Marshaller. But the JSON Marshaller adds additional classes by default. When it tried to do that on an empty-set, an UnsupportedOperationException was thrown. So, we return an empty HashSet now. Also implemented unit-tests that tests the behaviour.

mswiderski:
jenkins retest this

PR #855, opened 2017-03-14
JBPM-5717 - Support for multiple kie-servers hosting different kjars …

mswiderski:
…sharing the same database

depends on https://github.com/kiegroup/jbpm/pull/780

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
@@ -62,6 +62,9 @@ public class JobServicesClientImpl extends AbstractKieServicesClientImpl impleme
                     Object.class);
 
         } else {
+            if (containerId == null) {
+                containerId = "";
+            }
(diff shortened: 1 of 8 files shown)


kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
```diff
@@ -179,6 +209,38 @@ public void requeueRequest(long requestId) {
     }
 
     @Override
+    public List<RequestInfoInstance> getRequestsByBusinessKey(String businessKey, List<String> statuses, Integer page, Integer pageSize) {
+        RequestInfoInstanceList list = null;
+        if( config.isRest() ) {
```
rsynek: Whenever I see this construction in Kie-Server Client, I wonder if it would be possible to refactor the code to have two client implementations and a factory deciding which one of them to provide based on `config.isRest()`. I am not saying it should be done now for this PR, just making a suggestion we can discuss further.

mswiderski: makes sense if we can find a nice way of doing it I am all in

PR #856, opened 2017-03-15
Fix JSONMarshaller.writeMap() check for double-wrapping for map values

tarilabs:
I believe this is a copy/paste typing mistake for [the analogous code at lines 521-529](https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java#L521-L529) preceding this proposed correction.


tarilabs:
please await before merging.
I think I found it did not entirely solved my issue

mswiderski:
is this still valid?

tarilabs:
I believe this correction is needed, however in my test scenario generated more troubles than benefits. So I'm not sure yet if we are missing some analogous correction on the unmarshalling side, or something else. We could proceed as a blind flix, or close this PR and raise a low priority JIRA to investigate this further, separately, later. What do you think?

mswiderski:
it's fine with me to include it as I think it is a fix anyway, so +1 from me

tarilabs:
PR renamed accordingly

mswiderski:
closing as it needs more investigation before merging, jira issue will follow

PR #859, opened 2017-03-19
Update KSessionFactoryBean.java

zhugw:
When concurrently get prototype kSession, I found sometimes there are some kSession instances are repeated. So it is not strictly prototype.  If let kSession as a local variable, I think could solve this problem.

### DIFF
--- a/kie-spring/src/main/java/org/kie/spring/factorybeans/KSessionFactoryBean.java
+++ b/kie-spring/src/main/java/org/kie/spring/factorybeans/KSessionFactoryBean.java
@@ -171,7 +171,7 @@ public class KSessionFactoryBean
     public Object getObject() throws Exception {
         if ("prototype".equalsIgnoreCase(scope)) {
             helper.setKieBase(kBase);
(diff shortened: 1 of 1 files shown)


mswiderski:
@zhugw could you please add a test case for this?

zhugw:
@mswiderski please see https://github.com/zhugw/droolsprototypedemo

mswiderski:
thanks, but please add it to this PR so it can verify the actual fix as part of the build

zhugw:
Hi I added a test case : `org.kie.spring.tests.KieSpringScopeTest#testConcurrentlyGetStatefulPrototypeKieSession`, please see https://github.com/kiegroup/droolsjbpm-integration/pull/868

zhugw:
Does in production environment rare people use prototype session? so has not found this problem till now.

mswiderski:
closing as this is replaced by #868

mswiderski:
it looks like it's not so frequently used as prototype as you seem to be the first one hitting this issue, btw could you please open jira issue for this as well?

PR #860, opened 2017-03-20
DROOLS-1463 Align with Kie-DMN-API changes

tarilabs:


manstis:
@tarilabs This is to fix master compilation right?

tarilabs:
@manstis yes indeed

mswiderski:
@tarilabs thanks!

PR #863, opened 2017-03-21
[reproducer] Spring issue reproducer

sutaakar:
@mswiderski Failure occurs in testPersistenceTimer(). It should be reproducible in IDE.


mswiderski:
Is this suppose to fail/hang or it should be fine as you removed the test annotations?

sutaakar:
It should hang, I removed other test annotations so we can focus just on this test. It doesn't seem to be affected by other tests.

mswiderski:
Since you pin pointed the issue maybe ignore the test for now so we don't block PR builds. Wdyt?

sutaakar:
Yes, we can, will prepare PR.

sutaakar:
PR prepared: https://github.com/kiegroup/droolsjbpm-integration/pull/864

PR #865, opened 2017-03-21
[DROOLS-1473] allow to configure body parameters in camel component

mariofusco:


mariofusco:
@nicolaferraro Can you please give a look?

mariofusco:
ok to test

nicolaferraro:
@mariofusco  I'll do a check later this morning

nicolaferraro:
looks good (although I don't know why 1 test is failing in my workspace: `testBodyParam2`).
This way you can define which parameters will be taken from the body, and also let the user customize it, if needed.

Just a minor issue, use camel built-in conversion also for the body [here](https://github.com/kiegroup/droolsjbpm-integration/pull/865/files#diff-4fbed41f7b44d484c33b04e3ec8e24fbR184), so that the body is converted to the parameter type automatically, if needed.

About headers, I'd remove the "dot" notation to avoid problems with Camel DSL, and try to use camel case with the same prefix if possible (or a sub-prefix, like "CamelKieParamXxx", if there are conflicts).
Not "kie.containerId" but "CamelKie**C**ontainerId", or "CamelKieParamContainerId".

The same for kie client and kie operation: "kie-client" --> "CamelKieClient", "kie-operation" --> "CamelKieOperation".
For these two, a common pattern is (already told you in the other PR, sorry for the repeat..) to allow users to specify them in endpoint options:
```
...to("kie:http://srv?client=process&operation=signal")
```
In the code, this means looking at the headers or the endpoint, if the header is not present.
(there are also other good candidates for endpoint options).

mariofusco:
@tarilabs I pushed a 2nd commit addressing all @nicolaferraro hints. Please give one last look and let me know if it is ok to be merged.

kie-camel/src/test/java/org/kie/camel/KieComponentOpOnUriTest.java
```diff
@@ -0,0 +1,148 @@
...
+                                                      "           </entry>\n" +
+                                                      "       </decisionResults>\n" +
+                                                      "   </dmn-evaluation-result>\n" +
+                                                      "</response>")));
+
+        return new RouteBuilder() {
+            @Override
+            public void configure() {
+                from("direct:start")
+                        .to("kie:" + getAuthenticadUrl("admin", "admin") + "?client=dmn&operation=evaluateAllDecisions")
```
tarilabs: demonstrates as passed on URI of endpoint

kie-camel/src/main/java/org/kie/camel/KieCamelConstants.java
```diff
@@ -18,12 +18,12 @@
 
 public class KieCamelConstants {
 
-    public static final String KIE_HEADERS_PREFIX = "kie.";
+    public static final String KIE_HEADERS_PREFIX = "CamelKie";
```
tarilabs: All component are KieCamel, while this is `CamelKie` ?

tarilabs: Consistent with the other Camel components "CamelInfinispan..." or "CamelJdbc..."

baldimir:
jenkins retest this

PR #871, opened 2017-03-23
JBPM-5835: Test coverage for KeepLatestContainerOnlyPolicy

sutaakar:
It could be fixed by not disposing containers in CREATING state or disposing just containers with earlier version. @mswiderski What do you think?

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import org.kie.server.api.model.KieContainerStatus;
 import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.services.api.KieServer;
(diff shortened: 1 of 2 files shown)


sutaakar:
Also I have noticed that this policy implementation use its own controller instance: https://github.com/kiegroup/droolsjbpm-integration/blob/9cddb3873c926c4c9d46bfb885e287a438c26e64/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/policy/KeepLatestContainerOnlyPolicy.java#L90 . I think it would be better to use centralized solution, for example provided by KieServerImpl: https://github.com/kiegroup/droolsjbpm-integration/blob/f7dcde6ad77f79efcb286d5327f3e00b8a8cd0c4/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java#L808

mswiderski:
@sutaakar the idea for the fix is valid, do you mind adding it as a fix next to the test?

and using the same controller ... hmm what would be the gain here? Other than reusing that instance?

sutaakar:
Done. About that controller, the only problem what could happen is if customer will implement its own controller with different endpoints. For KieServerImpl there could be provided new controller using service loader, but KeepLatestContainerOnlyPolicy has hardcoded controller implementation.

sutaakar:
Jenkins retest this

sutaakar:
Jenkins retest this

PR #885, opened 2017-03-29
JBPM-5887: Add support for WebSphere 9

sutaakar:
Initial implementation, to be discussed.

Contains own WebSphere platform implementation as implementation provided by Hibernate doesn't work - throwing UnsupportedOperationException.

Transaction API exclusion - will check if this works with WildFly/WLS.
cxf-rt-ws-policy - was causing issues with WebSphere's own cxf.  It is brought by jbpm-workitems, I am not sure what functionality depends on it.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee7-container.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-test-war/src/main/assembly/assembly-ee7-container.xml
@@ -51,6 +51,7 @@
 
         <include>com.sun.xml.bind:jaxb-impl</include>
         <include>com.sun.xml.bind:jaxb-core</include>
+        <include>org.slf4j:slf4j-api</include>
       </includes>
       <outputDirectory>WEB-INF/lib</outputDirectory>
       <useTransitiveFiltering>true</useTransitiveFiltering>
--- a/kie-server-parent/kie-server-tests/pom.xml
+++ b/kie-server-parent/kie-server-tests/pom.xml
@@ -253,7 +253,7 @@
               </goals>
               <configuration>
                 <excludedGroups>${failsafe.excluded.groups}</excludedGroups>
-                <additionalClasspathElements>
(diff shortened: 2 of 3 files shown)


psiroky:
@sutaakar does it mean we will need to support both WAS 8 and WAS9?

sutaakar:
@psiroky We should support WAS 9 for version 7

psiroky:
@sutaakar so could we in that case remove the old WAS 8 config?

sutaakar:
From my point of view we can remove WAS 8

sutaakar:
Hibernate issue reported as https://hibernate.atlassian.net/browse/HHH-11606

triceo:
@psiroky WAS8 confirmed removable.

triceo:
@psiroky If you've got any support for WLS 12c R1 specifically, that can go too. 7 is exclusively WLS 12c R2.

psiroky:
@triceo I guess the kie-server-tests have a profile for WLS, but that can be reused for the newer version I think.

sutaakar:
Jenkins retest this

sutaakar:
WebSphere 9 seems to have a problem with cxf-rt-ws-policy artifact as it collides with internal cxf implementation. cxf-rt-ws-policy is brought by jbpm-workitems.
@mswiderski Do you know what is this artifact used for?

mswiderski:
@sutaakar it is brought by cxf-rt-frontend-jaxws dependency in jbpm-workitems but I don't think we use id directly. So we could exclude it since it causes an issue on WAS.

sutaakar:
@mswiderski In that case I think it is ready for review/merge. Btw should I create a JIRA for this WebSphere 8 -> 9 upgrade?

mswiderski:
the change looks good to me but in fact you're the expert here :)

what do you mean by WebSPhere upgrade jira? is there anything we need to do for that?

sutaakar:
I meant if there should be some JIRA to be paired with this PR

mswiderski:
sure, let's have a jira for this for the record, thanks @sutaakar

PR #888, opened 2017-03-31
KIE integration test coverage

winklerm:
This PR creates a new project _kie-integration-test-coverage_ - an analogy of _drools/drools-testcoverage_ and _jbpm/jbpm-testcoverage_ projects. The new project will contain additional QE tests for the code in _droolsjbpm-integration_ repository.

The project currently contains a draft of _kie-maven-plugin_ model classes instrumentation tests, more tests will follow.

### DIFF
--- a/kie-integration-test-coverage/kie-maven-plugin-tests/kjar-with-instrumentation-wrapper/pom.xml
+++ b/kie-integration-test-coverage/kie-maven-plugin-tests/kjar-with-instrumentation-wrapper/pom.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.drools.testcoverage</groupId>
+    <artifactId>kie-maven-plugin-tests-parent</artifactId>
+    <version>7.1.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>kjar-with-instrumentation-wrapper</artifactId>
+  <packaging>jar</packaging>
+
+  <name>KIE Integration :: KIE Maven Plugin Instrumented kJAR Wrapper</name>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-invoker-plugin</artifactId>
+        <configuration>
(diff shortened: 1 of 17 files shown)


kie-integration-test-coverage/kie-maven-plugin-tests/kjar-with-instrumentation/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+    <!-- required due to kie-maven-plugin instrumentation dependencies
+         on org.drools.core.phreak.ReactiveObject and others -->
+    <enforcer.skip>true</enforcer.skip>
+  </properties>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-maven-plugin</artifactId>
```
psiroky: I am afraid this is not going to work. The plugin itself is in the same Maven reactor build, which means it is being used before it's build (this is a limitation of Maven) and thus the parsing of the POMs will fail. This will only happen for the release versions as with SNAPSHOT Maven will just download the already available one from remote repo

kie-integration-test-coverage/kie-maven-plugin-tests/kjar-with-instrumentation/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+  <artifactId>kjar-with-instrumentation</artifactId>
+  <packaging>kjar</packaging>
+
+  <name>KIE Integration :: KIE Maven Plugin Instrumented kJAR</name>
+
+  <properties>
+    <!-- required due to kie-maven-plugin instrumentation dependencies
+         on org.drools.core.phreak.ReactiveObject and others -->
+    <enforcer.skip>true</enforcer.skip>
```
psiroky: I believe it would be better to skip just the specific enforcer rule which is failing.

winklerm: That's right - I was thinking about investigating it more and then maybe file a JIRA, but your suggestion totally makes sense.

kie-integration-test-coverage/kie-maven-plugin-tests/kjar-with-instrumentation/pom.xml
winklerm: Right, I have not realized that the whole droolsjbpm-integration project is the same reactor build. I will think about some other way, thanks for the suggestion!

winklerm:
Updated to _7.1.0-SNAPSHOT_ and rebased to the current _master_.

psiroky:
I had to temporarily revert the changes via https://github.com/kiegroup/droolsjbpm-integration/commit/722fc91645e5e9d30a575512429b3aaedc149772. There are intermittent build order issues we need to address and the add the changes back.

winklerm:
Added the artificial dependencies that should make the parallel build aware of the correct order of projects.

PR #892, opened 2017-04-05
RHBPMS-1328 - business rule task supports boundary error event - fix …

tsurdilo:
…for osgi


mswiderski:
looks good to me, @psiroky do you see any potential issues or we are good?

psiroky:
This should be good. Let's wait for the build and then we can merge.

PR #901, opened 2017-04-11
[GUVNOR-2991] use custom xstream-hibernate converters

psiroky:
This is needed to make sure that KIE Server is not sending away the
marshalled inernal hibernate types (like e.g. PersistentBag). Those
then can not be un-marshalled unless hibernate is on classpath.
This was an issue in workench where the hibernate was recently removed
from WEB-INF/lib.

Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/428

@mswiderski could you please take a look?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -15,10 +15,6 @@
 
 package org.kie.server.api.marshalling.xstream;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
-
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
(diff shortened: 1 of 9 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -103,7 +116,16 @@ public Class realClass(String elementName) {
...
             }
         });
+    }
 
+    private MapperWrapper chainMapperWrappers(List<XStreamMarshallerExtension> extensions, MapperWrapper last) {
+        if (extensions.isEmpty()) {
+            return last;
+        } else {
+            XStreamMarshallerExtension head = extensions.get(0);
+            List<XStreamMarshallerExtension> tail = extensions.subList(1, extensions.size());
```
mswiderski: wouldn't it be more readable to replace these two lines with extensions.remove(0) so it will directly have the list shorter of that one item? Assuming the extensions initially will be a copy of the original list.

psiroky: Yeah, it may be a bit easier to read. I was trying not to mutate the provided params as I consider that a bad practice. I can change this though.

psiroky: Replaced with extensions.remove(0);

mswiderski: looks like we have it now both get and remove in use. 
```
XStreamMarshallerExtension head = extensions.get(0);
 extensions.remove(0);
```

maybe replacing it with 
```
XStreamMarshallerExtension head = extensions.remove(0);
``` 
would be enough...

psiroky: Hopefully fixed now...

PR #905, opened 2017-04-12
Add artificial dep to fix parallel build

psiroky:
@winklerm could you please take a look?


psiroky:
Closing as this is not the fix we are looking for.

PR #909, opened 2017-04-13
RHBPMS-4687 - [GSS](6.4.x)Endpoint /history/variable/{varId}/value/{value} not work

jesuino:
…ing for variables with space


mswiderski:
shouldn't we have a test for this?

mswiderski:
@jesuino are you going to add a test case?

mswiderski:
Closing as it is included in https://github.com/kiegroup/droolsjbpm-integration/pull/991

PR #911, opened 2017-04-13
[7.0.x] JBPM-4646 - case management - case file data audit trail

mswiderski:
depends on https://github.com/kiegroup/jbpm/pull/812

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/cases/CaseFileDataItem.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/cases/CaseFileDataItem.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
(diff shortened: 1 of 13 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/cases/CaseFileDataItem.java
```diff
@@ -0,0 +1,143 @@
+/*
+ * Copyright 201y Red Hat, Inc. and/or its affiliates.
```
psiroky: This looks weird, maybe some mis-configuration in the IDE template?

mswiderski: not really, it's just simple typo

psiroky: ah, ok :) I thought the `y` represents the last digit of current year, but it somehow wasn't replaced by the actual value.

mswiderski: fixed

PR #914, opened 2017-04-19
kie-server-tests: Extend TimerIntegrationTest for runtime strategies

sutaakar:


mswiderski:
looks good, is this suppose to fail on some app servers (weblogic)?

sutaakar:
It should fail on WebLogic - it was failing even with original test. I am extending it to catch possible regressions in different runtime strategies in the future.

PR #930, opened 2017-04-25
Feature/kie jbpm task query service 2

DuncanDoyle:
Initial implementation of BPMSPL-605.


kie-server-parent/kie-server-jbpm-queries-api/src/main/java/org/kie/server/jbpm/queries/api/model/definition/TaskField.java
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+package org.kie.server.jbpm.queries.api.model.definition;
+
+/**
+ * HumanTask fields.
+ * <p/>
+ * These are the filterable fields in our TaskQuery API. 
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-jbpm-queries-api/src/main/java/org/kie/server/jbpm/queries/util/ProcessInstanceQueryFilterSpecBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+import org.kie.server.jbpm.queries.api.model.definition.TaskField;
+
+/**
+ * QueryFilterSpecBuilder targeted at filters for Process Instances.
+ * <p/>
+ * This provides a higher level-api than the Advanced Query API (i.e. {@link QueryFilterSpecBuilder} as the ProcessInstance field-names (column-names)
+ * are pre-defined and exposed via a strongly-typed API, and thus not bound to specific column-names in the database. This guards users of
+ * this API for potential changes in the jBPM schema. 
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-jbpm-queries-api/src/main/java/org/kie/server/jbpm/queries/util/TaskQueryFilterSpecBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+import org.kie.server.jbpm.queries.api.model.definition.TaskQueryFilterSpec;
+
+/**
+ * QueryFilterSpecBuilder targeted at filters for Tasks.
+ * <p/>
+ * This provides a higher level-api than the Advanced Query API (i.e. {@link QueryFilterSpecBuilder} as the Task field-names (column-names)
+ * are pre-defined and exposed via a strongly-typed API, and thus not bound to specific column-names in the database. This guards users of
+ * this API for potential changes in the jBPM schema. 
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-jbpm-queries-api/src/main/java/org/kie/server/jbpm/queries/util/TaskQueryFilterSpecBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+		return this;
+	}
+
+	public TaskQueryFilterSpecBuilder notEqualsTo(TaskField field, Comparable<?>... values) {
+		parameters.add(new QueryParam(field.toString(), "NOT_EQUALS_TO", Arrays.asList(values)));
+
+		return this;
+	}
+
+	@SuppressWarnings("unchecked")
```
domhanak: Any particular reason for these annotations in this class?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/ProcessInstanceQueryServicesClient.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+
+import org.kie.server.api.model.instance.ProcessInstance;
+import org.kie.server.jbpm.queries.api.model.definition.ProcessInstanceQueryFilterSpec;
+
+/**
+ * KIE-Server Client API for the advanced Process Instance Queries provided by the <code>kie-server-services-jbpm-queries</code> extension.
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/ProcessInstanceQueryServicesClientImpl.java
```diff
@@ -0,0 +1,60 @@
...
+		super(config, classLoader);
+	}
+
+	@Override
+	public List<ProcessInstance> findProcessInstancesWithFilters(ProcessInstanceQueryFilterSpec filterSpec, Integer page, Integer pageSize) {
+		ProcessInstanceList result = null;
+
+		if (config.isRest()) {
+			String queryString = getPagingQueryString("?", page, pageSize);
+			result = makeHttpPostRequestAndCreateCustomResponse(loadBalancer.getUrl() + "/" + PROCESS_INSTANCES_GET_FILTERED_URI + queryString,
```
domhanak: no check for response here, would be nice to add something to check if the result is OK

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/TaskQueryServicesClientImpl.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	@Override
+	public List<TaskInstance> findHumanTasksWithFilters(TaskQueryFilterSpec filterSpec, Integer page, Integer pageSize) {
+
+		TaskInstanceList result = null;
+
+		if (config.isRest()) {
+			String queryString = getPagingQueryString("?", page, pageSize);
+			result = makeHttpPostRequestAndCreateCustomResponse(loadBalancer.getUrl() + "/" + TASKS_GET_FILTERED_URI + queryString,
```
domhanak: no check for response here, would be nice to add something to check if the result is OK

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/ProcessInstanceQueryResource.java
```diff
@@ -0,0 +1,71 @@
...
+	public ProcessInstanceQueryResource(ProcessInstanceQueryServiceBase processInstanceQueryServiceBase, KieServerRegistry context) {
+		this.processInstanceQueryServiceBase = processInstanceQueryServiceBase;
+		this.context = context;
+	}
+	
+	
+	//TODO: We now only support standard QueryFilterSpec configurations ... Do we also need to support builders???
+	@POST
+	@Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+	public Response getProcessInstancesWithFilters(@Context HttpHeaders headers,
```
domhanak: @POST annotation but method name starts with GET and I don't see anything being modified here.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/TaskQueryResource.java
```diff
@@ -0,0 +1,70 @@
...
+	private KieServerRegistry context;
+	
+	public TaskQueryResource(TaskQueryServiceBase taskQueryServiceBase, KieServerRegistry context) {
+		this.taskQueryServiceBase = taskQueryServiceBase;
+		this.context = context;
+	}
+	
+	
+	//TODO: We now only support standard QueryFilterSpec configurations ... Do we also need to support builders???
+	@POST
```
domhanak: @POST annotation but method name starts with GET and I don't see anything being modified here.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
```diff
@@ -0,0 +1,80 @@
...
+		assertEquals(3, appComponents.size());
+	}
+	
+	
+	/**
+	 * Helper class to test retrieval of AppComponents. 
+	 * <p>
+	 * This class is loaded by the {@link ServiceLoader}.
+	 * 
+	 * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/TaskQueryServicesClient.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+
+import org.kie.server.api.model.instance.TaskInstance;
+import org.kie.server.jbpm.queries.api.model.definition.TaskQueryFilterSpec;
+
+/**
+ * KIE-Server Client API for the advanced Task Queries provided by the <code>kie-server-services-jbpm-queries</code> extension.
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/QueryCallback.java
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package org.kie.server.services.jbpm.queries;
+
+import org.kie.server.services.jbpm.queries.util.QueryStrategy;
+
+/**
+ * Provides the context for query execution, including things like {@link QueryStrategy}, query-name, mapper-name, etc.
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/QueryServiceTemplate.java
```diff
@@ -0,0 +1,134 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Template which provides functionality to do a query via {@link QueryService}.
+ * <p>
+ * Users of the template need to provide a {@link QueryCallback} and a {@link RequestCallback}, which provide the context in which to
+ * execute the query.
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/RequestCallback.java
```diff
@@ -0,0 +1,31 @@
...
+package org.kie.server.services.jbpm.queries;
+
+import org.kie.server.jbpm.queries.api.model.definition.BaseQueryFilterSpec;
+import org.kie.server.services.jbpm.queries.util.QueryStrategy;
+
+/**
+ * Provides the context for query request execution, including things like payload and unmarshalled {@link BaseQueryFilterSpec}. 
+ * 
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/util/QueryStrategyFactory.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.kie.server.api.KieServerConstants;
+import org.kie.server.services.api.KieServerRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Factory responsible for building the database-specific {@link QueryStrategy}. 
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/util/QueryStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+package org.kie.server.services.jbpm.queries.util;
+
+import java.util.Map;
+
+import org.jbpm.services.api.query.model.QueryParam;
+
+/**
+ * Provides DB specific query and column-mappings for queries.
+ * 
+ * @author <a href="mailto:duncan.doyle@redhat.com">Duncan Doyle</a>
```
domhanak: author tag

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
```diff
@@ -0,0 +1,80 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.services.jbpm.queries;
+
+import static org.junit.Assert.*;
```
domhanak: star import

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/util/QueryStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface QueryStrategy {
```
domhanak: What' the purpose of various QueryStrategies here? Why do we handle data/column types here, when it can be handled by JPA.

DuncanDoyle: It can't be handled by JPA because we don't use JPA. We use the jBPM AdvancedQueries API which does a direct SQL call into the DB. The reason we use the AdvancedQueries API instead of JPA is because it allows us to almost fully reuse the existing (remote) query interface and query services without needing to come up with all sorts of JPA-based queries. So main reason was that this is simply a small layer on top of Advanced Queries rather than something completely new.

The Advanced Queries API requires us to pre-register a query (which can be DB specific) and the mapping of columns to types (which can be DB specific), hence the decision to use a Strategy per DB type.

DuncanDoyle: Basically the API uses this concept: http://mswiderski.blogspot.nl/2016/01/advanced-queries-in-kie-server.html

This allowed us to reuse all of the existing API constructs to implement those filters.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/ProcessInstanceQueryResource.java
DuncanDoyle: Look closely at the method signature. It has a payload (which means a HTTP body). That body contains, among other things, the filters for our query. You can't send a body with a HTTP GET (well, technically you can, but it completely screws up the HTTP caching concept of GETs because those caches don't take the body of a GET into account). So, we need to do a POST to 
be able to pass a query-filter. Same interface as this one actually:  https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/QueryDataResource.java#L187

I could change the method name into something like "runProcessInstancesQueryWithFilters" if necessary.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/TaskQueryResource.java
DuncanDoyle: Same comment as above. I could change the method name in "runTaskInstancesQueryWithFilters" if necessary.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
DuncanDoyle: We can't do "*" imports on "Assert" in JUnit in a test? That's a first :-)

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
```diff
@@ -30,6 +30,7 @@
     public static final String KIE_OPTAPLANNER_SERVER_EXT_DISABLED = "org.optaplanner.server.ext.disabled";
     public static final String KIE_CASE_SERVER_EXT_DISABLED = "org.jbpm.case.server.ext.disabled";
     public static final String KIE_DMN_SERVER_EXT_DISABLED = "org.kie.dmn.server.ext.disabled";
+    public static final String KIE_JBM_QUERIES_SERVER_EXT_DISABLED = "org.jbpm.taskqueries.server.ext.disabled";
```
mswiderski: shouldn't the value of the property not include task any more as it's not only for tasks?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/ProcessInstanceQueryServicesClientImpl.java
mswiderski: check of the state of response is already inside the makeHttpPost... method if the response code is not successful it will throw exception

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/TaskQueryServicesClientImpl.java
mswiderski: same as above

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/JBPMQueryServicesClientBuilder.java
```diff
@@ -0,0 +1,45 @@
...
+	public String getImplementedCapability() {
+		return KieServerConstants.CAPABILITY_BPM_QUERIES;
+	}
+
+	@Override
+	public Map<Class<?>, Object> build(KieServicesConfiguration configuration, ClassLoader classLoader) {
+		Map<Class<?>, Object> services = new HashMap<Class<?>, Object>();
+
+		services.put(ProcessInstanceQueryServicesClient.class, new ProcessInstanceQueryServicesClientImpl(configuration, classLoader));
+        services.put(TaskQueryServicesClient.class, new TaskQueryServicesClientImpl(configuration, classLoader));
```
mswiderski: what's the point of splitting this into two client types where each will only have single method? Do we foresee lots on additional methods in these clients?

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/ProcessInstanceQueryResource.java
mswiderski: GET does not support payload and thus we can use GET here

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/TaskQueryResource.java
mswiderski: same as above

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/test/java/org/kie/server/remote/rest/jbpm/queries/TaskQueryResourceTest.java
```diff
@@ -0,0 +1,105 @@
+package org.kie.server.remote.rest.jbpm.queries;
```
mswiderski: missing license header

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+			<groupId>org.kie.server</groupId>
+			<artifactId>kie-server-services-common</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>org.kie.server</groupId>
+			<artifactId>kie-server-common</artifactId>
+		</dependency>
+    
+    <!-- TODO: This dependency is needed because we need access to ConvertUtils. Maybe it's an idea to put those utils in a common lib. -->
```
mswiderski: makes sense

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/AbstractQueryServiceBase.java
```diff
@@ -0,0 +1,78 @@
...
+				logger.debug("Converting collection of UserTaskInstanceDesc to TaskInstanceList");
+				actualResult = convertToTaskInstanceList((Collection<UserTaskInstanceDesc>) result);
+			} else if (TaskSummary.class.isAssignableFrom(resultMapper.getType())) {
+
+				logger.debug("Converting collection of TaskSummary to TaskSummaryList");
+				actualResult = convertToTaskSummaryList((Collection<TaskSummary>) result);
+			} else if (List.class.isAssignableFrom(resultMapper.getType())) {
+
+				logger.debug("Converting collection of List to ArrayList");
+				actualResult = new ArrayList((Collection) result);
```
mswiderski: just so we won't loose it, today I merged the error handling stuff which introduced one additional type to convert to

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtension.java
```diff
@@ -0,0 +1,147 @@
...
+    private boolean initialized = false;
+
+    @Override
+    public boolean isInitialized() {
+        return initialized;
+    }
+
+    @Override
+    public boolean isActive() {
+        return disabled == false;
```
mswiderski: since this requires jbpm extension would be worth checking if it was not disabled - similar how jbpm-ui does it

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/ProcessInstanceQueryServicesClientImpl.java
DuncanDoyle: I need to discuss with @mswiderski, because this pattern is used all over the KIE-Server client, e.g.: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/QueryServicesClientImpl.java

So, any changes in that pattern would need to be implemented in the whole client to keep the code consistent.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/RequestCallback.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public interface RequestCallback {
```
mswiderski: is this interface and then specific impl really needed? can't we simply add argument of type BaseQueryFilterSpec?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/TaskQueryServicesClientImpl.java
DuncanDoyle: Same reply as above, I need to discuss this with @mswiderski because this pattern is used everywhere in the client code.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.ServiceLoader;
+
+import org.jbpm.services.api.query.QueryService;
+import org.junit.Test;
+import org.kie.server.services.api.KieServerApplicationComponentsService;
+import org.kie.server.services.api.KieServerRegistry;
+import org.kie.server.services.api.SupportedTransports;
+import org.mockito.Mockito;
+
+public class JbpmQueriesKieServerExtensionTest {
```
mswiderski: do we see a big value in this test?

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/main/java/org/kie/server/services/jbpm/queries/util/QueryStrategy.java
mswiderski: I don't think we need these strategies, dasgbulder dataset should do that for us - at least this is what we rely on in workbench where we use same data sets across multiple dbs. Dashbuilder does retrieve that information based on db metadata so we should only provide type of fields that are known to dash builder 


or do I missing something big?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
DuncanDoyle: Oops, yes. Leftover from previous version (where I splitted Process and Tasks queries in 2 different extensions. Will fix.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/JBPMQueryServicesClientBuilder.java
DuncanDoyle: Not really. It started out as simple the TaskQueryServicesClient, so I wanted to have a clean split between the Process engine queries and the Task engine queries. However, we could combine them in one API. Question is, what should we call that API. We already have a Query API, so we need to come up with a proper name which describes that this is higher lever.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/pom.xml
DuncanDoyle: Shall we do that after PR merge (so create a JIRA for it), or do we want to do this in this PR?

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
DuncanDoyle: Depends on who "we" are. I just started adding tests for the major components. This one simply tests the "appComponents" method to validate it returns what we expect it to return. Not sure if that is very valuable.

DuncanDoyle:
The separation of ProcessQuery and TaskQuery came out of the fact that I originally created 2 different extensions for them (which was ridiculous), so that eventually ended up in 2 different interfaces. I agree that it's a bit of an overkill, and we could come up with a single interface. I'm only struggling with the name a bit. We already have a QueryService, that's why ProcessInstanceQueryServicesClient and TaskQueryServicesClient made sense from a "they describe what they do" point of view. So, any suggestions are appreciated.

The DB stuff came out of my concern that different DBs would need different queries and mappings (and my fetish for design patterns so combining a Strategy pattern with Factory pattern sort of felt satisfying :-)). If that is the case, I would need to re-design this part to cater for differences. With the current solution, the code can catch these differences and I can simply fix them by fixing the proper strategy.

I don't mind taking that strategy part out, as long as we're sure that we can deal with all different databases with a single query and a single set of mappings.

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
domhanak: Ignore if we can, maybe I got carried away a bit :-)

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm-queries/src/main/java/org/kie/server/remote/rest/jbpm/queries/ProcessInstanceQueryResource.java
domhanak: makes sense, thanks

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/ProcessInstanceQueryServiceBaseTest.java
```diff
@@ -0,0 +1,109 @@
...
+		assertEquals(1, processInstances.getItems().size());
+		
+		ProcessInstance pi1 = processInstances.getItems().stream().findFirst().get();
+		
+		assertEquals("test-process", pi1.getProcessName());
+		assertEquals(42L, pi1.getId().longValue());
+		assertEquals("mswiderski", pi1.getInitiator());
+	}
+	
+	private String getPayload() {
```
domhanak: Is there any better way to build/create the query and put it into payload? builder for example?

domhanak: Ignore this comment I misunderstood this tests :)

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/helper/JBPMQueryServicesClientBuilder.java
mswiderski: maybe simply SearchAPI or variation of it e.g. PredefinedSearch, ParameterizedSearch ...

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/pom.xml
mswiderski: fine with me to do that after this PR is merged, though remove the todo from code and maybe file jira instead

kie-server-parent/kie-server-services/kie-server-services-jbpm-queries/src/test/java/org/kie/server/services/jbpm/queries/JbpmQueriesKieServerExtensionTest.java
mswiderski: personally I don't see much value in such tests as there are others that would fail anyway if there is no resource deployed.

mswiderski:
As suggested above I think we can use SearchService as generic interface for the extension.

> I don't mind taking that strategy part out, as long as we're sure that we can deal with all different databases with a single query and a single set of mappings.

I'd opt at removing this unless there is really something not working. Though based on what we have so far I don't think this is the case

domhanak:
@DuncanDoyle  @mswiderski I do not see any additionns to <includes> into kie-server-wars assembly files. Aren't they required here? 
Also, locally I had to add <include>org.kie.server:kie-server-jbpm-queries-api</include> to kie-server-wars assembly flies to make my IntegrationTest module work.  Any ideas on why aren't includes of service and rest jbpm-queries modules sufficient?

DuncanDoyle:
@mswiderski: If we create the SearchService, should we also rename the extension from "jbpm-queries" to "jbpm-search"?

DuncanDoyle:
@domhanak I'll add the new modules to the kie-server-wars as well.

DuncanDoyle:
@mswiderski Currently I have a separate kie-server-jbpm-search-api JAR which contains the API for this extension. Is it OK if we just add this to kie-server-api? I think qe also need to add the TaskQueryFilterSpec and ProcessInstanceQueryFilterSpec to the KIE_SERVER_JAXB_CLASSES in JaxbMarshaller, so users don't need to manually add those classes to their client (via extraClasses). WDYT?

mswiderski:
Fine with me to include these classes in kie-server-api

domhanak:
@DuncanDoyle Missing unit tests for Task/ProcessQueryFilterSpecBuilders.
Right now I can create filters on String field comparing to integers and etc. which is not correct. IMO these builders should be unit tested.

mswiderski:
jenkins retest this

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/SearchServicesClientImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+
+import org.kie.server.api.model.instance.ProcessInstance;
+import org.kie.server.api.model.instance.ProcessInstanceList;
+import org.kie.server.api.model.instance.TaskInstance;
+import org.kie.server.api.model.instance.TaskInstanceList;
+import org.kie.server.client.KieServicesConfiguration;
+import org.kie.server.client.SearchServicesClient;
+import org.kie.server.jbpm.search.api.model.definition.ProcessInstanceQueryFilterSpec;
+import org.kie.server.jbpm.search.api.model.definition.TaskQueryFilterSpec;
```
domhanak: Imports from non-existent module fixed [here](https://github.com/kiegroup/droolsjbpm-integration/pull/967)

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/SearchServicesClient.java
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+package org.kie.server.client;
+
+import java.util.List;
+
+import org.kie.server.api.model.instance.ProcessInstance;
+import org.kie.server.api.model.instance.TaskInstance;
+import org.kie.server.jbpm.search.api.model.definition.ProcessInstanceQueryFilterSpec;
+import org.kie.server.jbpm.search.api.model.definition.TaskQueryFilterSpec;
```
domhanak: Imports from non-existent module fixed [here](https://github.com/kiegroup/droolsjbpm-integration/pull/967)

domhanak:
@DuncanDoyle there are still some dependencies on search-api module which you have removed. 
Also some imports form that api.
I have fixed it in this [PRQ](https://github.com/kiegroup/droolsjbpm-integration/pull/967) that also includes integration tests.

domhanak:
Check failing because of problems I have mentioned in new comments.

mswiderski:
@DuncanDoyle is this already with fixed date json marshaller?

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/util/TaskQueryFilterSpecBuilderTest.java
```diff
@@ -0,0 +1,16 @@
+package org.kie.server.api.util;
+
+import org.junit.Test;
+
+public class TaskQueryFilterSpecBuilderTest {
```
domhanak: @DuncanDoyle why is here a test-class with no tests?

DuncanDoyle: @domhanak Good question, don't know :-) Probably a placeholder that I never implemented ....

DuncanDoyle: @domhanak I've implemented the test class and also added a unit-test for ProcessInstanceQueryFilterSpecBuilder: https://github.com/DuncanDoyle/droolsjbpm-integration/commit/870dcbf74c1523b130dec00e06ac7bf20aca71a7

domhanak:
jenkins retest this

DuncanDoyle:
@domhanak Just pushed that fix, however that will fail the build until this PR has been merged: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/446

domhanak:
@DuncanDoyle Sure, thanks! Let's wait till it is merged and retest after.

domhanak:
jenkins retest this

PR #939, opened 2017-05-03
JBPM-5931 - Controller can update container from other container URL

tsurdilo:


tsurdilo:
@mswiderski looks to me that there is no impact on the ui parts

PR #943, opened 2017-05-06
RHBPMS-4592 Fix content-type from http header

muff1nman:
Accodring to rfc2616 a http media type may contain
a type/subtype followed by one or more parameters delimited by
semicolons.

### DIFF
--- a/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/MarshallingFormatTest.java
+++ b/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/MarshallingFormatTest.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.api.marshalling;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.assertEquals;
+
+public class MarshallingFormatTest {
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void testEmptyMarshallingFormat() {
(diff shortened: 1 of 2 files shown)


mswiderski:
wouldn't it be easier to use contains/startsWith instead of equalsIgnoreCase, rather than introduce (way more complex regex)?

so instead of 

```
if ("xstream".equalsIgnoreCase( type ) || "application/xstream".equalsIgnoreCase( type ) ) {
```

we would have

```
if ("xstream".equalsIgnoreCase( type ) || type.contains("application/xstream") ) {
````

and do the type.toLowerCase before so it covers it nicely. I believe this is way more readable to others.

P.S.
Please include test case as well

**Comment by muff1nman** [comment]
Added tests and took out regex.

File: kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/MarshallingFormatTest.java [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    public void testMarshallingFormatCase() {
+        assertEquals(MarshallingFormat.JSON, MarshallingFormat.fromType("JSON"));
+    }
+
+    @Test
+    public void testEdgeCaseWithJaxb() {
+        exception.expectMessage("Invalid marshalling format [jaxb]");
+        MarshallingFormat.fromType("jaxb");
+    }
```
muff1nman: Actually, thinking about it, I'm not sure this failure is desired as the text "JAXB" would have worked previously. @mswiderski thoughts?

mswiderski: correct, that would have worked. Though the special content type header should only be used for xstream as other types should rely on default application/json or application/xml

mswiderski:
ok to test

mswiderski:
@muff1nman thanks a lot for the changes, looks good to me, let's wait for the build to finish

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/MarshallingFormatTest.java
```diff
@@ -0,0 +1,57 @@
+package org.kie.server.api.marshalling;
```
mswiderski: could you please add license header?

muff1nman: Done. Can squash if desired.

mswiderski:
Jenkins retest this

mswiderski:
test failures seems related to this change, @muff1nman do you mind taking a quick look?

and once done feel free to squash :)

muff1nman:
Looking at just the first failure on the jenkins, it looks like its caused by not accepting "JAXB". I suppose I can put that back in, just seems weird to me to do that.

mswiderski:
Thanks @muff1nman

PR #948, opened 2017-05-09
RHBPMS-4375 - Container is not started when being created

mswiderski:


kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java
```diff
@@ -128,6 +128,16 @@ public synchronized void saveServerTemplate( ServerTemplate serverTemplate ) {
         }
 
         notificationService.notify( new ServerTemplateUpdated( serverTemplate ) );
+
+        Collection<ContainerSpec> containerSpecs = serverTemplate.getContainersSpec();
+        if (containerSpecs != null && !containerSpecs.isEmpty()) {
+            for (ContainerSpec containerSpec : containerSpecs) {
+                if (containerSpec.getStatus().equals(KieContainerStatus.STARTED)) {
+                    List<Container> containers = kieServerInstanceManager.startContainer( serverTemplate, containerSpec );
+                    notificationService.notify( serverTemplate, containerSpec, containers );
```
manstis: I assume the "notification service" also checks the ```ContainerSpec``` and starts the servers if applicable?

mswiderski: notification service is only responsible for updating other parts that might be interested - like UI. kieServerInstanceManager is the one that is responsible for actually broadcasting signals to connected kie servers for this given container spec.

mswiderski:
jenkins retest this

PR #949, opened 2017-05-10
[RHBRMS-2804] Align BRMS Feature Files for Fuse 6.3

mariofusco:


mariofusco:
@psiroky There are still a few failing test. Can you please give a look at them?

psiroky:
@grgrzybek surefire translates `group:artifact` to the `-tests` artifact (I tested that locally). The artifact needs to be on classpath though. It is true that not all of the tests are passing right now.

For some reason I thought the `kie-osgi-platforms/*` should stay as a way for us to verify that our libs are working in the Fuse runtime, even if we choose to upgrade some 3rd party dep (which would be in different version that the one in Fuse). However, I agree that this would require additional work. Not sure if it worth it. This was already merged into 6.4.x and 6.5.x branches. Should we then remove it from there?

psiroky:
@mariofusco I was running the tests yesterday locally and they passed. I believe the current fail is caused by a recent drools commit: https://github.com/kiegroup/drools/commit/a8016c32b651431f6067ec6191a7a079b35e156a

grgrzybek:
@psiroky 

> Should we then remove it from there?

I'd like to have it there. I planned to run the tests when there's a need to check some changed dependency, but these tests should rather never be run automatically.

psiroky:
@grgrzybek tbh, having tests which should never be run automatically sounds weird to me. However, I definitely do not have the whole context here.

So we should probably revert the changes I did, correct?

psiroky:
@mariofusco could you please drop the last commit https://github.com/kiegroup/droolsjbpm-integration/pull/949/commits/cd465d6f4eb18edc78d0eddbc75d615408d02c0f? After a discussion with @grgrzybek we agreed the original approach is better in this case.

mariofusco:
Replace by https://github.com/kiegroup/droolsjbpm-integration/pull/957

PR #950, opened 2017-05-10
[reproducer] kie-server-tests: update not existing release id test

sutaakar:
Reproducer for https://issues.jboss.org/browse/DROOLS-1562 .
The test should be updated according to fix resolution.


mariofusco:
Reproduced and fixed in drools with this https://github.com/kiegroup/drools/pull/1270

PR #958, opened 2017-05-15
JBPM-5946 Added Standalone controller

lanceleverich:
Added the following modules
* kie-server-controller-standalone
* kie-server-controller-standalone-war
Added the FileBasedServerTemplateStorage class to kie-server-controller-impl. This class provides a file-based persistence of server templates for the standalone controller.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+package org.kie.server.controller.impl.storage;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.StringUtils;
+import org.drools.core.util.ClassUtils;
+import org.kie.server.api.marshalling.Marshaller;
+import org.kie.server.api.marshalling.MarshallerFactory;
+import org.kie.server.api.marshalling.MarshallingFormat;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.controller.api.model.spec.ServerTemplateKey;
+import org.kie.server.controller.api.storage.KieServerTemplateStorage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FileBasedKieServerTemplateStorage implements KieServerTemplateStorage {
+    public static final String SERVER_TEMPLATE_FILE_NAME_PROP = "org.kie.server.controller.templatefile";
+    public static final String DEFAULT_SERVER_TEMPLATE_FILENAME = "/opt/kie/server-controller/templates.xml";
+    private static Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorage.class);
+    private static FileBasedKieServerTemplateStorage INSTANCE = new FileBasedKieServerTemplateStorage();
+    private Map<String, ServerTemplate> templateMap = new ConcurrentHashMap<>();
+    private Map<String, ServerTemplateKey> templateKeyMap = new ConcurrentHashMap<>();
+    private String templatesLocation = System.getProperty(SERVER_TEMPLATE_FILE_NAME_PROP, DEFAULT_SERVER_TEMPLATE_FILENAME);
+    private Marshaller templateMarshaller = MarshallerFactory.getMarshaller(MarshallingFormat.XSTREAM,ServerTemplate.class.getClassLoader());
+
+    private static class ClassCaster {
+        @SuppressWarnings("unchecked")
+        public static <T> Class<T> castClass(Class<?> clazz) {
+            return (Class<T>) clazz;
+        }
+    }
+
+    protected FileBasedKieServerTemplateStorage() {
+    }
+
+    /**
+     * Writes the map of server templates to the file pointed at by templatesLocation
+     */
+    private void writeTemplateMap() {
+        try (FileOutputStream target = new FileOutputStream(templatesLocation)) {
+            List<ServerTemplate> templates = templateMap.values().stream().collect(Collectors.toList());
+            String templatesString = templateMarshaller.marshall(templates);
+            target.write(templatesString.getBytes());
+        } catch (Throwable e) {
+            logger.error("Unable to write template maps for standalone controller",e);
+        }
+    }
+
+    /**
+     * Loads the map of server templates from the file pointed at by the templatesLocation
+     */
+    private synchronized void loadTemplateMapsFromFile() {
+        StringBuilder stringBuilder = new StringBuilder();
+        try(InputStream inputStream = new FileInputStream(templatesLocation)) {
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream));
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                stringBuilder.append(line);
+            }
+        } catch (Throwable e) {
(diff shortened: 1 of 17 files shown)


sutaakar:
One idea on the beginning, can you please replace tabs with spaces to keep codebase consistent?

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneController.java
```diff
@@ -0,0 +1,382 @@
...
+        Set<KieServerInstanceInfo> infoSet = new HashSet<>();
+        infoSet.add(instanceInfo);
+        KieServerInstance instance = new KieServerInstance(serverInfo.getServerId(),serverInfo.getName(),serverInfo.getVersion(),infoSet,KieServerStatus.UNKNOWN,setup);
+        kieServerControllerStorage.store(instance);
+
+        return createCorrectVariant(responseString, headers,Response.Status.CREATED);
+    }
+
+    @PUT
+    @Path("server/{id}/containers")
```
jakubschwan: Missing container id in path for container save.
I think that this should be same as for REST controller from KIE-WB:
[PUT] /management/servers/{id}/containers/{containerId}

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+    public static final String SERVER_TEMPLATE_FILE_NAME_PROP = "org.kie.server.controller.templatefile";
+    public static final String DEFAULT_SERVER_TEMPLATE_FILENAME = "/opt/kie/server-controller/templates.xml";
+    private static Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorage.class);
+    private static FileBasedKieServerTemplateStorage INSTANCE = new FileBasedKieServerTemplateStorage();
+    private Map<String, ServerTemplate> templateMap = new ConcurrentHashMap<>();
+    private Map<String, ServerTemplateKey> templateKeyMap = new ConcurrentHashMap<>();
+    private String templatesLocation = System.getProperty(SERVER_TEMPLATE_FILE_NAME_PROP, DEFAULT_SERVER_TEMPLATE_FILENAME);
+    private Marshaller templateMarshaller = MarshallerFactory.getMarshaller(MarshallingFormat.XSTREAM,ServerTemplate.class.getClassLoader());
+
+    private static class ClassCaster {
```
tsurdilo: Do we really need this? Does not throw or handle ClassCastException

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+import org.kie.server.api.marshalling.MarshallingFormat;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.controller.api.model.spec.ServerTemplateKey;
+import org.kie.server.controller.api.storage.KieServerTemplateStorage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FileBasedKieServerTemplateStorage implements KieServerTemplateStorage {
+    public static final String SERVER_TEMPLATE_FILE_NAME_PROP = "org.kie.server.controller.templatefile";
+    public static final String DEFAULT_SERVER_TEMPLATE_FILENAME = "/opt/kie/server-controller/templates.xml";
```
mswiderski: this won't work on windows at all so I'd say we should remove it or place it by default either in the current directory where jvm starts or in temp dir that is given via system property - java.io.tmpdir

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+    private static Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorage.class);
+    private static FileBasedKieServerTemplateStorage INSTANCE = new FileBasedKieServerTemplateStorage();
+    private Map<String, ServerTemplate> templateMap = new ConcurrentHashMap<>();
+    private Map<String, ServerTemplateKey> templateKeyMap = new ConcurrentHashMap<>();
```
mswiderski: +1 it's the same instances so don't see any need for keeping them next to templateMap

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+
+    protected FileBasedKieServerTemplateStorage() {
+    }
+
+    /**
+     * Writes the map of server templates to the file pointed at by templatesLocation
+     */
+    private void writeTemplateMap() {
+        try (FileOutputStream target = new FileOutputStream(templatesLocation)) {
+            List<ServerTemplate> templates = templateMap.values().stream().collect(Collectors.toList());
```
mswiderski: do we really need to go via stream here? isn't it enough to create ArrayList(templateMap.values())?

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,277 @@
...
+import org.kie.server.controller.impl.service.SpecManagementServiceImpl;
+import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+
+public class FileBasedKieServerTemplateStorageTest {
+	private static final Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorageTest.class);
+	private static final String TEST_SERVER_TEMPLATE_FILENAME = "/tmp/templates.xml";
```
mswiderski: this will fail on windows jenkins slaves if I am not mistaken

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+
+public class FileBasedKieServerTemplateStorage implements KieServerTemplateStorage {
+    public static final String SERVER_TEMPLATE_FILE_NAME_PROP = "org.kie.server.controller.templatefile";
+    public static final String DEFAULT_SERVER_TEMPLATE_FILENAME = "/opt/kie/server-controller/templates.xml";
+    private static Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorage.class);
+    private static FileBasedKieServerTemplateStorage INSTANCE = new FileBasedKieServerTemplateStorage();
+    private Map<String, ServerTemplate> templateMap = new ConcurrentHashMap<>();
+    private Map<String, ServerTemplateKey> templateKeyMap = new ConcurrentHashMap<>();
+    private String templatesLocation = System.getProperty(SERVER_TEMPLATE_FILE_NAME_PROP, DEFAULT_SERVER_TEMPLATE_FILENAME);
+    private Marshaller templateMarshaller = MarshallerFactory.getMarshaller(MarshallingFormat.XSTREAM,ServerTemplate.class.getClassLoader());
```
mswiderski: I would actually use Stream directly instead of kie server marshallers as it will make it easier - for instance read from input stream etc.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,198 @@
...
+            target.write(templatesString.getBytes());
+        } catch (Throwable e) {
+            logger.error("Unable to write template maps for standalone controller",e);
+        }
+    }
+
+    /**
+     * Loads the map of server templates from the file pointed at by the templatesLocation
+     */
+    private synchronized void loadTemplateMapsFromFile() {
```
mswiderski: see comment about using xstream directly

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneController.java
```diff
@@ -0,0 +1,382 @@
...
+import org.kie.server.controller.impl.KieServerInstanceManager;
+import org.kie.server.controller.impl.service.RuntimeManagementServiceImpl;
+import org.kie.server.controller.impl.service.SpecManagementServiceImpl;
+import org.kie.server.controller.impl.storage.FileBasedKieServerTemplateStorage;
+import org.kie.server.controller.impl.storage.InMemoryKieServerControllerStorage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/standalone")
+public class StandaloneController extends KieServerControllerImpl {
```
mswiderski: we should use the default RestKieServerContollerImpl and they should be mapped to the same urls - the only thing that should differ is the context path of the app - which is then in hands of users. So any client code written now towards workbench's controller will work here.

lanceleverich:
Making significant changes to the approach for this application. Consequently, I will close this PR and create a new one when I have finished with all of the changes.

PR #962, opened 2017-05-16
kie-server-tests: Raise scanning poll interval

sutaakar:
Interval set to 20ms is too small, it causes issues on IBM java.
Scanning consumed so many resources making JMS tests occasionally timed out.


psiroky:
If I remember correctly (maybe I don't though :)), the test uses this very small interval on purpose. That was the only way to reproduce the bug in the kie-server. Increasing the interval likely means, the original bug would not be caught. So we could in the end remove the test altogether (which I am not opposed to at all).

sutaakar:
@psiroky yes, it was reproducer: https://github.com/kiegroup/droolsjbpm-integration/pull/710 . The issue should be covered by https://github.com/kiegroup/drools/commit/c40bced38032399f9f3629ba0aa87c52ded7fad7 . I think we could delete this test.

sutaakar:
jenkins retest this

PR #966, opened 2017-05-17
JBPM-5946 - Create Standalone Controller

lanceleverich:
JBPM-5946 - Add Standalone KIE Server Controller
    * kie-server-controller-standalone module contains classes to extend those found in kie-server-controller-rest
    * kie-server-controller-standalone-war module to create deployable artifacts that are based on kie-server-controller-standalone
    * kie-server-controller-impl has FileBasedKieServerTemplateStorage class to provide persistence of server templates


kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+                                   serverTemplate);
+            writeTemplateMap();
+        }
+        return template;
+    }
+
+    @Override
+    public List<ServerTemplateKey> loadKeys() {
+        if (templateKeyMap.isEmpty()) {
+            loadTemplateMapsFromFile();
```
mswiderski: I'd say this should be loaded when new instance of this class is being initialized otherwise there is a chance that multiple loads will be performed in multi threaded environment - e.g. when several kie servers connects at pretty much same time.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        return new ArrayList<ServerTemplateKey>(templateKeyMap.values());
+    }
+
+    @Override
+    public List<ServerTemplate> load() {
+        if (templateKeyMap.isEmpty()) {
+            loadTemplateMapsFromFile();
```
mswiderski: same as above

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        return new ArrayList<ServerTemplate>(templateMap.values());
+    }
+
+    @Override
+    public ServerTemplate load(String identifier) {
+        if (templateKeyMap.isEmpty()) {
+            loadTemplateMapsFromFile();
```
mswiderski: same as above

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        return templateMap.get(identifier);
+    }
+
+    @Override
+    public boolean exists(String identifier) {
+        if (templateKeyMap.isEmpty()) {
+            loadTemplateMapsFromFile();
```
mswiderski: same as above

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        return templateMap.containsKey(identifier);
+    }
+
+    @Override
+    public ServerTemplate update(ServerTemplate serverTemplate) {
+        ServerTemplate updated = null;
+        synchronized (templateMap) {
+            if (templateKeyMap.isEmpty()) {
+                loadTemplateMapsFromFile();
```
mswiderski: and here as well.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,169 @@
...
+import org.slf4j.LoggerFactory;
+
+public class FileBasedKieServerTemplateStorage implements KieServerTemplateStorage {
+    public static final String SERVER_TEMPLATE_FILE_NAME_PROP = "org.kie.server.controller.templatefile";
+    public static final String DEFAULT_SERVER_TEMPLATE_FILENAME = System.getProperty("java.io.tmpdir")+
+    		System.getProperty("file.separator")
+    		+"template_store.xml";
+    private static Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorage.class);
+    private Map<String, ServerTemplate> templateMap = new ConcurrentHashMap<>();
+    private Map<String, ServerTemplateKey> templateKeyMap = new ConcurrentHashMap<>();
```
mswiderski: what's the reason for having both templateMap and templateKeyMap?

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,279 @@
...
+import org.kie.server.controller.impl.service.SpecManagementServiceImpl;
+import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+
+public class FileBasedKieServerTemplateStorageTest {
+    private static final Logger logger = LoggerFactory.getLogger(FileBasedKieServerTemplateStorageTest.class);
+    private static final String TEST_SERVER_TEMPLATE_FILENAME = System.getProperty("java.io.tmpdir")
```
mswiderski: this would be good to have placed in target folder so it does not affect anything else when there are parallel builds on same machine - jenkins

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,279 @@
...
+            StringBuilder templateName = new StringBuilder("test server : ").append(x);
+            ServerTemplate template = createServerTemplateWithContainer(templateName.toString(),x+1);
+            ServerTemplateKey key = new ServerTemplateKey(template.getId(), template.getName());
+            templateMap.put(key, template);
+        }
+    }
+    
+    @Before
+    public void setup() {
+        storage = new FileBasedKieServerTemplateStorage();
```
mswiderski: for tests (to avoid to set and reset system properties) we could have extra constructor that accepts file path directly.

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
```diff
@@ -0,0 +1,43 @@
...
+
+
+    public StandaloneKieServerControllerImpl() {
+        super();
+        ServiceLoader<KieServerTemplateStorage> storages = ServiceLoader.load(KieServerTemplateStorage.class);
+        if (storages != null && storages.iterator().hasNext()) {
+            KieServerTemplateStorage storage = storages.iterator().next();
+            this.setTemplateStorage(storage);
+            logger.debug("Server template storage for standalone kie server controller is {}",storage.getClass().getName());
+            if (this.getTemplateStorage() instanceof FileBasedKieServerTemplateStorage) {
```
mswiderski: maybe instead of this if implement this as part of toString method of FileBasedKieServerTemplateStorage so it could be generic and reusable without modifying this class for new implementations of the storage

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.slf4j.LoggerFactory;
+
+public class StandaloneKieServerControllerImpl extends RestKieServerControllerImpl {
+    private static Logger logger = LoggerFactory.getLogger(StandaloneKieServerControllerImpl.class);
+
+
+    public StandaloneKieServerControllerImpl() {
+        super();
+        ServiceLoader<KieServerTemplateStorage> storages = ServiceLoader.load(KieServerTemplateStorage.class);
+        if (storages != null && storages.iterator().hasNext()) {
```
mswiderski: shouldn't we throw exception otherwise - if there are not storage implementations found?

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StandaloneKieServerControllerImpl extends RestKieServerControllerImpl {
+    private static Logger logger = LoggerFactory.getLogger(StandaloneKieServerControllerImpl.class);
+
+
+    public StandaloneKieServerControllerImpl() {
+        super();
+        ServiceLoader<KieServerTemplateStorage> storages = ServiceLoader.load(KieServerTemplateStorage.class);
```
mswiderski: +1 for service loader - one question though - how would users add their own implementation?

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/ControllerUtils.java
```diff
@@ -0,0 +1,232 @@
...
+import org.kie.server.controller.api.model.spec.ContainerSpecKey;
+import org.kie.server.controller.api.model.spec.ContainerSpecList;
+import org.kie.server.controller.api.model.spec.ProcessConfig;
+import org.kie.server.controller.api.model.spec.RuleConfig;
+import org.kie.server.controller.api.model.spec.ServerConfig;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.controller.api.model.spec.ServerTemplateKey;
+import org.kie.server.controller.api.model.spec.ServerTemplateList;
+
+public class ControllerUtils {
```
sutaakar: This class seems to be copy of org.kie.server.controller.rest.ControllerUtils.

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/resources/META-INF/services/org.kie.server.controller.api.storage.KieServerTemplateStorage
```diff
@@ -0,0 +1 @@
+org.kie.server.controller.impl.storage.FileBasedKieServerTemplateStorage
```
sutaakar: It would be good to place this file into same module where FileBasedKieServerTemplateStorage is located. It will be less error prone in case of future changes.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,279 @@
...
+    private static Map<ServerTemplateKey, ServerTemplate> templateMap;
+    private FileBasedKieServerTemplateStorage storage;
+    
+    /**
+     * Method that creates a server template instance and saves it in the specManageService
+     * @param templateName The name of the template being created
+     * @param templateCount The expected number of templates saved in the specManageService, including the one being created 
+     * @return A new server template instance
+     */
+    private static ServerTemplate createServerTemplateWithContainer(String templateName, int templateCount) {
```
sutaakar: This method is very similar to AbstractServiceImplTest.createServerTemplateWithContainer(). Can you modify it to satisfy your needs and reduce duplication?

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
sutaakar: It seems that this storage isn't passed to specManageService and runtimeManagementService. Do you need these services in your tests?

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
lanceleverich: There is a method that returns a list of ServerTemplateKey objects, and it is faster to grab the keys from the map than it would be to generate them each time. We could store the ServerTemplateKey objects in a List, instead of the Map, but I don't think it would save us much. I can add a method to retrieve a ServerTemplateKey, based on a passed in id. That would at least give a reason for the map to exist.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
lanceleverich: The version of this method, in FileBasedKieServerTemplateStorageTest, takes two parameters that are necessary for the purpose of testing multiple template situations. The version in AbstractServiceImplTest only expects the creation of a single server template (see line 62 of AbstractServiceImplTest). If necessary, I can break apart the method in AbstractServiceImplTest and create multiple methods that do the work that isn't dependent on the server template name and the number of templates that are expected to be in the specManagementService. I'm not sure that this is worth that level of effort, though.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
lanceleverich: Those service instances are being removed as they did not contribute to the actual testing of the FileBasedKieServerTemplateStorage

sutaakar: ok

mswiderski:
There are still few not answered/addressed comments, could you please take a look at them, @lanceleverich

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
lanceleverich: Template maps are now loaded during constructor/initialization of the object

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
lanceleverich: Service definition has been moved to the WAR construction project. Users could change the value there or in their own WAR assembly project.

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
lanceleverich: Storage defaults to InMemoryKieServerTemplateStorage, with a WARN in the log file, "No server template storage defined. Default storage: InMemoryKieServerTemplateStorage will be used"

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
lanceleverich: Added toString method to the FileBasedKieServerTemplateStorage and updated StandaloneKieServerControllerImpl and StandaloneSpecManagementServiceImpl to use the new method.

mswiderski:
And what about ControllerUtils class that seems to be a duplicate from controller impl project? I believe it should be removed and standalone controller should rely on already existing

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/ControllerUtils.java
lanceleverich: File removed from the kie-server-controller-standalone project

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lanceleverich: File removed from the kie-server-controller-standalone project

mswiderski:
Jenkins retest this

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
```diff
@@ -0,0 +1,212 @@
...
+        return new ArrayList<ServerTemplateKey>(templateKeyMap.values());
+    }
+    
+    /**
+     * Returns the ServerTemplateKey object associated with a server template's id
+     * @param id The server template id that points to the ServerTemplateKey
+     * @return The ServerTemplateKey that is associated with the id, or null if the
+     * id does exist in the templateKeyMap
+     */
+    public synchronized ServerTemplateKey getTemplateKey(String id) {
```
mswiderski: Why is this synchronized? We only read from the map

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
mswiderski: In that case do we need this loadTemplateMapsFromFile here?

kie-server-parent/kie-server-controller/kie-server-controller-standalone-war/src/main/shared-ee6-ee7-resources/META-INF/services/org.kie.server.controller.api.storage.KieServerTemplateStorage
```diff
@@ -0,0 +1 @@
+org.kie.server.controller.impl.storage.FileBasedKieServerTemplateStorage
```
mswiderski: I belive this will not be included in web container assembly, could you please double check after the build?

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
mswiderski: ServiceLoader will create new instance of the template storage impl so this means it will load twice from backend which might not be too big issue but it might result in unexpected results as they might get out of sync and can lead to loss of data when different instances will save the data.
This refers to StandaloneKieServerControllerImpl and StandaloneSpecManagementServiceImpl

kie-server-parent/kie-server-controller/kie-server-controller-standalone-war/src/main/shared-ee6-ee7-resources/META-INF/services/org.kie.server.controller.api.storage.KieServerTemplateStorage
lanceleverich: It was showing up in the ee6 and ee7 war files. It should now show up in all war files generated.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
lanceleverich: Keyword `synchronized` has been removed

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorage.java
lanceleverich: Missed one.

sutaakar:
@mswiderski Yes, I think that this controller could be used as full replacement of kie-server-controller-test-war

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,252 @@
...
+        try {
+            Files.deleteIfExists(tmpTemplateStore.toPath());
+        } catch (IOException e) {
+            logger.warn("Exception while deleting test server template storage",e);
+            e.printStackTrace();
+        }
+    }
+    
+    @AfterClass
+    public static void afterClass() {
```
jakubschwan: If this empty method is not needed, then delete it please.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,252 @@
...
+         */
+        assertEquals("Mismatched number of server template keys",templateMap.keySet().size(),keys.size());
+        templateMap.keySet().forEach(key -> {
+            assertTrue("Key for server template not found",keys.contains(key));
+        });
+    }
+    
+    @Test
+    public void testLoadList() {
+        storage.reloadTemplateMaps();;
```
jakubschwan: Two ';'

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
```diff
@@ -0,0 +1,252 @@
...
+        List<ServerTemplate> templates = storage.load();
+        assertEquals("Mismatched number of server templates",templateMap.values().size(),templates.size());
+        templateMap.values().forEach(value -> {
+            assertTrue("Server template not found",templates.contains(value));
+        });
+    }
+ 
+    @Test
+    public void testLoadSingle() {
+        storage.reloadTemplateMaps();;
```
jakubschwan: Two ';'

mswiderski:
@sutaakar in that case I'd say we should have the switch done here that way we ensure that standalone controller can do exactly the same things as the one we already tested with

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
lanceleverich: Deleted

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
lanceleverich: Second ; removed

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/storage/FileBasedKieServerTemplateStorageTest.java
lanceleverich: Second ; removed

kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
lanceleverich: ServiceLoader will now load a service that provides a method to get a singleton instance of the FileBasedKieServerTemplateStorage

sutaakar:
Just one thing, I still miss possibility to add custom persistent storage. For example when I create jar with database storage, I can register it for service loader, but I don't have any possibility how to tell controller to use my database storage instead of FileBasedKieServerTemplateStorage.

lanceleverich:
That's not correct.
Your jar needs to contain a class that implements
PersistingServerTemplateStorageService. In that class (e.g.
DbServerTemplateStorageService), like in the
FileBasedServerTemplateStorageService, you would get a reference to the
singleton instance of your new persistence class (e.g.
DbKieServerTemplateStorage). The StandaloneKieServerControllerImpl and the
StandaloneSpecManagementServiceImpl use the ServiceLoader to retrieve a
PersistingServerTemplateStorageService, and call its getTemplateStorage
method to retrieve whatever persistent storage model you had created.

Does that make sense?


Lance Leverich

sutaakar:
Yes, just we need to make sure that only one PersistingServerTemplateStorageService implementation is retrievable by service loader as current implementation takes first service it founds and uses it: https://github.com/kiegroup/droolsjbpm-integration/blob/cde5bb220dd0a05f390b02fdadaa233457e76070/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java#L34 . In this case when user adds e.g. DbServerTemplateStorageService he needs to edit value of https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-standalone-war/src/main/shared-ee6-ee7-resources/META-INF/services/org.kie.server.controller.api.service.PersistingServerTemplateStorageService to instruct service loader to use his service.

lanceleverich:
Correct.

romartin:
Hey @lanceleverich , @mswiderski 
I'm having some compile errors on the stunner showcase, see for example [this Jenkins' build log](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/1249/console).
I think there is an error on the `kie-server-controller-api` GWT module descriptor:
- The `PersistingServerTemplateStorageService` depends on `KieServerTemplateStorage`, [see here](https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/service/PersistingServerTemplateStorageService.java#L27).
- But the package  for `KieServerTemplateStorage` is not included in the GWT module descriptor, [see here](https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/resources/org/kie/server/controller/api/KieServerControllerAPI.gwt.xml#L10).
So finally the GWT compilation fails. Can you please double-check? 
Thanks!

cristianonicolai:
@romartin same error here with the jbpm-wb-showcase, pushing a fix here: https://github.com/kiegroup/droolsjbpm-integration/pull/977

romartin:
oh great @cristianonicolai !! thanks! :)

tsurdilo:
@romartin @cristianonicolai fix merged: https://github.com/kiegroup/droolsjbpm-integration/commit/3ae3e721eec3dc795cdccb58d64afe39fa78d176

romartin:
thanks @tsurdilo !!

sutaakar:
@lanceleverich Ok.
One more thing, can you please add new artifacts to https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/kie-bom/pom.xml ?

sutaakar:
@lanceleverich One more thing, just tried to build controller locally. It is installed to maven repo based on project artifact id, in this case artifacts are named like "kie-server-controller-standalone-war-7.1.0-SNAPSHOT-ee7.war". It would be better if they won't contain "war" in artifactId (to be for example "kie-server-controller-standalone-7.1.0-SNAPSHOT-ee7.war"), to make name simpler and nicer.

PR #971, opened 2017-05-22
DROOLS-1565 Implement a KieScanner listener

tarilabs:


tarilabs:
Added kiegroup/drools#1279

Jenkins test again

tarilabs:
Added commit to import `org.kie.api.event.kiescanner` in addition to `org.drools.core.event` in referenced PR.

Jenkins, test again.

PR #975, opened 2017-05-22
[DROOLS-1571] allow kie-spring to find resources even when the contex…

mariofusco:
…t file is defined in a different path

### DIFF
--- a/kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
+++ b/kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
@@ -16,6 +16,12 @@
 
 package org.kie.spring;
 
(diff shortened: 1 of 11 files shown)


kie-spring/src/main/java/org/kie/spring/KModuleBeanFactoryPostProcessor.java
```diff
@@ -275,6 +281,23 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
         log.debug("KieModule root URL (based on application context {}): {}", applicationContext.getDisplayName(), kModuleRootUrl);
     }
 
+    private URL tryGetRootUrlFromContextFile(ApplicationContext applicationContext) {
```
tarilabs: ok

PR #991, opened 2017-05-26
[GSS](6.4.x)Endpoint /history/variable/{varId}/value/{value} not working for variables with space

markcoble:
@krisv @mswiderski it's not much,  but it does test that the regex from the Path annotation works for a given test value.


krisv:
+1

PR #992, opened 2017-05-27
JBPM-6012 Standalone controller can't be deployed on Tomcat 8 container

lanceleverich:
- Re-applied the dependency on kie-server-rest-common, in kie-server-controller-standalone-impl. Removal of this dependency causes the assembly to not include the rest-easy jars in the resultant war files.
- Removed the include of slf4j-jdk14, and added an exclude of slf4j-api, to the assembly file for Tomcat (servlet container)
- Added dependency on kie-server-controller-standalone-impl to kie-server-integ-tests-common
- Changed KieControllerExecutor (in kie-server-integ-tests-common) so that it uses the new Standalone...Impl classes
- Updated kie-server-integ-tests-controller pom to use standalone components, and the pingURL to point to a valid endpoint
- Updated kie-server-tests pom to include a system property that provides a template file location for the controller

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-standalone-impl/pom.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-standalone-impl/pom.xml
@@ -12,6 +12,7 @@
   <artifactId>kie-server-controller-standalone-impl</artifactId>
   <name>KIE :: Execution Server :: Controller :: Standalone Implementation</name>
   <description>KIE Execution Server Controller Standalone Implementation</description>
+  <packaging>jar</packaging>
 
   <dependencies>
     <dependency>
@@ -42,6 +43,11 @@
     <dependency>
       <groupId>org.kie.server</groupId>
       <artifactId>kie-server-common</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-rest-common</artifactId>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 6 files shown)


mswiderski:
@lanceleverich adding back kie-server-rest-common means now we have resteasy in ee6 and ee7 assemblies as well which most likely will cause issues on other app servers that use another jax-rs impl than resteasy. Take a look here on excludes https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-wars/kie-server/src/main/assembly/assembly-ee7-container.xml#L84

Slf4f-jdk is prefered for logging in tomcat so would be good to have it

Since controller tests now use the standalone I think the test war of controller can be completely removed

lanceleverich:
New commit addresses the issues raised regarding the placement of dependency for kie-server-rest-common and the use of includes/excludes in the assembly files for kie-server-controller-standalone. The kie-server-controller-test-war module has been removed, and the kie-server-controller pom file updated to take into account the removal.

sutaakar:
@lanceleverich I have a question. Why do you reintroduce kie-server-rest-common dependency? Do you need it just for resteasy or are there other reasons to have it?

sutaakar:
@lanceleverich My point is that if we add unneeded dependencies to controller then it affects usability of controller. Unneeded dependencies increase size of resulting war, increase risk of class incompatibility as for some application servers we need to load war classes before parent ones, it also decreases extensibility in case users would like to extend/reuse our controller war.

lanceleverich:
Removed the dependency on kie-server-rest-common and added dependency on resteasy-jaxrs.

sutaakar:
@lanceleverich It seems that your merge mixed up things, for example returning back kie-server-rest-common to assembly filters. Please see kie-server-controller-standalone assembly output log to see inclusion/exclusion filters which weren't triggered.

sutaakar:
@lanceleverich Looks good now. One last thing, can you please exclude resteasy artifact from ee6 and ee7 assemblies? It may cause issues as Maciej mentioned.

PR #836, opened 2017-03-02
DROOLS-1460 DMN extension for kie-server

tarilabs:
depends on droolsjbpm/droolsjbpm-build-bootstrap#402 .

PR #994, opened 2017-05-29
Clean revapi-config after 7.0.0.Final release

MarianMacik:
Hi, @psiroky,

I have cleaned revapi ignores since we have a new Final version.

Marian


MarianMacik:
Previous run failed because revapi config file was missing in kie-server-rest-dmn module. I really do not know why PR #836 has passed previously. But now it should work.

PR #996, opened 2017-06-02
JBPM-6048: Test for process update by Kie scanner

sutaakar:


mswiderski:
@mariofusco do we properly support LATEST in kie scanner? I remember some issues with it in general - maven wise...

jhrcek:
@mswiderski Here's relevant discussion from the past: https://issues.jboss.org/browse/RHBRMS-901

mswiderski:
@sutaakar @jhrcek  So taking this into account is this still a valid issue?

sutaakar:
@mswiderski I have noticed similar behaviour with SNAPSHOT version, will update PR and JIRA

sutaakar:
Issue is related to https://issues.jboss.org/browse/DROOLS-1603

mariofusco:
@mswiderski @sutaakar Since maven no longer supports RELEASE and LATEST metaversion of course we also cannot support them

mswiderski:
@sutaakar after looking into it in more details the problem is not that kie scanner is not updating the kbase with new processes, but jbpm services are not aware of the updates. Since kie scanner directly updates the kbase but does not notify in anyway upper layers (jbpm services) so they do not "index" new things. Same applies to remove or update of processes.

In general, kie scanner with processes is discouraged as that might lead to unforeseen updates which can then cause unexpected errors in case of long running processes when changes are not compatible with running instances - which might actually be quite common.

We could try to fix it based on recently implemented KieScannerListener (https://github.com/kiegroup/droolsjbpm-knowledge/commit/11e504ae88a27d313afa1642640d3a1cc21419fc) though I do believe it could lead to more trouble on users than the value. 

I opt at closing this as won't fix. Thoughts?

sutaakar:
Sounds reasonable, processes should be rather updated by direct container update call.
About KieScannerListener, I think it should be implemented to Kie server anyway, to to make scanner usage in Kie server cleaner and less error prone [1].

[1] https://issues.jboss.org/browse/DROOLS-1603

mswiderski:
@sutaakar agreed, processes should be upgraded via dedicated endpoint/call instead of kie scanner. 

So can we then close this PR and JIRA?

sutaakar:
@mswiderski Sure

PR #999, opened 2017-06-07
[DROOLS-1608] move processes related classes from drools to jbpm

mariofusco:


mariofusco:
depends on kiegroup/drools#1307

PR #1003, opened 2017-06-13
PLANNER-815: Execution server: Support real-time planning

mcimbora:
@ge0ffrey First implementation of real-time planning on the ES side.

From https://issues.jboss.org/browse/PLANNER-815:
The problem fact changes will be registered by sending POST request to solverId/problemfactchanged with the following payload:

XSTREAM:
```
<org.kie.server.testing.AddComputerProblemFactChange>
  <computer>
    <id>5</id>
    <cpuPower>3</cpuPower>
    <memory>4</memory>
    <networkBandwidth>2</networkBandwidth>
    <cost>390</cost>
  </computer>
</org.kie.server.testing.AddComputerProblemFactChange>
```

where AddComputerProblemFactChange is a ProblemFactChange implementation (taken from optaplanner-examples with slight modifications). See JIRA for more details.

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/CommandScript.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/CommandScript.java
@@ -28,8 +28,11 @@ import javax.xml.bind.annotation.XmlType;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
+import org.kie.server.api.commands.optaplanner.AddProblemFactChangeCommand;
+import org.kie.server.api.commands.optaplanner.AddProblemFactChangesCommand;
 import org.kie.server.api.commands.optaplanner.CreateSolverCommand;
 import org.kie.server.api.commands.optaplanner.DisposeSolverCommand;
+import org.kie.server.api.commands.optaplanner.IsEveryProblemFactChangeProcessedCommand;
 import org.kie.server.api.commands.optaplanner.GetSolverCommand;
 import org.kie.server.api.commands.optaplanner.GetSolverWithBestSolutionCommand;
 import org.kie.server.api.commands.optaplanner.GetSolversCommand;
@@ -72,7 +75,10 @@ public class CommandScript implements Serializable {
             @XmlElement(name = "get-solvers", type = GetSolversCommand.class),
             @XmlElement(name = "get-solver", type = GetSolverCommand.class),
             @XmlElement(name = "start-solver", type = SolvePlanningProblemCommand.class),
-            @XmlElement(name = "terminate-solver", type = TerminateSolverEarlyCommand.class)
+            @XmlElement(name = "terminate-solver", type = TerminateSolverEarlyCommand.class),
+            @XmlElement(name = "add-problem-fact-change", type = AddProblemFactChangeCommand.class),
+            @XmlElement(name = "add-problem-fact-changes", type = AddProblemFactChangesCommand.class),
+            @XmlElement(name = "is-every-problem-fact-change-processed", type = IsEveryProblemFactChangeProcessedCommand.class)
     })
     @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
     protected List<KieServerCommand> commands;
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,15 +16,20 @@
 
 package org.kie.server.api.commands.optaplanner;
 
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import org.kie.server.api.model.KieServerCommand;
+import org.optaplanner.core.impl.solver.ProblemFactChange;
 
-import javax.xml.bind.annotation.*;
-
-@XmlRootElement(name = "create-solver")
-@XStreamAlias("create-solver")
+@XmlRootElement(name = "add-problem-fact-change")
+@XStreamAlias("add-problem-fact-change")
 @XmlAccessorType(XmlAccessType.NONE)
-public class CreateSolverCommand
+public class AddProblemFactChangeCommand
         implements KieServerCommand {
 
     private static final long serialVersionUID = -1803374525440238478L;
@@ -37,17 +42,20 @@ public class CreateSolverCommand
     @XStreamAlias("solver-id")
     private String solverId;
 
-    @XmlElement(name = "solver-config-file")
-    @XStreamAlias("solver-config-file")
-    private String solverConfigFile;
+    @XmlElement
+    @XStreamAlias("problem-fact-change")
+    // It's not possible to use ProblemFactChange type here due to JAXB marshaller limitations
+    private Object problemFactChange;
 
-    public CreateSolverCommand() {
+    public AddProblemFactChangeCommand() {
     }
 
-    public CreateSolverCommand(String containerId, String solverId, String solverConfigFile) {
+    public AddProblemFactChangeCommand(String containerId,
+                                       String solverId,
+                                       ProblemFactChange problemFactChange) {
         this.containerId = containerId;
         this.solverId = solverId;
-        this.solverConfigFile = solverConfigFile;
+        this.problemFactChange = problemFactChange;
     }
 
     public String getContainerId() {
@@ -66,20 +74,20 @@ public class CreateSolverCommand
         this.solverId = solverId;
     }
 
-    public String getSolverConfigFile() {
-        return solverConfigFile;
+    public ProblemFactChange getProblemFactChange() {
+        return (ProblemFactChange) problemFactChange;
     }
 
-    public void setSolverConfigFile(String solverConfigFile) {
-        this.solverConfigFile = solverConfigFile;
+    public void setProblemFactChange(ProblemFactChange problemFactChange) {
+        this.problemFactChange = problemFactChange;
     }
 
     @Override
     public String toString() {
-        return "CreateSolverCommand{" +
-               "containerId='" + containerId + '\'' +
-               ", solverId='" + solverId + '\'' +
-               ", solverConfigFile='" + solverConfigFile + '\'' +
-               '}';
+        return "AddProblemFactChangeCommand{" +
+                "containerId='" + containerId + '\'' +
+                ", solverId='" + solverId + '\'' +
+                ", problemFactChange='" + problemFactChange + '\'' +
+                '}';
     }
 }
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
@@ -16,18 +16,23 @@
 
 package org.kie.server.api.commands.optaplanner;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import org.kie.server.api.model.KieServerCommand;
+import org.optaplanner.core.impl.solver.ProblemFactChange;
 
-@XmlRootElement(name = "terminate-solver-early")
-@XStreamAlias("terminate-solver-early")
+@XmlRootElement(name = "add-problem-fact-changes")
+@XStreamAlias("add-problem-fact-changes")
 @XmlAccessorType(XmlAccessType.NONE)
-public class TerminateSolverEarlyCommand
+public class AddProblemFactChangesCommand
         implements KieServerCommand {
 
     private static final long serialVersionUID = -1803374525440238478L;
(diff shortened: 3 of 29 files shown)


ge0ffrey:
Interesting prototype - sounds like a simple and clean API.
I'll review in depth next week when I am back from PTO.

Can we do a set of bulk problem fact changes as in Solver#addProblemFactChanges(List<...>)?

mcimbora:
Jenkins retest this.

mcimbora:
Jenkins retest this.

mcimbora:
Added support for submitting multiple problem fact changes & querying problem fact change processing status.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    @XStreamAlias("container-id")
+    private String containerId;
+
+    @XmlAttribute(name = "solver-id")
+    @XStreamAlias("solver-id")
+    private String solverId;
+
+    @XmlElement
+    @XStreamAlias("problem-fact-change")
+    private String problemFactChange;
```
ge0ffrey: Why is this of type String and not of type ProblemFactChange?

mcimbora: @ge0ffrey Changed that to type "Object". JAXB marshaller doesn't handle interfaces, unfortunately one cannot simply use ProblemFactChange type, as the actual object type is not known.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/CommandScript.java
```diff
@@ -72,7 +75,10 @@
             @XmlElement(name = "get-solvers", type = GetSolversCommand.class),
             @XmlElement(name = "get-solver", type = GetSolverCommand.class),
             @XmlElement(name = "start-solver", type = SolvePlanningProblemCommand.class),
-            @XmlElement(name = "terminate-solver", type = TerminateSolverEarlyCommand.class)
+            @XmlElement(name = "terminate-solver", type = TerminateSolverEarlyCommand.class),
+            @XmlElement(name = "add-problem-fact-change", type = AddProblemFactChangeCommand.class),
+            @XmlElement(name = "add-problem-fact-changes", type = AddProblemFactChangesCommand.class),
+            @XmlElement(name = "get-problem-fact-changes-processed", type = GetProblemFactChangesProcessedCommand.class)
```
ge0ffrey: I'd like the REST api to use the same terminology and even method names as the Solver API as much as reasonable.

The API method is Solver.isEveryProblemFactChangeProcessed(). Instead of

     @XmlElement(name = "get-problem-fact-changes-processed", type = GetProblemFactChangesProcessedCommand.class)

can it be


     @XmlElement(name = "is-every-problem-fact-change-processed", type = IsEveryProblemFactChangeProcessedCommand.class)

?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    @XStreamAlias("container-id")
+    private String containerId;
+
+    @XmlAttribute(name = "solver-id")
+    @XStreamAlias("solver-id")
+    private String solverId;
+
+    @XmlElement
+    @XStreamAlias("problem-fact-change")
+    private Object problemFactChange;
```
ge0ffrey: Can we make the type of this field ProblemFactChange instead of Object?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
```diff
@@ -0,0 +1,93 @@
...
+    @XStreamAlias("container-id")
+    private String containerId;
+
+    @XmlAttribute(name = "solver-id")
+    @XStreamAlias("solver-id")
+    private String solverId;
+
+    @XmlElement
+    @XStreamAlias("problem-fact-changes")
+    private List problemFactChanges;
```
ge0ffrey: Can we make the type of this field `List<ProblemFactChange>` instead of `List`?

We might even want to use `List<? extends ProblemFactChange>`? Not sure if that adheres to Josh Bloch's PECS principle (Producer Extends, Consumer Supers) - I always get a headache there.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public AddProblemFactChangesCommand() {
+        super();
```
ge0ffrey: no need to call super explicitly. Any construction implicitly always starts with a super() call (unless it's a this() call to another construction).

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/SolverServicesClient.java
```diff
@@ -41,6 +42,17 @@ void solvePlanningProblem(String containerId,
     void terminateSolverEarly(String containerId,
                               String solverId);
 
+    void addProblemFactChange(String containerId,
+                              String solverId,
+                              ProblemFactChange problemFactChange);
+
+    void addProblemFactChanges(String containerId,
+                              String solverId,
+                              List<ProblemFactChange> problemFactChange);
```
ge0ffrey: Here they are nicely typed to ProblemFactChange instead of Object. So I presume there's a reason for that - if it's not fixable in the Commands (I am intrested why in that case anyway), do add a comment that we except a ProblemFactChange instances there...

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/SolverServicesClient.java
```diff
@@ -41,6 +42,17 @@ void solvePlanningProblem(String containerId,
...
+
+    Boolean isEveryProblemFactChangeProcessed(String containerId,
+                                              String solverId);
```
ge0ffrey: When does this return null?

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/pom.xml
```diff
@@ -44,6 +44,13 @@
         <groupId>org.kie</groupId>
         <artifactId>kie-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
```
ge0ffrey: It's better to not do it like this, but like this instead:

```
  <properties>
    <maven.compiler.target>1.8</maven.compiler.target>
    <maven.compiler.source>1.8</maven.compiler.source>
  </properties>
```

See org.jboss:jboss-parent pom

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/filtered-resources/kjars-sources/cloudbalance/src/main/java/org/kie/server/testing/CloudBalancingGenerator.java
```diff
@@ -50,31 +51,73 @@ public int getCost() {
...
+                      300),
+            new Price(32,
+                      "32 gigabyte",
+                      400),
+            new Price(64,
+                      "64 gigabyte",
+                      600),
+            new Price(96,
+                      "96 gigabyte",
+                      1000),
```
ge0ffrey: My eyes are bleeding :) but ignore this comment, it's better to have a style that everyone follows than to bikeshed this further :)

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -447,4 +601,42 @@ public Object loadPlanningProblem(int computerListSize,
...
+            Object cloudComputer = method.invoke(cbgi,
+                                                 currentNumberOfComputers);
+
+            Class<?> cloudComputerClass = kieContainer.getClassLoader().loadClass(CLASS_CLOUD_COMPUTER);
+            Class<?> problemFactChangeClass = kieContainer.getClassLoader().loadClass(CLASS_ADD_COMPUTER_PROBLEM_FACT_CHANGE);
+
+            Constructor constructor = problemFactChangeClass.getConstructor(cloudComputerClass);
+            addProblemFactChange = (ProblemFactChange) constructor.newInstance(cloudComputer);
+        } catch (Exception e) {
+            e.printStackTrace();
```
ge0ffrey: printStrackeTrace() is bad because it's not logging friendly.

In this particular case, I 'd just rethrow the exception wrapped:

` throw new RuntimeException("Failed to create " + CLASS_ADD_COMPUTER_PROBLEM_FACT_CHANGE + " instance", e);`

But I am wondering why we're catching "Exception". No API should never throw Exception or Throwable and - except for real bottom-of-the-barrel exception handlers - nothing should ever catch Exception or Throwable (including JUnit tests in most cases).

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/CommandScript.java
mcimbora: @ge0ffrey Sounds like an improvement +1

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
mcimbora: @ge0ffrey Unfortunately not, JAXB marshaller is not able to handle interface types. This can be fixed by adding a converter, however that only works if your implementation set is known in advance (e.g. score impls). Won't work here unfortunately.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
mcimbora: @ge0ffrey I've got impression I tried the approach, however got the same issue with JAXB marshaller not being able to handle interfaces.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
mcimbora: @ge0ffrey +1, copied the class from elsewhere with this little issue. I'll do a clenup on existing commands.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
mcimbora: @ge0ffrey +1, I'll rework this & other generators to catch specific exception types.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangeCommand.java
ge0ffrey: Ah yes, that sucks. In SolvePlanningProblemCommand we workaround that problem differently, by having a String field.

1) Is there a good reason not to use the same workaround consistently? And if so, which is the better one?
2) Can we add a single comment line above it to explain it in all those workaround cases? Something like // Workaround: Field is not of type ProblemFactChange due to JAXB limitations

https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/SolvePlanningProblemCommand.java

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
ge0ffrey: Can we make it List<Object>? With the workaround comment.
I 'd rather avoid all use of non generic collections

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/SolverServicesClient.java
mcimbora: @ge0ffrey Yes, it doesn't look right to me as well. The problem is you can set different response handlers on the client. One of them org.kie.server.client.jms.FireAndForgetResponseHandler makes the client not wait for the server response. Instead the client just returns a null value.

mcimbora:
@ge0ffrey Thanks for the thorough review.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/SolverServicesClient.java
ge0ffrey: Ok, so we can't return boolean, we need to return Boolean due to jms limitations?
I can live with that.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/commands/optaplanner/AddProblemFactChangesCommand.java
ge0ffrey: Other question:
Why do we have a ...ChangesCommand and not just multiple ...ChangeCommands that all go into one add-problem-fact-changes REST call?

mcimbora:
@ge0ffrey The following approach looks promising https://github.com/kiegroup/droolsjbpm-integration/pull/1003/files#diff-831e9f3ce2bd9f25b8f7d587cb90f8e2R55. ProblemFactChange is internally stored as a Object type to make JAXB marshaller happy, however all AddProblemFactChange mutators use ProblemFactChange parameter.

mcimbora:
@ge0ffrey Similar approach is used for AddProblemFactChanges class, however a new list copy is required due to java generics type incompatibility.

ge0ffrey:
ok, sounds good.
Give me a signal when it's ok to merge again (no rush).

mcimbora:
@ge0ffrey Added tests for daemon mode too https://github.com/kiegroup/droolsjbpm-integration/pull/1003/files#diff-f7bca1e4d38f6dab2d958bfbbcbc9deaR288.

Can be merged once the build is green.

mcimbora:
Jenkins retest this.

mcimbora:
@ge0ffrey All green here, this can go in.

ge0ffrey:
@mcimbora Looks good, great work!
Can you document the new REST methods in kie-docs and make a bit of noise for it in the upgradeRecipe for the website?

yurloc:
@mcimbora Please don't forget to open an identical PR for the 7.2.x branch :-)

PR #1004, opened 2017-06-14
JBPM-6073 getComments method not returning desired page

annabaker:
During REST call, getComments was returning all comment results instead of only the desired page.  See JIRA here for reproduction and more info: https://issues.jboss.org/browse/JBPM-6073


cristianonicolai:
ok to test

cristianonicolai:
@annabaker if you compare with the other calls in `CaseServicesClientImpl.java` you will notice that this class is only a wrapper/router arounds jBPM service classes. I think the real issue is in the core service that is not paginating the results. Please check both `getCaseComments` methods in here: https://github.com/kiegroup/jbpm/blob/c385fdbc3821701d943002f3428f82d221ea70d4/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java#L454

annabaker:
@cristianonicolai hummm, I misunderstood what I was tracing from the workbench side. :/ Upon glancing at the methods you posted this makes more sense.  I am going to close this for now and create a new PR once I investigate and make the fix in the proper place.

PR #1006, opened 2017-06-14
Do not keep folders with Karaf containers in target folder.

jiripetrlik:
Folder with containers may grow quite fast in pax-exam folder (few gigabytes). It is not good to keep these folders in default build. This commit disables keeping these folders. It is possible to keep folders using karaf.keep.runtime.folder property. See more details in README in karaf-itests folder.


psiroky:
@jiripetrlik do we need the same for master? If so, could you please submit a PR?

PR #1008, opened 2017-06-15
Advanced search integration tests fixes

domhanak:
Contains fixes for issues revealed while testing different databases.

Makes tests send correlation key as string.
_Note: This seems to happen only on postgresql and other databases seem to handle correlation key even when sent as number, without 500 error being returned_

Updates tests using date in filter to use equalsTo rather than greaterThan
filter, since they tend to return more results and desired instance wouldn't get in the result.


domhanak:
jenkins retest this

domhanak:
jenkins retest this

domhanak:
jenkins retest this

PR #1009, opened 2017-06-15
JBPM 6073: getComments method not returning desired page

annabaker:
Modfication to base class for Case Management Service to use correct method call with page and page size parameters.  (in relation to https://github.com/kiegroup/jbpm/pull/883)


tsurdilo:
ok to test

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseManagementServiceBase.java
```diff
@@ -334,7 +334,7 @@ public void removeCommentFromCase(String containerId, String caseId, String comm
     }
 
     public CaseCommentList getComments(String containerId, String caseId, Integer page, Integer pageSize) {
-        Collection<CommentInstance> caseComments = caseService.getCaseComments(caseId, ConvertUtils.buildQueryContext(page, pageSize));
+        Collection<CommentInstance> caseComments = caseService.getCaseComments(caseId, page, pageSize, ConvertUtils.buildQueryContext(page, pageSize));
```
mswiderski: These page and pageSize is not needed here as same info is already set on he query context

PR #1010, opened 2017-06-16
[DROOLS-1582] add uf-maven into EAP modules

psiroky:


psiroky:

PR #1017, opened 2017-06-20
JBPM-6122: Invalid query test

sutaakar:
Test expects Kie server to returns HTTP status 500 when registering malformed query. Question is if Kie server can distinguish if error is caused by wrong input or some internal DB related fail. If so then it should return HTTP 400.


mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

PR #1020, opened 2017-06-20
JBPM-6125 - Move KIE Server on Tomcat to use Narayana as Transaction …

mswiderski:
…Manager instead of bitronix

depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/461

@sutaakar this completely replaces use of bitronix on kie server when running on tomcat. I didn't reconfigure local tests to use narayana at the moment it might be a good thing to do down the road.

@psiroky is there a way to test PR on tomcat for kie server? I ran most of the tests locally but would prefer to have a green light on PR that actually ran on tomcat :)


psiroky:
@mswiderski not right now I am afraid. Additional changes to the way the PRs builds work would be needed.

mswiderski:
jenkins build on tomcat 8 with narayana is green https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/master/job/narayana-droolsjbpm-integration/5/

sutaakar:
Looks good, just a question - have you tried local test runs? They were dependent on Bitronix too. They will probably need some adjustments.

mswiderski:
correct, I haven't changed to local ones yet as it is for testing only and not pushed to customer as official setup, but needs to be done

mswiderski:
@sutaakar looking at this a bit more it might not be as simple as it sounds mainly because narayana does not come with connection pooling on jndi provider so we would have to make way more config for it. So most likely we would still have to rely on bitronic for local tets.

Will merge this later today if no other comments

tsurdilo:
@mswiderski is this ok to be merged ?

PR #1022, opened 2017-06-20
JBPM-6073: getComments method not returning desired page

annabaker:
@tsurdilo  Not sure if this is overkill, but thought it might be safe to include an end to end test in part with https://github.com/kiegroup/jbpm/pull/885.


tsurdilo:
@annabaker not overkill. nice! thanks

PR #1031, opened 2017-06-25
JBPM-4480: Add Swagger to KIE-Server REST API

DuncanDoyle:
Initial implementation for JBPM-4480: https://issues.jboss.org/browse/JBPM-4480

Note that this PR requires the following PR to be merged first (as it contains the dependencyManagement elements for the Swagger dependencies we're using): https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/464


kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-swagger/src/main/java/org/kie/server/remote/rest/swagger/SwaggerRestApplicationComponentsService.java
```diff
@@ -0,0 +1,68 @@
...
+		// skip calls from other than owning extension
+        if ( !OWNER_EXTENSION.equals( extension )) {
+            return Collections.emptyList();
+        }
+        
+        BeanConfig beanConfig = new BeanConfig();
+        //TODO: Set the API version (or retrieve it from somewhere ...)
+        //beanConfig.setVersion("1.0.2");
+        String contextRoot = KieServerEnvironment.getContextRoot();
+        if (contextRoot != null && !("".equals(contextRoot))) {
```
mswiderski: wouldn't it be easier to use !contextRoot.isEmpty()?

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-swagger/src/main/java/org/kie/server/remote/rest/swagger/SwaggerRestApplicationComponentsService.java
```diff
@@ -0,0 +1,68 @@
...
+	@Override
+	public Collection<Object> getAppComponents(String extension, SupportedTransports type, Object... services) {
+		// skip calls from other than owning extension
+        if ( !OWNER_EXTENSION.equals( extension )) {
+            return Collections.emptyList();
+        }
+        
+        BeanConfig beanConfig = new BeanConfig();
+        //TODO: Set the API version (or retrieve it from somewhere ...)
+        //beanConfig.setVersion("1.0.2");
```
mswiderski: shouldn't we use product/project version for it?

kie-server-parent/kie-server-wars/kie-server/pom.xml
```diff
@@ -13,6 +13,24 @@
   <name>KIE :: Execution Server :: Wars :: Distribution Wars</name>
   <description>KIE Execution Server Distribution Wars. Name of the module is just 'kie-server' so that the final WARs have nicer names.</description>
 
+  <!-- TODO: Add to BOM or parent POM. -->
```
mswiderski: same here, can we move this out?

kie-server-parent/kie-server-wars/kie-server/pom.xml
```diff
@@ -124,6 +142,11 @@
     </dependency>
     <dependency>
       <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-services-swagger</artifactId>
+      <!-- TODO Exclusions? -->
```
mswiderski: are there any needed?

kie-server-parent/kie-server-wars/kie-server/src/main/swagger/index.html
```diff
@@ -0,0 +1,93 @@
+<!-- HTML for static distribution bundle build -->
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Swagger UI</title>
+  <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Source+Code+Pro:300,600|Titillium+Web:400,600,700" rel="stylesheet">
```
mswiderski: what about this behind firewall?

mswiderski:
@DuncanDoyle haven't we discussed that since it is runtime based thing it should properly react if the extensions are on or off?

I tried with disabled jbpm extension but it sill shows all the endpoints via the swagger ui (http://localhost:8230/kie-server/swagger/)

Did I misunderstand it?

Anyway it looks really nice so adding all descriptions and some data types to it will make it really nice thing!

DuncanDoyle:
@mswiderski  The UI in kie-server-war is simply copied from the Swagger UI distribution: http://swagger.io/swagger-ui/

When the extensions are off, the RESTful endpoint that serves the swagger spec should not be available, so the UI should probably state that it can't find that file.  So not sure why you still see all the endpoints to be honest (cache?).

kie-server-parent/kie-server-wars/kie-server/pom.xml
DuncanDoyle: Probably not :-) I think I put that there because I was using the swagger-jaxrs-jersey dependency at first, which brought in all sorts of jersey transitive dependencies .... I think we now simply depend on swagger-jars.

kie-server-parent/kie-server-wars/kie-server/pom.xml
DuncanDoyle: OK, missed that one. Will add those dependencies to the parent.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-swagger/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>kie-server-rest-swagger</artifactId>
+
+  <name>KIE :: Execution Server :: Remote :: REST :: Swagger</name>
+  <description>KIE Server Swagger Extension</description>
+
+  <dependencyManagement>
+  	<!-- TODO: Move to parent POM or BOM. -->
```
DuncanDoyle: Yes, forgot to do that. I'll fix that.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-swagger/src/main/java/org/kie/server/remote/rest/swagger/SwaggerRestApplicationComponentsService.java
DuncanDoyle: Yes, but I didn't know whether we already have that version available somewhere from which I can retrieve it .....

mswiderski:
@DuncanDoyle doesn't look like cache issue - I opened it in different browser - first in firefox and then in safari

DuncanDoyle:
@mswiderski I'll double check. Maybe it isn't properly disabled.

mswiderski:
when I try it out from the swagger ui it does respond with 404 so there is no jbpm extension enabled for sure.

Did this work for you? Maybe I didn't build all that is needed - I took bootstrap and integration repos only.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-swagger/src/main/java/org/kie/server/remote/rest/swagger/SwaggerRestApplicationComponentsService.java
DuncanDoyle: Old habbit. I've been writing checks like that for years :-) Can change to "isEmpty".

DuncanDoyle:
Ahhh, wait, I misunderstood that part. I thought you meant disabling the Swagger extension, but you're disabling the jBPM extension.

I think the problem is that the jBPM JAX-RS resources are, in that case, still on the classpath, so they will be picked up by the Swagger scanner ....

One possible fix would be to register the packages that Swagger should scan from the individual extensions, and then have Swagger only process those packages. In that case we would need to:
- define a place where we can store those package-names (KieServerEnvironment? Or another shared config object).
- Make sure the Swagger extension is always loaded last.

mswiderski:
yes, that's what I thought is the value of having this as kie server extension rather than build thingy. If we don't have option to have this to be hidden on runtime then it has much less value to be extension.

you can already take the active extensions from the context so that could be one way to do that too.

DuncanDoyle:
But even if we get the active extensions from the context, they should provide a way to retrieve the package-names that contain their JAX-RS packages.

Can we make this part of a second JIRA? I'll assign it to myself, and fix it before 7.0 product, but I need to give this some thought. In the meantime, I want this Swagger stuff in ASAP so we can give it some community love.

mswiderski:
@DuncanDoyle if you're sure this can be done - so it will remove endpoints from the UI if the extension is disabled then I am fine with it.

My main concern is if that can't be done then I don't see a need for extension, but more of making it build time activity.

Another thing I would like to have it done for the first commit - remove the /docs we have now and then use this instead. Would it be possible to map swagger ui to /docs so it will be sort of transparent to end users ... only improved

DuncanDoyle:
In that case, I'll look into it. Will take a bit of time though.

And yes, it would definitely be possible to map swagger-ui to docs. That's done at assemble time.

djeremiah:
I was actually working on this very same thing (swagger on the kie-server apis) locally in order to enable codegen of an angular2 client. 

One possibility for the dynamic enable/disable is below. I moved the BeanConfig from the getAppComponents call (which doesn't make a whole lot of sense to me) and moved it into the Extension init. I looped over all the app components and grabbed their packages. There's some fragility here. If someone introduces a package to the wrong extension, Swagger may still display resources that are currently disabled, but as long as we follow our current naming conventions, it'll be ok.

```
    @Override
    public void init(KieServerImpl kieServer, KieServerRegistry registry) {
    	this.kieServer = kieServer;
    	this.context = registry;

        BeanConfig beanConfig = new BeanConfig();
        //TODO: Set the API version (or retrieve it from somewhere ...)
        //beanConfig.setVersion("1.0.2");
        String contextRoot = KieServerEnvironment.getContextRoot();
        if (contextRoot != null && !contextRoot.isEmpty()) {
            beanConfig.setBasePath(contextRoot + "/services/rest");
        }
        beanConfig.setResourcePackage(getPackages(kieServer));
        beanConfig.setScan(true);
        beanConfig.setPrettyPrint(true);

    	initialized = true;
    }

    private String getPackages(KieServerImpl kieServer){
        return kieServer.getServerExtensions().stream()
                .map(extension-> extension.getAppComponents(SupportedTransports.REST))
                .flatMap(l -> l.stream())
                .map(component -> component.getClass().getPackage().getName())
                .distinct()
                .collect(Collectors.joining(","));
    }
```

DuncanDoyle:
Actually, I found that Swagger already includes a scanner which does this: JaxrsScanner. The trick is to not use BeanConfig as the scanner (BeanConfig implements the Scanner interface), but use the JaxrsScanner instead.

The advantage of that approach is that we're not scanning packages, which has a potential flaw. If an extension that's disabled contains @Api annotated resources in the same package as another extension (yes, I know, bad practice), these resources, which are disabled, are still added to the Swagger doc with the BeanConfig + resourcePackages approach. With the JaxrsScanner it works as expected.

One downside of my approach that I can only pass the scanner via the ServletContext parameters, as BaseApiListingResource does not allow to pass the id of the scanner we want to use (it does not pass anything to SwaggerContextService). But we can configure it via the ServletContext.

I'll update the PR if my tests are successful.

DuncanDoyle:
Configuring the Scanner to be used via the ServletConfig turned out to be problematic. Instead, I now set the DefaultJaxrsScanner as the scanner with the `SwaggerContextService.SCANNER_ID_DEFAULT` key. This makes sure that BeanConfig does not register itself as the default scanner. The DefaultJaxrsScanner only scans JAX-RS Resources that have defined on the JAX-RS Application, so only Resources from active extensions.

DuncanDoyle:
The old "jrapidoc" at "/kie-server/docs" has now been completely replaced by Swagger.

mswiderski:
Jenkins retest this

mswiderski:
Could you please remove this folder as it is not used any more https://github.com/kiegroup/droolsjbpm-integration/tree/master/kie-server-parent/kie-server-wars/kie-server/src/main/docs

DuncanDoyle:
I always forget something ..... :-) 'docs' folder removed!

mswiderski:
just tested it and it does filter out docs based on active extensions, excellent :)

the build fails as the revapi plugin complains that annotations were added, @MarianMacik can we do something about it globally to avoid failures after adding swagger annotations to our REST endpoints?

@DuncanDoyle does the swagger ui works slowly for you too? Overtime I want to expand given endpoint it takes around 2-30 sec before it expands...

mswiderski:
@DuncanDoyle oh and one more thing, looks like now to access docs page you need to logon, can we avoid this?

DuncanDoyle:
@mswiderski: I knew that one would come .... :-)

I already had a look at that (before you asked ;-)), but it requires URL whitelisting in web.xml, which is where you get into "ServletContainer specific implementation"-land. I can give it a go on EAP 7 (Undertow) and Tomcat, however, we would need QE to properly test this on WAS and WebLogic. I'll give it a go tonight.

MarianMacik:
@mswiderski I think that you should only add those 2 justifications to revapi-config.json file and it will be ok ;) The reason why we check annotations here is that we track backward compatibility of REST endpoints which are heavily configured by annotations.

mswiderski:
@MarianMacik I believe there will be more than just two as it will be per each *Resource class and in several modules. Soon when we add annotations for actual docs/descriptions that will be huge list of annotations. That's why I think we need to figure out if there is any better way to do that. like ignoring if annotations are from io.swagger packaged or similar, is that possible?

MarianMacik:
@mswiderski so we don't want to track these changes? If I understand it correctly Swagger will only generate something like documentation of our REST API, right? If that is the case we can ignore it, I will paste code in a minute.

mswiderski:
@MarianMacik correct, we should not track changes to swagger annotations as part of the revapi checks as it's not changing the api itself. It is only for documentation purpose.

MarianMacik:
So I think if you add this to ignores:

```
{
    "code": "java.annotation.added",
    "annotationType": "io.swagger.annotations.Api",
    "elementKind": "annotation",
    "justification": <<<<< ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE >>>>>
}
```
it should work.
Notice the omitted "old" and "new" JSON keys. I think revapi should take these two as wildcards, so it will ignore all cases when there is io.swagger.annotations.Api annotation added.

Filters (includes, excludes) can't be used for that since they are filtering based on methods, fields, classes - but e.g. we do not want to exclude some REST endpoint since then it would exclude all changes including JAX-RS annotations etc. I hope it is clear.

DuncanDoyle:
I found an issue with the ContextRoot which was longer properly set after moving init-code to the kie-server-services-swagger extension (it was set too late in the bootstrap process, so we're now setting it in the Bootstrap class).

Removed security from the Swagger resource (kie-server/services/rest/server/swagger.json). This approach works on both JBoss EAP 7/WildFly and Tomcat. This should really be properly tested on WAS and WebLogic as well.

mswiderski:
@DuncanDoyle here are the sections in swagger ui that hang browser completely:
- process-instance-search
- case-query
- task-search
- case-admin
- jbpm-forms

whenever I try to expand any of endpoints of above sections it hangs. Does that happen to you?

DuncanDoyle:
@mswiderski Yes, same here. Firefox warns about an unresponsive script: swagger-ui-bundle.js (different lines)

I would need to dive into this, but I'm not really a JavaScript guru. We could open a bug/issue against the swagger-ui ....

mswiderski:
@DuncanDoyle ok, thanks for confirming that. at least it's happening overall so should be easier to track down. 

+1 at opening jira for it against swagger-ui

mswiderski:
@DuncanDoyle any update on this?

DuncanDoyle:
@mswiderski Sorry for the delay. I upgraded to the latest version of Swagger-UI today and that seems to have fixed the issues we had. Also rebased the PR, and upgraded the new components to 8.0.0-SNAPSHOT (including the PR against droolsjbpm-build-bootstrap: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/464).

Can you re-test to see if this works for you as well?

mswiderski:
Thanks for letting me know, will take a look early next week

mswiderski:
Jenkins retest this

mswiderski:
works as expected now, thanks.

MarianMacik:
@mswiderski BTW does my advice work? https://github.com/kiegroup/droolsjbpm-integration/pull/1031#issuecomment-314077859

mswiderski:
That's the last thing I will check tomorrow

mswiderski:
jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
@DuncanDoyle finally green :)

@etirelli @krisv what do you guys think? I'd opt to include that as it enables to have always up to date docs based on enabled extensions. Further we will start actually provide full documentation over swagger api.

DuncanDoyle:
@mswiderski @etirelli @krisv Another advantage is that 3Scale supports Swagger files, so we get an easier integration with 3Scale API Management for free.

etirelli:
@DuncanDoyle @mswiderski I am all for bringing back swagger as long as we can confirm that it is not leaking memory anymore. The reason we had to remove swagger from the code a couple years ago was that it was blowing up memory consumption on the kie-server.

mswiderski:
the good thing here is that it is implemented as extension so it can be turned off

@etirelli anyway that's good to know, can you provide bit more details when did it manifest with high memory usage?

etirelli:
@mswiderski that happened I believe 2 years ago, so they must have fixed it by now. We just need to have an action item for QE to double check memory usage with this plugin enabled, but as long as we can disable it if we find a problem, we are good to merge it.

PR #1032, opened 2017-06-26
JBPM-6133 - Get Case comments from Kie Server with sort options

mswiderski:


kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/CaseServicesClient.java
```diff
@@ -118,6 +121,8 @@
 
     List<CaseComment> getComments(String containerId, String caseId, Integer page, Integer pageSize);
 
+    List<CaseComment> getComments(String containerId, String caseId, String sortBy, Integer page, Integer pageSize);
```
jakubschwan: Should be better to have getComments(String containerId, String caseId, Integer page, Integer pageSize, String sortBy, boolean sortOrder)

mswiderski: as you can see backend - in jbpm-case-mgmt does not support sort ordering since its very basic to just change the way objects in a list are placed. So instead of adding that in the backend I'd say client can then post process the list in the way they need. 
Whole point is that there is no storage that does this sorting on comments, it is simply in memory list manipulation ...

jakubschwan: Ok, thanks. This make sense.

PR #1034, opened 2017-06-27
kie-server-tests: Remove duplicated dependency, ignore test

sutaakar:


mswiderski:
jenkins retest this

domhanak:
jenkins retest this

domhanak:
Failure is related to bug with rounding miliseconds of the date and equalsTo filter than does not match the date because it is different by 002 milliseconds. Should be fixed by this #1036 and we should no longer have these random failures

mswiderski:
jenkins retest this

PR #1038, opened 2017-06-29
RHBPMS-4794 - History REST endpoints not behaving correctlly on WAS

mswiderski:
@krisv @markcoble I decided to relax the regex for the variable value on the rest endpoint as it causes issues on WAS but more importantly we don't limit/restrict on variable values on process level.


krisv:
@mswiderski last time a change in the url pattern also required a fix in the documentation as otherwise some junit test failed, I guess this might be required as well? See https://github.com/kiegroup/kie-wb-distributions/commit/1f3e8df42a4caed3dbb136cb2780c11577525e12

mswiderski:
created PR for it https://github.com/kiegroup/kie-wb-distributions/pull/544

thanks at pointing this out!

PR #1042, opened 2017-06-30
JBPM-6170 - Include kie server router and kie server standalone contr…

mswiderski:
…oller is kie server zip

@krisv @errantepiphany here it is what we talked about yesterday - adding standalone controller and router into the kie server zip


errantepiphany:
Awesome, thanks @mswiderski.

PR #1048, opened 2017-07-13
JBPM-6183 - Unhandled exception when advanced search api is used with…

mswiderski:
… Dates and JAXB on oracle, db2 and sybase

@domhanak would it be possible to confirm this fix on the failing data bases?


domhanak:
@mswiderski I will give it a try locally and let you know, but I am afraid that we do not have an efficient way to check this with multiple db's before merging.

mswiderski:
that's fine, one is enough to see if the date is then properly interpreted

domhanak:
@mswiderski looks good, war built with this PRQ is no longer producing the exception and tests are passing on oracle.

PR #1040, opened 2017-06-29
JBPM-6027: Advanced Search API shows error code 500 instead of 400

markcoble:

PR #1036, opened 2017-06-27
jbpm-search-tests: Do not use EqualsTo filters with dates

domhanak:
Removes equalsTo filters when querying with date and replaces them
with greaterThanOrEqualsTo to prevent random failures.
Modifies filters to use process instance id to minimize size
of the result returned.

PR #1057, opened 2017-07-19
JBPM-5457: Start and Stop container buttons change states even though they fail

karreiro:
This PR solves the issue [JBPM-5457](https://issues.jboss.org/browse/JBPM-5457) and the [RHBRMS-2874](https://issues.jboss.org/browse/RHBRMS-2874).

----
Part of an ensemble:
- https://github.com/kiegroup/droolsjbpm-integration/pull/1057
- https://github.com/kiegroup/kie-wb-common/pull/987

----

- https://issues.jboss.org/browse/JBPM-5457

Before:
before

After:
after

----

- https://issues.jboss.org/browse/RHBRMS-2874

Before:
before

After:
after
(my gif has a super low FPS rate, but when you’re using the screen, the containers are shown)

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
@@ -55,189 +55,282 @@ public class KieServerInstanceManager {
         return INSTANCE;
     }
 
-    public List<Container> startScanner(ServerTemplate serverTemplate, final ContainerSpec containerSpec, final long interval) {
+    public List<Container> startScanner(ServerTemplate serverTemplate,
+                                        final ContainerSpec containerSpec,
+                                        final long interval) {
 
-        return callRemoteKieServerOperation(serverTemplate, containerSpec, new RemoteKieServerOperation<Void>(){
-            @Override
-            public Void doOperation(KieServicesClient client, Container container) {
-                KieScannerResource scannerResource = new KieScannerResource();
-                scannerResource.setPollInterval(interval);
-                scannerResource.setStatus(KieScannerStatus.STARTED);
+        return callRemoteKieServerOperation(serverTemplate,
+                                            containerSpec,
+                                            new RemoteKieServerOperation<Void>() {
+                                                @Override
+                                                public Void doOperation(KieServicesClient client,
+                                                                        Container container) {
+                                                    KieScannerResource scannerResource = new KieScannerResource();
+                                                    scannerResource.setPollInterval(interval);
+                                                    scannerResource.setStatus(KieScannerStatus.STARTED);
 
-                ServiceResponse<KieScannerResource> response = client.updateScanner(containerSpec.getId(), scannerResource);
-                if (!response.getType().equals(ServiceResponse.ResponseType.SUCCESS)) {
-                    logger.debug("Scanner failed to start on server instance {} due to {}", container.getUrl(), response.getMsg());
-                }
-                collectContainerInfo(containerSpec, client, container);
+                                                    ServiceResponse<KieScannerResource> response = client.updateScanner(containerSpec.getId(),
+                                                                                                                        scannerResource);
+                                                    if (!response.getType().equals(ServiceResponse.ResponseType.SUCCESS)) {
+                                                        logger.debug("Scanner failed to start on server instance {} due to {}",
+                                                                     container.getUrl(),
+                                                                     response.getMsg());
+                                                    }
+                                                    collectContainerInfo(containerSpec,
+                                                                         client,
+                                                                         container);
 
-                return null;
-            }
-        });
+                                                    return null;
+                                                }
+                                            });
     }
 
-    public List<Container> stopScanner(ServerTemplate serverTemplate, final ContainerSpec containerSpec) {
+    public List<Container> stopScanner(ServerTemplate serverTemplate,
+                                       final ContainerSpec containerSpec) {
 
-        return callRemoteKieServerOperation(serverTemplate, containerSpec, new RemoteKieServerOperation<Void>(){
-            @Override
-            public Void doOperation(KieServicesClient client, Container container) {
-                KieScannerResource scannerResource = new KieScannerResource();
-                scannerResource.setPollInterval(null);
-                scannerResource.setStatus(KieScannerStatus.STOPPED);
+        return callRemoteKieServerOperation(serverTemplate,
+                                            containerSpec,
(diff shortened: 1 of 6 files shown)


kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java
```diff
@@ -184,175 +200,216 @@ public synchronized void deleteContainerSpec( String serverTemplateId,
...
-            throw new KieServerControllerNotFoundException( "No server template found for id " + containerSpecKey.getServerTemplateKey().getId() );
+        final ContainerSpec containerSpec = serverTemplate.getContainerSpec(containerSpecKey.getId());
+        if (containerSpec == null) {
+            throw new KieServerControllerNotFoundException("No container spec found for id " + containerSpecKey.getId() + " within server template with id " + serverTemplate.getId());
         }
 
-        ContainerSpec containerSpec = serverTemplate.getContainerSpec( containerSpecKey.getId() );
-        if ( containerSpec == null ) {
-            throw new KieServerControllerNotFoundException( "No container spec found for id " + containerSpecKey.getId() + " within server template with id " + serverTemplate.getId() );
+        final List<Container> containers = kieServerInstanceManager.stopContainer(serverTemplate,
```
karreiro: The real change is here.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
```diff
@@ -55,189 +55,282 @@ public static KieServerInstanceManager getInstance() {
...
-            }
-        });
-    }
+            public Void doOperation(KieServicesClient client,
+                                    Container container) {
 
+                final ServiceResponse<Void> response = client.disposeContainer(containerSpec.getId());
 
-    public List<Container> startContainer(ServerTemplate serverTemplate, final ContainerSpec containerSpec) {
+                if (response.getType().equals(ServiceResponse.ResponseType.SUCCESS)) {
```
karreiro: ...and here! ;-)

karreiro:
Jenkins please retest this.

sutaakar:
@karreiro Can you please avoid such big formatting changes? This makes the PR very hard to review. It can  also bring there unwanted changes and regressions as GitHub merger couldn't recognize formatting changes, overriding any modifications made by other PR in affected classes.

manstis:
@sutaakar The "formatting" is something QE have been pressing for "us" to do. @hasys initiated and mandated a new _standard_ formatter for all code. It is in ```droolsjbpm``` "[Developing Drools and jBPM](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/README.md)" guide. @triceo suggested the PR URL be appended with ```?w=1``` to remove noise.

An alternative is to format whole repositories in one swoop; which has indeed been performed on some.

Could you please explain your last comment "t can also bring there unwanted changes and regressions as GitHub merger couldn't recognize formatting changes, overriding any modifications made by other PR in affected classes" too. I don't see how _formatting_ could override other modifications. Github would report merge conflicts **if** formatting led to the inability to merge.

sutaakar:
@manstis For example see https://github.com/kiegroup/droolsjbpm-integration/pull/1040/files#diff-ab90edf589d98992b49fac48e330ca6bR76 . Test method was changed, instead of "testFindTaskInstanceWithSearchService(createQueryFilterGreaterThanOrEqualsTo..." it contains "testFindTaskInstanceWithSearchService( createQueryFilterEqualsTo..." . It passed unnoticed as the PR is very big and Mark mentioned that he changed some different tests. This PR then overrode changes introduced in https://github.com/kiegroup/droolsjbpm-integration/pull/1036 .
GitHub didn't report any merge conflict.

manstis:
@sutaakar Now you have me really confused. PR #1036 touches files: ```KieServerSynchronization.java```, ```JbpmQueriesKieServerBaseIntegrationTest.java```, ```ProcessSearchServiceIntegrationTest.java``` and ```TaskSearchServiceIntegrationTest.java```. NONE of these files are in this PR so I don't see how this PR overrides the changes made in #1036. Am I being completely thick?

sutaakar:
@manstis Sorry, I was talking about the case when a change presented as "formatting change" actually modified the code without anyone noticing it. It doesn't have direct relation to this PR.
I used it as an example that such big formatting change can potentially affect code functionality as these PRs are hard to be properly reviewed.

karreiro:
@mswiderski Could you please take a look? :-)

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/KieControllerStartupIntegrationTest.java
```diff
@@ -220,10 +220,9 @@ public void testContainerDisposedAfterStartup() throws Exception {
         assertEquals(1, instanceList.size());
 
         // Turn kie server off, dispose container and start kie server again.
-        server.stopKieServer();
```
sutaakar: This test verify that managed Kie server reflects container removal (made in controller) while being stopped during that removal. Please don't change order of the steps, that makes the tests useless.

mswiderski: +1 here, moreover that does not seem related to the actual change

mswiderski:
@karreiro in general this change the contract of what controller is suppose to do - it's a configuration management with option to broadcast change of the config to running servers. Though these updates might not be immediately accepted (as in this case) or servers can't be reachable. 

With current change, imagine situation where we have 10 servers behind single server template. So if you attempt to stop it controller with first send updates to each of them which will result in 10 updates to the server template - meaning 10 times it will update the state of the server template. This might not be of such a big issue if all the responses are same - either success or failure. But what if 9 of them succeeded but last one failed - that will then show in controller as started while it's not really the case.

That's why it was implemented to always first update the controller config regardless of what happened on the running servers. This is suppose to show the intention of the servers state rather than reflection of actual server instances. The real server instances are in the remote section of the execution servers perspective.

Another thing here is that we cannot easily distinguish what is the type of failure - is it because there are running process instances that disallow the container to be removed or is it a communication or authorisation issue which simply prevents from contacting with server.

mswiderski:
@karreiro as discussed on IRC this should be closed and alternative implementation is to follow

PR #1058, opened 2017-07-20
JBPM-6027: Advanced Search API shows error code 500 instead of 400

markcoble:
@domhanak tests are working now.


kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/rest/HttpStatusCodeException.java
```diff
@@ -0,0 +1,20 @@
+package org.kie.server.common.rest;
```
mswiderski: missing license header

markcoble: Thanks.  I knew I had forgotten it and then forgot to update!

PR #1060, opened 2017-07-21
JBPM-6230 - Kie server client load balancer test for URL with path

mswiderski:
@sutaakar here is your test case and the fix


sutaakar:
Looks good. One minor thing, you have TAB characters instead of spaces on several places.

mswiderski:
@sutaakar thanks, I think I fixed it

sutaakar:
@mswiderski Looks good now

PR #1071, opened 2017-07-26
RHBPMS-4790 - KIE Scanner not starting correctly when started using C…

lanceleverich:
…ontroller REST API

* Updated SpecManagementServiceImpl.updateContainerConfig(...) so that the update of a container's configuration (i.e. changing the scanner settings) results in the server being informed of the change; thus causing the kie-server to change the container's scanner also
* Updated SpecManagementServiceImpl.deleteContainerSpec(...) so that deleting the container would call the KieServerInstanceManager.stopContainer(...) method, which tells the kie-server to stop (undeploy) the container


lanceleverich:
Jenkins retest this

mswiderski:

PR #1079, opened 2017-07-28
JBPM-6271: Kie server custom extension REST test

sutaakar:
Basic implementation of Kie server custom extension. Covering usage of KieServerApplicationComponentsService and KieServerExtension.
Inspired by http://mswiderski.blogspot.cz/2015/12/kie-server-extend-existing-server.html

Reflection functionality used in tests separated into dedicated KieServerReflections class.

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
@@ -84,6 +84,7 @@ import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.ReleaseIdFilter;
 import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.api.model.ServiceResponsesList;
+import org.kie.server.api.model.Wrapped;
 import org.kie.server.api.model.admin.EmailNotification;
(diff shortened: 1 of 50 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/jaxb/JaxbMarshaller.java
```diff
@@ -392,12 +393,20 @@ public String marshall(Object input) {
     @Override
     public <T> T unmarshall(String input, Class<T> type) {
         try {
-            return (T) getUnmarshaller().unmarshal(new StringReader(input));
+            return (T) unwrap(getUnmarshaller().unmarshal(new StringReader(input)));
```
mswiderski: why was this needed here?

sutaakar: It is needed because test sends a list as a request body. On endpoint side the unmarshalled body for JAXB was wrapped, returning JaxbList. I added it there to avoid unwrapping.
Now the functionality is equal to JSON, where unwrapping in part of unmarshalling - https://github.com/kiegroup/droolsjbpm-integration/blob/50ce3c6481f61a16ad4ca0cbc27e091db09a4472/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java#L277

PR #1091, opened 2017-08-02
JBPM-6027: Advanced Search API shows error code 500 instead of 400

markcoble:


mswiderski:
@markcoble looks like the prefix [7.2.x] is on wrong PR

PR #1127, opened 2017-08-11
DROOLS-845 - @KSession throws exception when using kmodule in separate jar

jfneis:
KSession and KBase annotations were both assuming that a ReleaseId was provided, resulting in a NPE when only annotation value was present.

If ReleaseId is null just delegates to Spring Bean Factory to create the named bean.

### DIFF
--- a/kie-spring/src/main/java/org/kie/spring/annotations/AnnotationsPostProcessor.java
+++ b/kie-spring/src/main/java/org/kie/spring/annotations/AnnotationsPostProcessor.java
@@ -255,7 +255,7 @@ class AnnotationsPostProcessor implements InstantiationAwareBeanPostProcessor,
                 }
                 return kieContainer.getKieBase();
             }
-            if( getReleaseId().equals(AnnotationsPostProcessor.this.getReleaseId())) {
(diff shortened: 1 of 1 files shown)


mswiderski:
could you please add test case for this change?

jfneis:
@mswiderski it's my first PR to drools, could you provide me some guidance?

To test this issue we need to introduce another JAR, as it only happens when loading a module from a separate JAR. That's why the current tests weren't able to reproduce the problem, as all contexts in org/kie/spring/annotations declare modules/sessions load Kie* from the same JAR.

In this case, is it OK to add/commit a second JAR inside the test/resources folder? This second JAR project should be committed too? If yes (as I believe), under what directory structure?

Thank you!

mswiderski:
@jfneis there is always the first time - welcome and make sure it's not the last PR :) 

You can create kjar programatically, see this test class https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-spring/src/test/java/org/kie/spring/jbpm/services/AbstractJbpmServicesTest.java#L90

Hope this helps

jfneis:
All right, very nice piece of code! I'll work on it in the next days and update the PR! Thanks!

mswiderski:
No update for quite some time so closing. Feel free to reopen once test is added

PR #1128, opened 2017-08-14
RHBRMS-2784 - XStream: DoS when unmarshalling void type - leftovers

mbiarnes:


mswiderski:
@mbiarnes can this be merged or it does have to wait for other PRs?

PR #1130, opened 2017-08-15
JBPM-6085: Exclude jboss-jaxrs-api_2.0_spec

sutaakar:
Should fix the WAS 9 controller issue.


mswiderski:
based on my test that was not needed, though it's actually valid change as jax-rs is by default available in JEE containers. 
Do you by any chance run controller on WAS with parent last class loader? In case of controller that is not needed.

sutaakar:
@mswiderski Yes, I run it with parent last class loader. It is a default test configuration as it is needed by Kie server.

mswiderski:
ok, that explains it then, thanks @sutaakar

PR #1137, opened 2017-08-17
JBPM-6351: Test and fix for update to broken container

sutaakar:
@mswiderski What should happen if there is an error while container is updated - like in the test? Should it be reverted back to the original version?


mswiderski:
Kie container should be back to the version it was before. As the upgrade did not succeed

sutaakar:
Failures caused by DROOLS-1704, reproduced in https://github.com/kiegroup/droolsjbpm-integration/pull/1139

mswiderski:
Jenkins retest this

PR #1148, opened 2017-08-22
[7.3.x] JBPM-6336 - KIE Server Router support for OpenShift with services

mswiderski:
@tsurdilo @sutaakar here is a bacport of #1145


mswiderski:
Jenkins retest this

PR #1154, opened 2017-08-26
RHBPMS-4876 KieServerMDB retains jms connection retrieved from manage…

lanceleverich:
…d connection pool...

* Added private method startConnectionAndSession
* Added private method closeConnectionAndSession
* Removed code that did connection/session create from the init method
* Removed code the closed connection/session from the cleanup method
* Added calls to the startConnectionAndSession and closeConnectionAndSession methods in the onMessage.
This means that connections and sessions will be created/closed within the single onMessage

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/KieServerMDB.java
+++ b/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/KieServerMDB.java
@@ -70,6 +70,7 @@ public class KieServerMDB
 
     // Constants / properties
     private              String RESPONSE_QUEUE_NAME          = null;
+    private static final String RESPONSE_QUEUE_NAME_PROPERTY = "kie.server.jms.queues.response";
     private static final String DEFAULT_RESPONSE_QUEUE_NAME  = "queue/KIE.SERVER.RESPONSE";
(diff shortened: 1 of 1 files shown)


lanceleverich:
Jenkins retest please

mdproctor:
+1 ACK

sutaakar:
@lanceleverich According to [1] the Connection is heavyweight object. I think it should be initialized just once for MDB lifetime, otherwise it may have negative performance impact. What do you think?

[1] https://docs.oracle.com/javaee/7/api/javax/jms/Connection.html

lanceleverich:
Some additional information...
1) We are using pooled connections, which if properly configured should
mean that we are not actually establishing new connections each time.
2) This concern was raised when I received the initial patch file from CEE.
Mark and I were told that because of point #1, there should be little  (if
any) impact on performance.
3) This patch had a very short turn around time, and did not follow normal
process/protocol, due to the client and the deadline that the CEE was
working under.

Normally, I would not have suggested, nor been in favor of the way we are
now managing the connections/sessions. And there is still an outstanding
question of whether this fixes an actual problem, or is merely masking some
deeper issue(s). Consequently, we will continue to try to find a way to
reproduce the original issue, so that we can verify that this patch does
fix the actual reported problem.

~ Lance

PR #1156, opened 2017-08-28
[6.5.x] kie-server-tests: unify container profile ids

psiroky:
Most of the other repositories use the IDs without the "x" suffix.
Those are the chosen cannonical names that should be used everywhere.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
@@ -190,6 +190,20 @@
       <type>war</type>
       <scope>test</scope>
     </dependency>
+
+    <!-- This is an artificial dependency to make sure the kie-server-tests modules are executed one at a time during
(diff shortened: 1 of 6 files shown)


psiroky:
Not sure if the fails are related to my changes or not since these changes effectively just enable the tests (they were disabled before so could have been failing for some time). @sutaakar any hints?

sutaakar:
@psiroky It is because Kie server integration tests on this branch don't contain artificial dependency on previous module, causing maven to run them concurrently.

psiroky:
@sutaakar ah, right, I forgot about that. So I guess we should add that artificial dep then? The other option is to just use the serial build.

sutaakar:
@psiroky I guess the artificial dependency would be the easiest option.

sutaakar:
@psiroky Please update also eap64x in kie-server-integ-tests-jbpm

psiroky:
Should be fixed now. Let's see what Jenkins thinks.

PR #1159, opened 2017-08-28
RHBPMS-4876 [7.3.x] KieServerMDB retains jms connection retrieved fro…

lanceleverich:
…m managed connection pool...

* Added private method startConnectionAndSession
* Added private method closeConnectionAndSession
* Removed code that did connection/session create from the init method
* Removed code the closed connection/session from the cleanup method
* Added calls to the startConnectionAndSession and closeConnectionAndSession methods in the onMessage.
This means that connections and sessions will be created/closed within the


mswiderski:
@lanceleverich could you please prepare single PR (maybe update this one) with all changes to KieServerMDB that were done so we don't have this issue popping up again anytime soon? thanks

lanceleverich:
This PR now includes all changes that have been made in connection with RHBPMS-4876, including the two "fixes" that were made after the initial patch.

PR #1164, opened 2017-09-07
JBPM-6384 - Smart router synchronization with Workbench, JBPM-6400 - …

mswiderski:
…KIE Server router - support for job operations

### DIFF
--- a/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/handlers/JobsHttpHandler.java
+++ b/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/handlers/JobsHttpHandler.java
@@ -16,11 +16,32 @@
 package org.kie.server.router.handlers;
 
 import io.undertow.server.HttpHandler;
+import io.undertow.server.HttpServerExchange;
+import io.undertow.util.HttpString;
 
 public class JobsHttpHandler extends AbstractAggregateHttpHandler {
+    
+    private static final String PREFIX = "/jobs";
 
     public JobsHttpHandler(HttpHandler httpHandler, AdminHttpHandler adminHandler) {
         super(httpHandler, adminHandler);
+    }
+
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        if (!exchange.getRequestMethod().equals(HttpString.tryFromString("GET"))) {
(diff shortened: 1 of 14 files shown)


mswiderski:
Jenkins retest this

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/handlers/JobsHttpHandler.java
```diff
@@ -16,11 +16,32 @@
...
+    private static final String PREFIX = "/jobs";
 
     public JobsHttpHandler(HttpHandler httpHandler, AdminHttpHandler adminHandler) {
         super(httpHandler, adminHandler);
     }
 
+    @Override
+    public void handleRequest(HttpServerExchange exchange) throws Exception {
+        if (!exchange.getRequestMethod().equals(HttpString.tryFromString("GET"))) {
+            exchange.setRelativePath(PREFIX + exchange.getRelativePath());
```
sutaakar: Shouldn't we restore here full request path of the exchange? To be something like "exchange.setRelativePath(exchange.getRequestPath());" ?

mswiderski: this is required as undertow removed the prefix when it matched any of the path handlers. So essentially this is to put it back so when forwarding this exchange to proxyhandler it will have the full path where the request should be sent.

sutaakar: Yes, I just meant that by using "exchange.setRelativePath(exchange.getRequestPath());" instead of "exchange.setRelativePath(PREFIX + exchange.getRelativePath());" we can simplify the code.

mswiderski: main reason why I did it this way was to simply reverse what was removed and still keep the remaining part. Not sure if the RequestPath will not bring in anything else that might have impact on further proxy execution

sutaakar: ok, no problem. It was just an idea.

PR #1167, opened 2017-09-07
kie-performance-kit: Raise warmUp logging level from debug to info 

MarianMacik:
Same as #1165.


MarianMacik:
jenkins retest this

PR #1173, opened 2017-09-08
RHBPMS-4917 - Old version of container is being deployed after previo…

mswiderski:
…us deployment failed even though a new one is already built


mswiderski:
Jenkins retest this

PR #1176, opened 2017-09-08
RHBPMS-4876_6.5.x_fix

lanceleverich:
* Created JMSConnection class to hold values for the Connection and Session
* Updated KieServerMDB
- Removed instance variables for Connection and Session
- Changed startConnectionAndSession method to return a JMSConnection
- Changed closeConnectionAndSession method to take a JMSConnection argument
- Changed onMessage method to use the JMSConnection returned from the startConnectionAndSession method
- Changed sendResponse method to take a Session argument

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/JMSConnection.java
+++ b/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/JMSConnection.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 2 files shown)


bradsdavis:
@lanceleverich on this source -- onMessage seems to establish a JMS Session.  I believe that this is used for the correlation message to be sent as a response.  However, we are calling this with FireAndForgetResponseHandler.  

If we are using FireAndForgetResponseHandler, creating a connection and session seems unnecessary; however, I am not completely following the source here.  Is this indeed an issue?

mswiderski:
@lanceleverich +1 on @bradsdavis comment, connection, session etc should only be created when reply is going to be send

@mdproctor why do we rush so much with constant changes in kie server without a chance to comment/review by others? No test added does not makes it any better to understand what this change suppose fix/improve

PR #1177, opened 2017-09-10
RHBPMS-4876_6.5.x_fix2

lanceleverich:
* Moved the creation of the JMSConnection to only occur if the interactionPattern is not FIRE_AND_FORGET
* Updated the finally statement for the onMessage method so that it won't attempt to close non-existent connections


lanceleverich:
Jenkins retest this please

mswiderski:
requested by https://issues.jboss.org/browse/RHBPMS-4938

PR #1178, opened 2017-09-10
RHBPMS-4876_fix2

lanceleverich:
* Moved the creation of the JMSConnection to only occur if the interactionPattern is not FIRE_AND_FORGET
* Updated the finally statement for the onMessage method so that it won't attempt to close non-existent connections

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/KieServerMDB.java
+++ b/kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/KieServerMDB.java
@@ -197,7 +197,7 @@ public class KieServerMDB
     }
 
     public void onMessage(Message message) {
-        JMSConnection connect = startConnectionAndSession();
+        JMSConnection connect = null;
         try {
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-remote/kie-server-jms/src/main/java/org/kie/server/jms/KieServerMDB.java
```diff
@@ -313,11 +314,15 @@ public void onMessage(Message message) {
             }
 
         } finally {
-            try {
-                closeConnectionAndSession(connect);
-            } catch (JMSRuntimeException runtimeException) {
-                logger.error("Error while attempting to close connection/session",runtimeException);
-            } finally {
+            if (connect != null) { // Only attempt to close the connection/session if they were actually created
```
mdproctor: I'd probaby have done a debug log if conncetion == null. that included the "interactionPattern" and  "UPPER_LIMIT_REPLY_INTERACTION_PATTERNS" values. So in debug we know the onMessage was FireAndForge

PR #1192, opened 2017-09-18
[7.3.x] RHBPMS-4943: New ad-hoc job can't be created

nmirasch:
Adding check over the containerId to add it as a parameter, allowing ad-hoc creation

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
@@ -55,10 +55,13 @@ public class JobServicesClientImpl extends AbstractKieServicesClientImpl impleme
         if( config.isRest() ) {
 
             Map<String, Object> valuesMap = new HashMap<String, Object>();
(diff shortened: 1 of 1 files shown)


nmirasch:
@mswiderski could you take a look? thanks!

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/JobServicesClientImpl.java
```diff
@@ -55,10 +55,14 @@ public Long scheduleRequest(String containerId, JobRequestInstance jobRequest) {
         if( config.isRest() ) {
 
             Map<String, Object> valuesMap = new HashMap<String, Object>();
-            valuesMap.put(CONTAINER_ID, containerId);
+            String queryString = "";
+            if (containerId != null && !containerId.isEmpty()) {
+                valuesMap.put(CONTAINER_ID, containerId);
```
mswiderski: this is not needed, containerId does not have to be added to the values map sit is not a path param but query param only.

nmirasch:
yes, it was what I thought. Only wanted add the check and not to change to much the code. Change done!

PR #1195, opened 2017-09-18
RHBRMS-2953: Cannot start container on Kie Server if it contains decision table with org.optaplanner.workbench.screens.guidedrule.model.ActionSimpleConstraintMatch

mcimbora:
@yurloc Can you please take a look? Adding the module, which contains custom planner IAction implementations. Thanks in advance.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/pom.xml
@@ -46,6 +46,31 @@
       <groupId>org.optaplanner</groupId>
       <artifactId>optaplanner-core</artifactId>
     </dependency>
+
+    <dependency>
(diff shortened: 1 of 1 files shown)


yurloc:
I have verified that the reported issue is fixed by this change. Todo:
- [ ] Improve commit message
- [ ] Resolve enforce fail due to duplicate classes from `xml-apis:xml-apis:jar` and `javax.xml.stream:stax-api`

mcimbora:
@psiroky @yurloc As discussed on IRC, we need a new module in kiegroup/optaplanner which will include workbench APIs (something similar as drools-workbench-models).

mcimbora:
Closing, will be resolved as a part of https://github.com/kiegroup/droolsjbpm-integration/pull/1225.

PR #1199, opened 2017-09-19
JBPM-6396 - KIE Server extension for optaplanner-jbpm human task opti…

mswiderski:
…mization

@ge0ffrey @mcimbora @krisv 
guys, here is the initial implementation on how we could support the real time planning that we discussed a while ago. I planned to use new kie server extension but then realised it won't have too much value over what I did now. So the main thing here is ContinuousPlanning interface that is the integration point for all use case we could implement for planning - one of them (or reference one) will be continues task assignment. 

So it can be either implemented as global solution placed on kie server class path or can be implemented on kjar level and thus scoped to kjar. Planning will be done based on kjar anyway so the ContinuousPlanning provides all the information required - as far as I could figure out :)

So next step will be to implement jbpm/user task specific ContinuousPlanning that will provide the planning entities and planning problem impl so it can actually be solved.

How does this look like to you?

P.S.
@ge0ffrey @mcimbora I will need your help with next steps - to define planning problem, entities and everything else needed.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
@@ -16,6 +16,7 @@
 package org.kie.server.services.optaplanner;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
@@ -59,6 +60,10 @@ public class OptaplannerKieServerExtension
     private List<Object> services = new ArrayList<Object>();
     private boolean initialized = false;
     private OptaplannerCommandServiceImpl optaplannerCommandService;
+    
+    private List<RealtimePlanning> globalContinuousPlannings;
+    
+    private Map<String, List<RealtimePlanning>> containerContinuousPlannings = new HashMap<>();
 
     @Override
     public boolean isInitialized() {
@@ -96,6 +101,8 @@ public class OptaplannerKieServerExtension
         this.optaplannerCommandService = new OptaplannerCommandServiceImpl(registry, solverServiceBase);
 
         this.services.add( solverServiceBase );
+        
+        this.globalContinuousPlannings = loadPlannings(this.getClass().getClassLoader());
 
         initialized = true;
     }
@@ -105,10 +112,23 @@ public class OptaplannerKieServerExtension
         if( this.threadPool != null ) {
             this.threadPool.shutdownNow();
         }
+        
+        for (RealtimePlanning planning : globalContinuousPlannings) {
+            try {
+                planning.close();
+                logger.debug("ContinuousPlanning {} closed successfully", planning);
+            } catch (Exception e) {
+                logger.error("Unexpected error thrown when closing {}", planning, e);
+            }
+        }
     }
 
     @Override
     public void createContainer(String id, KieContainerInstance kieContainerInstance, Map<String, Object> parameters) {
+        
+        List<RealtimePlanning> containerPlannings = loadPlannings(kieContainerInstance.getKieContainer().getClassLoader());
+        containerContinuousPlannings.put(id, containerPlannings);
+        solverServiceBase.boostrapContinuousPlanning(id, containerPlannings);        
     }
 
     @Override
@@ -124,6 +144,16 @@ public class OptaplannerKieServerExtension
     @Override
     public void disposeContainer(String id, KieContainerInstance kieContainerInstance, Map<String, Object> parameters) {
         solverServiceBase.disposeSolversForContainer( id, kieContainerInstance );
+        
+        List<RealtimePlanning> containerPlannings = containerContinuousPlannings.getOrDefault(id, new ArrayList<>());
+        for (RealtimePlanning planning : containerPlannings) {
+            try {
+                planning.close();
+                logger.debug("Container's {} ContinuousPlanning {} closed successfully", id, planning);
+            } catch (Exception e) {
+                logger.error("Unexpected error thrown when closing {} for container {}", planning, id, e);
+            }
+        }
     }
 
     @Override
@@ -174,4 +204,17 @@ public class OptaplannerKieServerExtension
         return EXTENSION_NAME + " KIE Server extension";
     }
 
+    
+    protected List<RealtimePlanning> loadPlannings(ClassLoader classLoader) {
+        List<RealtimePlanning> plannings = new ArrayList<>();
+        ServiceLoader<RealtimePlanning> discoveredPlannings = ServiceLoader.load(RealtimePlanning.class, classLoader);
+        
+        for (RealtimePlanning planning : discoveredPlannings) {
+            if (planning.init()) {
+                plannings.add(planning);
+            }
+        }
+        
+        return plannings;
+    }
 }
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.services.optaplanner;
+
+import java.util.List;
+
+import org.optaplanner.core.api.solver.event.SolverEventListener;
+import org.optaplanner.core.impl.solver.ProblemFactChange;
+
+/**
+ * Provides integration for OptaPlanner to solve planning problems real time 
+ * based on support of adding problem facts.
+ *
+ */
+public interface RealtimePlanning {
+
+    /**
+     * Invoked at the initialization phase of OptaPlanner Extension to identify planning that are active.
+     * @return returns true if this planning was successfully activated, false otherwise. In case of false returned 
+     * this instance of planning will not be called anymore.
+     */
+    boolean init();
+    
+    /**
+     * Invoked when OptaPlanner Extension is being shutdown.
+     */
+    void close();
+    
+    /**
+     * Returns expected solver id to be used when initiating.
+     * @return solver id to be used
+     */
+    String getSolverId();
+    
+    /**
+     * Returns solver config path to be used when bootstraping
+     * @return solver config path
+     */
+    String getSolverConfigPath();
+    
+    /**
+     * Asserts if given containerId is accepted by this planning.
+     * @param containerId id of the container that is being initiated
+     * @return true if given container id is accepted otherwise false
+     */
+    boolean accept(String containerId);
+    
+    /**
+     * Returns callback that should be used by solver when best solution was found.
+     * @return actual implementation of the event listener
+     */
+    SolverEventListener<Object> getCallback();
+    
+    /**
+     * Invoked upon solver start to load initial set of problem facts to be solved.
+     * @return non null list of problem facts
+     */
+    List<ProblemFactChange<Object>> loadFacts();
+    
+    /**
+     * Returns planning problem to start solving
+     * @return actual planning problem to be solved
+     */
+    Object getPlanningProblem();
+}
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
@@ -33,6 +33,7 @@ import org.kie.server.services.impl.KieContainerInstanceImpl;
 import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.api.solver.event.SolverEventListener;
 import org.optaplanner.core.impl.solver.ProblemFactChange;
 import org.slf4j.Logger;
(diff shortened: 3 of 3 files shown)


kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
```diff
@@ -105,10 +112,28 @@ public void destroy(KieServerImpl kieServer, KieServerRegistry registry) {
...
     @Override
     public void createContainer(String id, KieContainerInstance kieContainerInstance, Map<String, Object> parameters) {
+        
+        List<ContinuousPlanning> containerPlannings = loadPlannings(kieContainerInstance.getKieContainer().getClassLoader());
+        containerContinuousPlannings.put(id, containerPlannings);
+        // first try with kjars defined plannings
+        boolean found = solverServiceBase.boostrapContinuousPlanning(id, containerPlannings);
+        if (!found) {
+            // if not found then use global ones
+            solverServiceBase.boostrapContinuousPlanning(id, globalContinuousPlannings);
```
MarianMacik: Just to be sure - so global ones will always return true for their accept method, right?

mswiderski: no, but it does not matter anymore as there is nothing more we can do so it's essentially done regardless if it found anything or not.

After I did this I think that global are actually not needed, reason for that is ProjectClassLoader has kie server app class loader as parent anyway so that should discover globals too.

MarianMacik: OK, but what is the difference between container and global continuous plannings? I mean container plannings are connected with a particular container - therefore they are in a map. However, global ones are not connected with any container so why are they loaded using the same method when they are not connected to any container? But maybe I am just missing something here...

MarianMacik: OK, I might understand it now - so the only difference is the classloader but still it is needed for each planning to have its accepted container on which it can run, right? :)

mswiderski: the reason here was to have single method that can load and init planning from different class loaders. globals were for these that are on kie server class path while the containers are for those that come with kjar so they need to be additionally tracked so can be closed the kjar is undeployed

MarianMacik: ok clear now. So if there aren't any container-specific plannings for the container being deployed then with each container deployment the global planning (if it is accepted for the container being deployed, let's say it always returns true) will be started, right? So theoretically if I set up it this way it will act as a fallback for containers that don't have their own container-specific planning.

mswiderski: correct

mcimbora:
@mswiderski Maciej, thanks for the initial implementation. I'll try to build a POC around the interface to identify potential pain points.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
rsynek: @mswiderski is it possible that due to the fact that "ProjectClassLoader has kie server app class loader as parent" might be some "global" planning loaded twice? Firstly as global planning, secondly as container-specific?

mswiderski: @rsynek yes, this is what I mentioned in the comment above that ProjectCLassLoader will find the global planning too. So most likely there is no need for them (global plannings) to be loaded separately. and yes, that will result in multiple loads in case there are several containers and planning on kie server class path.

rsynek: @mswiderski my concern is whether it will behave as you described - it seems that container-specific is not overriding the global planning, as in fact it's going to load both due to the classloader hierarchy.

rsynek: Ok, thanks for explanation

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/ContinuousPlanning.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.optaplanner.core.api.solver.event.SolverEventListener;
+import org.optaplanner.core.impl.solver.ProblemFactChange;
+
+/**
+ * Provides integration for OptaPlanner to continuously solve planning problems 
+ * based on support of adding problem facts.
+ *
+ */
+public interface ContinuousPlanning {
```
mcimbora: I think we should consider using term 'Real-time planning' here. We're aiming for a solving process, which will run without interruptions (as opposed to a regular task). I have to admit the terms are a bit confusing and the difference may not be obvious at first sight. I'll let @ge0ffrey add more insight.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -515,4 +516,41 @@ private void terminateSolverEarly(SolverInstanceContext sic) {
...
     }
+    
+    protected boolean boostrapContinuousPlanning(String containerId, List<ContinuousPlanning> plannings) {
+        for (ContinuousPlanning planning : plannings) {
+            if (planning.accept(containerId)) {                
+                logger.info("ContinuousPlanning {} is going to be used for container {}", planning, containerId);
+                
+                String solverId = planning.getSolverId();
+                SolverInstance instance = new SolverInstance();
+                instance.setSolverConfigFile(solverId + "-solver.xml");
```
mcimbora: I do believe we need a new method String getSolverConfigPath() in ContinousPlanning interface to return a relative path to a config file. This approach imposes certain naming format, which should be avoided. The Solver API that we have ATM allows setting the solver config path separately from the solverId used.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -515,4 +516,41 @@ private void terminateSolverEarly(SolverInstanceContext sic) {
...
+                instance.setContainerId(containerId);
+                instance.setSolverId(solverId);
+                
+                logger.debug("About to start solver {}", instance);
+                createSolver(containerId, solverId, instance);
+                
+                logger.debug("Solver with id {} created successfully", solverId);
+                SolverInstanceContext solverInstance = solvers.get(solverId);
```
mcimbora: This should use solvers.get(SolverInstance.getSolverInstanceKey(containerId, solverId)).

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -515,4 +516,41 @@ private void terminateSolverEarly(SolverInstanceContext sic) {
...
+                solver.addEventListener(eventListener);
+                logger.debug("Added solver event listner {} to solver {}", eventListener, solver);
+                
+                List<ProblemFactChange<Object>> problemFacts = planning.loadFacts();
+                logger.debug("About to add following problem fact changes {} to solver {}", problemFacts, solverId);
+                if (!problemFacts.isEmpty()) {
+                    solver.addProblemFactChanges(problemFacts);
+                }
+                
+                solver.solve(planning.getPlanningProblem());
```
mcimbora: Let's use SolverServiceBase.executor for boostrapContinuousPlanning() to avoid client timeouts. See SolverServiceBase.solvePlanningProblem for a reference.

mswiderski:
@mcimbora changes applied as requested

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.optaplanner.core.api.solver.event.SolverEventListener;
+import org.optaplanner.core.impl.solver.ProblemFactChange;
+
+/**
+ * Provides integration for OptaPlanner to solve planning problems real time 
+ * based on support of adding problem facts.
+ *
+ */
+public interface RealtimePlanning {
```
ge0ffrey: This interface needs to be renamed.
Suggestions: `SolverBootstrap`, `SolverProvisioning`, `SolverStartup` or ...

The concept of "real time planning" is orthogonal (~independent) to what this interface is, namely something to bootstrap a solver upon starting up the server.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Invoked at the initialization phase of OptaPlanner Extension to identify planning that are active.
+     * @return returns true if this planning was successfully activated, false otherwise. In case of false returned 
+     * this instance of planning will not be called anymore.
+     */
+    boolean init();
```
ge0ffrey: If something fails, it will probably always thrown a RuntimeException (in the fail fast philosophy).
Why would we ever want to return false? I soft suggest to make method this return void.

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+     * Returns callback that should be used by solver when best solution was found.
+     * @return actual implementation of the event listener
+     */
+    SolverEventListener<Object> getCallback();
+    
+    /**
+     * Invoked upon solver start to load initial set of problem facts to be solved.
+     * @return non null list of problem facts
+     */
+    List<ProblemFactChange<Object>> loadFacts();
```
ge0ffrey: Rename to loadInitialProblemFactChanges. I suspect more will follow later at runtime?

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+    
+    /**
+     * Returns planning problem to start solving
+     * @return actual planning problem to be solved
+     */
+    Object getPlanningProblem();
```
ge0ffrey: To be consistent with the rest of optaplanner-core, we 'd probably want to use a generic type "Solution_" for this entire class. Then 3 methods can use it:
- `Solution_ getPlanningProblem()`
- `SolverEventListener<Solution_>`
- `List<ProblemFactChange<Solution_>>`

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return solver config path
+     */
+    String getSolverConfigPath();
+    
+    /**
+     * Asserts if given containerId is accepted by this planning.
+     * @param containerId id of the container that is being initiated
+     * @return true if given container id is accepted otherwise false
+     */
+    boolean accept(String containerId);
```
ge0ffrey: Can you give an example use case where this method would be useful?

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+    
+    /**
+     * Returns callback that should be used by solver when best solution was found.
+     * @return actual implementation of the event listener
+     */
+    SolverEventListener<Object> getCallback();
```
ge0ffrey: There might be additional callbacks at some point. Let's rename this method to getSolverEventListener()?

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
```diff
@@ -0,0 +1,79 @@
...
+     * Returns expected solver id to be used when initiating.
+     * @return solver id to be used
+     */
+    String getSolverId();
+    
+    /**
+     * Returns solver config path to be used when bootstraping
+     * @return solver config path
+     */
+    String getSolverConfigPath();
```
ge0ffrey: Would it be more flexible if we simply have a `buildSolver()` method? That would allow us to get rid of:
- getSolverConfigPath()
- getCallback()

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/RealtimePlanning.java
mswiderski: when you have many implementations of this interface you don't want it to be loaded for all kjars in kie server. Or if you have other kjars (not optaplanner) on the server there is no need to load stuff or bootstrap solvers etc.

mswiderski:
@mcimbora @ge0ffrey looks like there is way too many optaplanner related stuff for me to carry on, so please either start fresh on top of these ideas or continue from this interface. It does not make much sense more me to simply go and apply your suggestions as I don't have enough knowledge about optaplanner and the way you want it to evolve.

ge0ffrey:
@mswiderski I've asked @yurloc to take a look at reviving this PR (probably as a new PR) and continue this story.

PR #1200, opened 2017-09-19
RHBPMS-4862, RHBPMS-4863: Container started by upgrade doesn't contain process configuration / Start/stop container buttons are in wrong state after Process Configuration change

karreiro:
See:
- https://issues.jboss.org/browse/RHBPMS-4862
- https://issues.jboss.org/browse/RHBPMS-4863

demo

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
@@ -45,10 +45,8 @@ public class KieServerInstanceManager {
 
     private static final Logger logger = LoggerFactory.getLogger(KieServerInstanceManager.class);
     private static final String CONTAINERS_URI_PART = "/containers/";
-
-    private List<KieServicesClientProvider> clientProviders = new ArrayList<>();
-
     private static KieServerInstanceManager INSTANCE = new KieServerInstanceManager();
+    private List<KieServicesClientProvider> clientProviders = new ArrayList<>();
 
     public KieServerInstanceManager() {
         ServiceLoader<KieServicesClientProvider> loader = ServiceLoader.load(KieServicesClientProvider.class);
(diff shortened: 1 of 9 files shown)


karreiro:
@mswiderski Could you please take a look? :-)

karreiro:
Jenkins please retest this.

karreiro:
Jenkins please retest this.

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
```diff
@@ -778,20 +770,11 @@ private KieScannerStatus mapScannerStatus(KieContainerInstanceImpl kieContainerI
                 messages.add(new Message(Severity.INFO, "Release id successfully updated for container " + id));
                 return new ServiceResponse<ReleaseId>(ServiceResponse.ResponseType.SUCCESS, "Release id successfully updated.", kci.getResource().getReleaseId());
             } else {
-                // no container yet, attempt to create it
```
sutaakar: @karreiro I guess that the test failures are related to this change. Do you have some specific reason why did you delete these lines?

karreiro: Hi @stuaakar,

You're absolutely right. The tests are failing because these lines were removed.

However, I think that we should remove the two failing tests:
```
KieServerContainerCRUDIntegrationTest.testUpdateReleaseIdForNotExistingContainer
KieControllerManagementIntegrationTest.testStartContainerByUpdateKieContainerStatus
```

These lines don't start the container properly, and it should be started before the updating process. Of course, I can rollback and the tests will pass again, but it sounds like an invalid behavior.

WDYT? cc @mswiderski

sutaakar: @karreiro The question is what should happen if Kie server receives update release id request for a non existing container. Current implementation creates a new container with updated release id.

mswiderski:
@karreiro @sutaakar before diving into the code review I would like to take one thing for discussion - what's the reason we would like to start container when we change the process configuration? I'd say start and stop operation should be solely in responsibility of user to decide when that should happen. Imagine when user decided to stop given container and then change its process config like the strategy but does not intend to start it yet. With these changes it can't be done, as soon as you save the config the it's already started. So I'd say this should not be the case.

and when it comes to upgrade maybe instead of starting container when hitting the upgrade button we should make it active only when container is running (is started). As upgrading of something that is not running is not really upgrade, isn't it?

Thoughts?

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
karreiro: @sutaakar the `KieServerImpl` class is not able to create the container properly during the update process.

If we really want to create a container in the `updateContainerReleaseId`, we would need to change signature from this:
```
updateContainerReleaseId(String id, ReleaseId releaseId)
```
to something like this:
```
ServiceResponse<ReleaseId> updateContainerReleaseId(String containerId, KieContainerResource container);
```
But, I had a chat with @mswiderski, and we decided that we should keep the API as is and not start the container during the update process.

To keep the old behavior I'm starting the container _before_ the upgrade process, see:
https://github.com/karreiro/droolsjbpm-integration/blob/14426be667e4d1e20749bac15f4ab46478f5ddd6/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java#L322

karreiro:
Thank you for the considerations, @mswiderski.

I agree with the idea of services doing only one thing (_having a well-defined responsability_), so I'm cool with the approach of the `KieServer.updateContainerReleaseId` just updating the release id.

Regarding the button, *personally*, I think that since we "teach" the user the that the "Upgrade" button starts and updates the container, maybe changing it sounds like something broken. So, I would keep this behavior.

mswiderski:
@karreiro 
- upgrade button - fair enough we can make it to start container when upgrading
- process config save - I would not any operation on the actual container. Maybe even disable it when it's started as changing strategy will be really dangerous then.

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
sutaakar: @karreiro sounds good

karreiro:
@mswiderski could you please clarify the second concern?

karreiro:
@mswiderski I agree :-)

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerContainerCRUDIntegrationTest.java
```diff
@@ -150,22 +150,6 @@ public void testDisposeNonExistingContainer() throws Exception {
     }
 
     @Test
-    public void testUpdateReleaseIdForNotExistingContainer() throws Exception {
```
sutaakar: @karreiro Please try to update the tests instead of deleting them. In this scenario Kie server should return error in case we try to update not existing container (according to new behaviour).

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/KieControllerManagementIntegrationTest.java
```diff
@@ -726,42 +726,6 @@ public void testUpdateContainerNonValidReleaseId() throws Exception {
     }
 
     @Test
-    public void testStartContainerByUpdateKieContainerStatus() throws Exception {
```
sutaakar: @karreiro What is the reason for deleting of this test? The test checks that it is possible to change container status by container spec update. When container status is changed to STARTED then controller should start container on Kie servers.

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/test/java/org/kie/server/integrationtests/common/KieServerContainerCRUDIntegrationTest.java
karreiro: Good catch, @sutaakar. Fixed.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java
```diff
@@ -112,9 +112,9 @@ public synchronized void updateContainerSpec(final String serverTemplateId, fina
         serverTemplate.deleteContainerSpec(currentVersion.getId());
         serverTemplate.addContainerSpec(containerSpec);
 
-        templateStorage.update( serverTemplate );
+        templateStorage.update(serverTemplate);
 
-        notificationService.notify( new ServerTemplateUpdated( serverTemplate ) );
+        notificationService.notify(new ServerTemplateUpdated(serverTemplate));
         // in case container was started before it was update or update comes with status started update container in running servers
         if (currentVersion.getStatus().equals(KieContainerStatus.STARTED) || containerSpec.getStatus().equals(KieContainerStatus.STARTED)) {
```
sutaakar: The test failure is related to this condition. Test expects that if you update container spec's status from STOP to START then controller will send create container request to Kie servers.
@karreiro I think that this condition should be splitted. If former spec was in state STOPPED and newer is in state STARTED then kieServerInstanceManager would start the container. If former spec was in state STARTED and newer is in state STARTED then kieServerInstanceManager would upgrade the container.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
```diff
@@ -254,30 +282,63 @@ public Void doOperation(KieServicesClient client,
...
-                                                                     container.getUrl(),
-                                                                     response.getMsg());
-                                                    }
-                                                    collectContainerInfo(containerSpec,
-                                                                         client,
-                                                                         container);
-                                                    return null;
-                                                }
-                                            });
+    public List<Container> upgradeAndStartContainer(final ServerTemplate serverTemplate,
```
sutaakar: Does this method have sense? When container is started (created on Kie server side) then all configurations are applied. So AFAIK this method has same functionality as startContainer(...).

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java
karreiro: @sutaakar In fact, the test was failing because the previous behavior was `startAndupgrade` the container, and I unintentionally changed it to only `upgrade`. I already applied the fix here, and everything is fine now.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
karreiro: @sutaakar Indeed. The only difference is in the `messages` section, where instead of `Release id successfully updated for container {containerName}`, the user will see `Container {containerId} successfully created with module {releaseId}`. I'm not sure if this is a problem, but if it's not, I think that we can replace the `upgradeAndStartContainer ` calls by `startContainer`.

karreiro:
All changes were applied.

@sutaakar I could not remove the `upgradeAndStartContainer ` method because it's needed in the `SpecManagementServiceImpl` class. However, I'm calling the `startContainer` in the `RuleCapabilitiesServiceImpl` (good catch).

@mswiderski we're not executing any operation in the container during the process of saving a configuration, as we've discussed, see:
demo2

Could you please take a look? :-)

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java
```diff
@@ -150,83 +148,113 @@ public Void doOperation(KieServicesClient client,
...
+                }
+
+                collectContainerInfo(containerSpec, client, container);
+
+                return null;
+            }
+        };
+    }
+
+    void log(final String message,
```
mswiderski: this method looks kind of weird as it include just one line inside, most likely not really needed, isn't it?

karreiro: @mswiderski I extract to a method, to verify (via unit tests) some behaviors, like this one:
https://github.com/karreiro/droolsjbpm-integration/blob/6af1b92b91b930caf3b6dae130a29b426bb8c023/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerInstanceManager.java#L168:L170
Since `logger` is declared like this:
```java
private static final Logger logger
```
I think that this is a good (and maybe weird) way to verify this logic :-)

mswiderski: fair enough, if there is a good reason for it then I am fine with it :)

PR #1215, opened 2017-09-25
RHBRMS-1220 removing drools-verifier

ryanzhang:
Removed the droosl-verifier from brms/bpms package


mswiderski:
7.3 is no longer maintained

PR #1216, opened 2017-09-25
RHBRMS-1220 removing drools-verifier

ryanzhang:
Remove the drools-verifier from brms/bpm package


mswiderski:
Jenkins retest this

PR #1220, opened 2017-10-02
[8.0.x] JBPM-6456: Unable to open tasks that contain subforms.

pefernan:
Adding package security to XStreamMarshaller to avoid `com.thoughtworks.xstream.security.ForbiddenClassException` when unmarshalling java objects comming from kie-server.

Added ability to configure a list of allowed packages using a comma separated list on the `org.kie.server.xstream.enabled.packages` system property. Like `-Dorg.kie.server.xstream.enabled.packages=packageone.**,packagetwo.**`

@jsoltes  @mswiderski could you please review? (Sorry for all the format changes)

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -26,6 +26,8 @@ import java.util.Set;
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
+import com.thoughtworks.xstream.security.WildcardTypePermission;
+import org.apache.commons.lang3.StringUtils;
 import org.drools.core.runtime.help.impl.XStreamXML;
 import org.kie.server.api.commands.CallContainerCommand;
 import org.kie.server.api.commands.CommandScript;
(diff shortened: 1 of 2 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -79,6 +81,9 @@
 public class XStreamMarshaller
         implements Marshaller {
 
+    public static final String SYSTEM_XSTREAM_ENABLED_PACKAGES = "org.kie.server.xstream.enabled.packages";
```
mswiderski: can we move this system property to KieServerConstants where all other system properties are defined?

pefernan:
@mswiderski sure np

sutaakar:
@pefernan Is it possible to provide some test coverage together with the fix?

pefernan:
@mswiderski better now?
@sutaakar the problem is that testing this implies a big integration test between jbpm-wb and kie-server. This fixes an error that happens only when jbpm-wb asks for task data and that task data contains a java object created with Data Modeller.

mswiderski:
@pefernan yes it is better, thanks

sutaakar:
@pefernan Can you provide example of marshalled task data? That way we can put a test to https://github.com/kiegroup/droolsjbpm-integration/blob/4f808b3499bbdb177f4c16b7fdfbdf20d9341a6e/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/XStreamMarshallerTest.java simulating the issue.

pefernan:
@sutaakar this is the kind of response that the kie-server gives https://gist.github.com/pefernan/cded0fa2f7a12b8ed47500b04e1e18d6

pefernan:
@mswiderski is it ok to merge?

mswiderski:
@pefernan yes, it is from my point of view @sutakar anything else you want to add for it?

sutaakar:
@mswiderski @pefernan I would like to see there some tests covering the change. I have tried to marshall the example from comment above, but it seems to pass with old marshaller.
I guess I will need to contact @jsoltes to figure out some example for the test.

mswiderski:
can we then merge this and add test later, asking mainly to avoid people hit this issue when working or doing tests on top of latest builds.

sutaakar:
Sure, will add the test later

PR #1224, opened 2017-10-03
[7.3.x]JBPM-6449 - Kie Server Router should retry when updating controller c…

mswiderski:
…onfig in case of controller's unavailability, JBPM-6448 - Connectivity lost between kie server router and kie server (#1219)

### DIFF
--- a/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/Configuration.java
+++ b/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/Configuration.java
@@ -25,6 +25,8 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 
+import org.kie.server.router.utils.FailedHostInfo;
+
 public class Configuration {
 
     private Map<String, List<String>> hostsPerServer = new ConcurrentHashMap<>();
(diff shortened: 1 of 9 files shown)


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

sutaakar:
I suspect that the issue is caused by immutability of controller URL in Smart router. Controller URL in Smart router used in KieServerRouterControllerRecoveryTest points to controller from previous test in case there was any.

mswiderski:
@sutaakar not sure where is that?
I removed the static final so it gets the value every time KieServerRouter is created https://github.com/kiegroup/droolsjbpm-integration/pull/1224/files#diff-55dbd4dfb4d92f718c1dbf4cb1edefbcR61

what do I miss?

sutaakar:
Here it is used too: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/handlers/AdminHttpHandler.java#L53

mswiderski:
thank you @sutaakar updated let's see how it goes now ...

why this didn't fail on master... anyway, once this is green will prepare same changes (remove static final) for master.

sutaakar:
The failure happens randomly as it depends on order of the tests. If KieServerRouterUnavailabilityRecoveryTest is executed as a first test then it doesn't fail.

mswiderski:
@sutaakar so we are so lucky on master and so unlucky on 7.3.x... interesting :)

sutaakar:
:)

PR #1233, opened 2017-10-16
Added Drool's ProjectClassloader store in the shared map

desmax74:
Classes created by Drools during the compilation of the .<>drl files and shared in the map to exports in the Maven compiler response


desmax74:
@porcelli @mariofusco

PR #1235, opened 2017-10-17
[RHBPMS-4312] Added loading password from from EAP vault

rstancel:
Jira: https://issues.jboss.org/browse/RHBPMS-4312

### DIFF
--- a/drools-android/pom.xml
+++ b/drools-android/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.drools</groupId>
     <artifactId>droolsjbpm-integration</artifactId>
-    <version>8.0.0-SNAPSHOT</version>
+    <version>7.5.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>drools-android</artifactId>
--- a/drools-benchmark/pom.xml
+++ b/drools-benchmark/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.drools</groupId>
     <artifactId>droolsjbpm-integration</artifactId>
-    <version>8.0.0-SNAPSHOT</version>
+    <version>7.5.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>drools-benchmark</artifactId>
--- a/drools-examples-android/kiecontainer/pom.xml
+++ b/drools-examples-android/kiecontainer/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.drools</groupId>
     <artifactId>drools-examples-android</artifactId>
-    <version>8.0.0-SNAPSHOT</version>
+    <version>7.5.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>kiecontainer</artifactId>
--- a/drools-examples-android/pom.xml
+++ b/drools-examples-android/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.drools</groupId>
     <artifactId>droolsjbpm-integration</artifactId>
-    <version>8.0.0-SNAPSHOT</version>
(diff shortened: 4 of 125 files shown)


mswiderski:
ok to test

mswiderski:
@rstancel the build failed due to some enforcer rules violations. You can reproduce it locally by running build with -Dfull param.

@etirelli what do you think about the approach?

etirelli:
@mswiderski @rstancel 

I have two concerns with the current approach:

1. When the "property" is set to use the vault (or a keystore, see point 2), there can't be a fallback. If the property was deliberately set and the app fails to open the vault (or keystore), it has to fail fast. Otherwise it becomes a vector for attack. 

2. I never used the vault, but I assume this is an EAP/wildfly specific component? I believe the requirement is for the kieserver to still be compatible with other containers (like WAS and tomcat). Can we use a java keystore instead? That way we would be container agnostic.

mswiderski:
@etirelli 

> When the "property" is set to use the vault (or a keystore, see point 2), there can't be a fallback. If the property was deliberately set and the app fails to open the vault (or keystore), it has to fail fast. Otherwise it becomes a vector for attack.

makes sense, I agree we should bring this in

> I never used the vault, but I assume this is an EAP/wildfly specific component? I believe the requirement is for the kieserver to still be compatible with other containers (like WAS and tomcat). Can we use a java keystore instead? That way we would be container agnostic.

I'd say the biggest problem is that access to key store must be protected and then we end up with exact same issue - how to pass password to gain access to keystore? I did a small prototype months ago with keystores and that was not really suitable. With this we have at least first step to have a secure on EAP/WildFly.
So unless i missed really big part about keystores I don't see any other option at the moment. Thoughts?

etirelli:
keystores use the premise of a trusted local environment. I.e., the keystore password is set locally, usually in plain text, but it is never sent over the wire or accessible to someone without the proper rights on the local machine. It the local environment is not trusted, then there are a number of other attack vectors that are possible, so I don't think this is really an issue. Unless I am missing something. 
What was the problem with your keystore prototype?

etirelli:
BTW, we already have code to access a local keystore in Drools, so it would just be a matter of refactoring that code to make it generic:

https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/util/KeyStoreHelper.java

mswiderski:
in general main issue was with gaining access to the keystore - as password must be given one way or another. I tried same approach as here https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/util/KeyStoreHelper.java#L94

but with our stuff it's really easy to get that value - once you're in the jvm get access to System.getProperty and then you have the password. So this is the main issue - I don't think we can consider local keystore as trusted as user's code can capture that info and gain access to keystore too.

etirelli:
If the user code is not trusted, then the security policy feature should be enabled preventing him access to the filesystem (so he won't have access to the keystore) and even to the system property if necessary. 

But let me put this another way: there is nothing wrong with adding support for the EAP/wildfly vault, just keep in mind that soon there might be a request to solve the same problem on WAS/Tomcat. 

If you think it is worth going with this solution for EAP/Wildfly, go ahead.

mswiderski:
Main idea with vault was to rely on already approved mechanisms

If using keystore and giving password in clear way to it won't be questioned then fine with me

mswiderski:
@etirelli do you mind adding your comments/idea about using keystore to the jira so we can get option from security stand point?

rstancel:
@mswiderski I have updated functionality based on this discussion. Now, when vault property is present it is mandatory to use it. If it is not present, it is loaded from configuration as before.

mswiderski:
Jenkins retest this

PR #1236, opened 2017-10-18
RHBPMS-5021 - Duplicate Kie server registration to Smart router

mswiderski:
@sutaakar here is the move of registration in the controller from router via async executor. The drawback is it might lead to bit of delays based on the interval the scheduler runs. I changed it to 5 secs to not delay to much but at the same time collect many requests from kie server (one per container and alias) before running the async job.


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1238, opened 2017-10-19
Use XStreamUtils.createTrustingXStream(..) helper

mswiderski:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/271

@ge0ffrey @mariofusco could you please have a look?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/KieServerTypePermission.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/KieServerTypePermission.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017 - 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.server.api.marshalling.xstream;
+
+import static org.kie.server.api.KieServerConstants.SYSTEM_XSTREAM_ENABLED_PACKAGES;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.kie.internal.utils.KieTypeResolver;
+
+import com.thoughtworks.xstream.security.WildcardTypePermission;
+
+/**
+ * Kie Server specific type permission implementation that allows:
+ * <ul>
+ *  <li>org.kie.server.api.model classes (including subpackages)</li>
+ *  <li>classes that come from kjar or its dependencies (were loaded by kjar class loader)</li>
+ *  <li>set of classes explicitly given when constructing this instance</li>
+ *  <li>optionally defined by wildcard that is given via system property: org.kie.server.xstream.enabled.packages (comma separated list of wildcard expressions)</li>
+ * </ul>
+ *
+ */
+public class KieServerTypePermission extends WildcardTypePermission {
+    
+    private static final String[] WHITELISTED_PACKAGES = new String[]{
+                                                                      "org.kie.server.api.model.**",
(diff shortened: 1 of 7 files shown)


kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/CustomXstreamMarshallerBuilder.java
```diff
@@ -33,7 +35,8 @@ public Marshaller build(Set<Class<?>> classes, MarshallingFormat format, ClassLo
             return new XStreamMarshaller(classes, classLoader) {
                 @Override
                 protected void buildMarshaller(Set<Class<?>> classes, ClassLoader classLoader) {
-                    xstream = new XStream(new PureJavaReflectionProvider(), new DomDriver("UTF-8", new XmlFriendlyNameCoder("_-", "_")));
+                    xstream = XStreamUtils.createNonTrustingXStream(new PureJavaReflectionProvider(), new DomDriver("UTF-8", new XmlFriendlyNameCoder("_-", "_")));
+                    xstream.addPermission(new WildcardTypePermission(new String[]{"org.kie.server.api.**"}));
```
ge0ffrey: Do you vouch that there every class in this package was designed with XStream marshalling in mind and therefore can't be abused as an exploit? :) Nah, just kidding, as that's too insane to ask. 

Soft suggestion: it would be preferable if all classes that are designed for XStream marshalling are in a separate subpackage (or subpackage pattern) so that we can only whitelist those. For example, the pattern could be "org.kie.server.api.**.domain.**" or "org.kie.server.api.**.data.**" or ...

mswiderski: this is for test purpose only so it tests completely different thing but in case this is shown as example I'd like to have this as non trusted with extra permissions ... less relevant what is allowed :)

ge0ffrey: agreed, for tests I even use the trusting version :)

mariofusco:
ok to test

mariofusco:
@mswiderski The test are now failing for this reason:

Caused by: com.thoughtworks.xstream.security.ForbiddenClassException: org.kie.server.controller.api.model.spec.ServerTemplate

Please update your pull request by whitelisting that package to the xstream instance used by the kie-server.

mswiderski:
@mariofusco will check it on Monday, thanks

mswiderski:
depends on https://github.com/kiegroup/drools/pull/1554

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -152,18 +148,7 @@ protected void configureMarshaller(Set<Class<?>> classes,
...
-        String[] filter;
-
-        if (StringUtils.isEmpty(packageList)) {
-            filter = new String[]{DEFAULT_PACKAGES_WILDCARD};
-        } else {
-            filter = packageList.split(",");
-        }
-
-        this.xstream.addPermission(new WildcardTypePermission(filter));
+        this.xstream.addPermission(new KieServerTypePermission(classes));
```
mariofusco: @mswiderski is it possible that the Set classes is null in some circumstances? KieServerTypePermission is now failing with a NPE at line 56 and this seems that only case where the Set is passed from the outside.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/KieServerTypePermission.java
```diff
@@ -0,0 +1,79 @@
...
+    
+    private static final String[] WHITELISTED_PACKAGES = new String[]{
+                                                                      "org.kie.server.api.model.**"
+    };
+
+    private Set<Class<?>> classes;
+    
+    public KieServerTypePermission(Set<Class<?>> classes) {
+        super(patterns());
+        this.classes = classes;
```
mariofusco: @mswiderski if you want you could prevent a NPE by creating a new Set in case you're passed a null one. Also I think that it could be a good idea to declare the Set classes final.

mswiderski:
Jenkins retest this

mswiderski:
@mariofusco @ge0ffrey @tsurdilo guys could you please review once again? Now it has all the changes that should be good enough and secure enough.

PR #1241, opened 2017-10-20
Skip JMS tests using Maven property

sutaakar:
Useful for example for testing WildFly with Keycloak.


sutaakar:
Jenkins retest this

PR #1243, opened 2017-10-26
DO NOT MERGE - RHBPMS-5026 - Version difference of jackson artifacts

mswiderski:
depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/583

see comments in ^^ PR for details
@mbiarnes @psiroky @sutaakar @krisv


mswiderski:
ok, this will not work here as expected as the builds are run with wildfly 10.x and this requires wildfly 11

PR #1248, opened 2017-10-27
AF-593+: Move XStreamUtils to kie-soup.

mbarkley:


kie-spring/src/test/java/org/kie/spring/CodeVersion.java
```diff
@@ -18,6 +18,6 @@
 public class CodeVersion {
 
     // Do not modify this: this is automatically incremented by the maven replacer plugin
-    public final static String VERSION = "8.0.0-SNAPSHOT";
+    public final static String VERSION = "7.5.0-SNAPSHOT";
```
ederign: should we manually update this?

mbarkley: This was automatically updated when I ran a build. Not sure what to do about it.

mbarkley: @ederign Looks like @psiroky has already fixed this on master. But I would prefer not to rebase as it would cause the tests to have to run again.

PR #1249, opened 2017-10-30
revapi: add missing justification

psiroky:
Please don't merge yet, will need to be merged together with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/584

### DIFF
--- a/kie-server-parent/kie-server-api/src/build/revapi-config.json
+++ b/kie-server-parent/kie-server-api/src/build/revapi-config.json
@@ -37,6 +37,15 @@
           "annotationType": "javax.xml.bind.annotation.XmlElements",
           "elementKind": "annotation",
           "justification": "PLANNER-815 Support real-time planning"
+        },
+        {
+          "code": "java.field.removedWithConstant",
+          "old": "field org.kie.server.api.marshalling.xstream.XStreamMarshaller.DEFAULT_PACKAGES_WILDCARD",
+          "package": "org.kie.server.api.marshalling.xstream",
+          "classSimpleName": "XStreamMarshaller",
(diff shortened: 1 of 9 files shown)


ge0ffrey:
Petr, feel free to merge when you're merging the prequal. Do note that the link above is not referencing another PR.

psiroky:
Sorry about that, this is the link to the PR: https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/588

psiroky:
I had to do further changes to make the revapi work. @MarianMacik could you please review again?

MarianMacik:
@psiroky Sorry for the late response, I overlooked this PR. I am a little bit worried about this change since you upgraded revapi to a quite high version. What was the reason? I suppose JDK 9 support. In addition to that, you raised previous KIE version to 7.4.1 - so now we are  comparing master vs 7.4.1 so all ignores should be deleted from all repositories where revapi is configured. And why there is now kie-server-rest-jbpm-search/../revapi-config.json configured when there was none before? Also elementKind: annotation has been removed - is there any reason behind it? Thanks.

psiroky:
Yes, JDK9 support was the only reason. I assumed that we want to test against the latest released version, so I did the upgrade as well (since with 7.0.0.Final the new revapi was reporting different violations which I would need to add ignores, which seemed useless as we would bump the version eventually anyway). I'll the clean-up to someone else though. I've already spent way more time on this that I would like. The new revapi-config is simply needed, otherwise revapi fails. The elementKind: annotation was either wrong also before or the behavior of the revapi changed. The elementKind should be the element the annotation is on (e.g. field, method). Considering the Swagger annotations are just metadata I decided to ignore all the element kinds. We can change that if needed.

PR #1250, opened 2017-10-30
Testr removing kie-ci as it's hopefully not needed

lucamolteni:
Do not merge. 

This should trigger CI


psiroky:
I don't wanna be the nitpicker here, but merging commit which starts with "[DO-NOT-MERGE]" is certainly not something I would expect :)

mariofusco:
@psiroky sorry I just forgot to edit the commit subject before merging it, but the commit itself is ok

PR #1252, opened 2017-10-31
RHBPMS-5042 - Signal fails with 'Could not find process instance with…

mswiderski:
… id' when the Audit mode is NONE

depends on https://github.com/kiegroup/jbpm/pull/1050


mswiderski:
Jenkins retest this

PR #1262, opened 2017-11-07
JBPM-6593 - Clustering support for Standalone controller

mswiderski:
@sutaakar could you please have a look?

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/storage/KieServerTemplateStorage.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/storage/KieServerTemplateStorage.java
@@ -67,4 +67,8 @@ public interface KieServerTemplateStorage {
      * @return
      */
     ServerTemplate delete(String identifier);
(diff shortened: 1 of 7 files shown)


psiroky:
Build finished. 4182 tests run, 35 skipped, 7 failed.

mswiderski:
Jenkins retest this

sutaakar:
@mswiderski Looks good for Kie servers connected through HTTP.
For Kie server connected through WebSockets it can become more complicated as some Kie servers can be connected to one controller, some to other ones. In such case controller would have to notify its Kie servers about changes in their configuration. What do you think about it?

mswiderski:
@sutaakar this change is to sync the configuration on each controller using same file, so as soon as one writes to it the other reloads that config. 

and what you're referring to is that you have cluster of controllers that behind load balancer and when you change state of the server template kie server is connected to you would like to notify them but for web socket it will only work when the change was done on the server where kie server is connected to right?

for that I would go for another jira as it does affect both (standalone and embedded) controllers.

sutaakar:
@mswiderski Right, will create a JIRA for that

mswiderski:
Jenkins retest this

PR #1264, opened 2017-11-08
[RHBPMS-4312] Added loading passwords from KeyStore in drools core

rstancel:
Jira: https://issues.jboss.org/browse/RHBPMS-4312


rstancel:
This PR depends on https://github.com/kiegroup/drools/pull/1588 and cannot pass until it is merged

mswiderski:
@rstancel if you name the branches the same the build will take both into account when running

mswiderski:
@rstancel if I am not mistaken this now completely switches to keystone - meaning existing installation when upgraded won't work anymore unless keystore is created with password, right?

while for product I think that might be acceptable I think for community we should have this fallback (with warning in the logs pointing to docs how to configure keystore) that will use defaults or system property password

@etirelli @rstancel what do you think guys?

rstancel:
@mswiderski you are exactly right. I made it because this was a result from discussion with security guys in previous approach

etirelli:
@rstancel as Maciej said, while we can recommend keystore usage, it presents other issues for community, specially on setup. On trusted environments, a keystore should not be mandatory.

rstancel:
Closing this PR. It is substituded by this one: https://github.com/kiegroup/droolsjbpm-integration/pull/1267

PR #1269, opened 2017-11-10
JBPM-6613: Reload templates from file in one step

sutaakar:
To avoid situation when template maps are empty.


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1270, opened 2017-11-12
Fix duplicated classes in test suites

sutaakar:
Fixed by moving dependencies for local tests into separated profile (active by default). It should be automatically picked by IDE when running JUnit tests (verified in Eclipse).

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RollingUpdateProcessServiceIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RollingUpdateProcessServiceIntegrationTest.java
@@ -16,10 +16,7 @@
 package org.kie.server.integrationtests.jbpm;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.kie.server.remote.rest.jbpm.resources.Messages.NO_PROCESS_AVAILABLE_WITH_ID;
-import static org.kie.server.remote.rest.jbpm.resources.Messages.PROCESS_DEFINITION_NOT_FOUND;
 
-import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


psiroky:
Is this fixing the following duplications?
```
Found in:
09:05:52     io.netty:netty-all:jar:4.0.33.Final:test
09:05:52     io.netty:netty-common:jar:4.1.16.Final:test
09:05:52   Duplicate classes:
09:05:52     io/netty/util/concurrent/ImmediateExecutor.class
09:05:52     io/netty/util/internal/logging/AbstractInternalLogger.clas
...
...
```

sutaakar:
@psiroky Yes

psiroky:
@sutaakar we may need some additional profiles, but ultimately the root cause is that two different netty versions are getting in. This needs to be fixed (e.g. via exclusions) at the appropriate places.

sutaakar:
@psiroky Yes, that is the root cause. I think that the important point is why this situation happened and how to prevent such issues in future.

netty is brought there from two sources:
- WildFly JMS client (which is required for JMS tests)
- through kie-server-services-drools artifact, transitively from uberfire-commons (which is in classpath to make local test runs available)

From my point of view the best way to fix this and prevent such issues in future is to separate dependencies brought by JMS clients and dependencies brought by local test run into separate profiles, so they won't be activated together.
Exclusions can bring errors and instabilities as we remove something what might be required by dependencies to work properly.

psiroky:
@sutaakar ok, that makes sense. I don't know if we can make sure the netty versions are synchronized, but if so we should at least try to do that.

psiroky:
Thinking about this again, it will be impossible to fix the alignment in all cases. E.g. when the projects are aligned to WF11, but we still want to be testing with the WF10 jms-bom (or the other way around).

sutaakar:
@mswiderski Please take a look.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RollingUpdateProcessServiceIntegrationTest.java
```diff
@@ -309,8 +306,8 @@ public void testContainerIdAsAlias() throws Exception {
 
         assertClientException(() -> processClient.startProcess(CONTAINER_ALIAS, PROCESS_ID_EVALUATION_2),
                 404,
-                MessageFormat.format(PROCESS_DEFINITION_NOT_FOUND, PROCESS_ID_EVALUATION_2, CONTAINER_ALIAS),
-                MessageFormat.format(NO_PROCESS_AVAILABLE_WITH_ID, PROCESS_ID_EVALUATION_2));
+                "Could not find process definition \"" + PROCESS_ID_EVALUATION_2 + "\" in container \"" + CONTAINER_ALIAS + "\"",
+                "No process available with given id : " + PROCESS_ID_EVALUATION_2);
```
mswiderski: why this has been changed?

sutaakar: It was changed because the message was imported from kie-server-rest-jbpm. To make the tests independent on rest/service dependencies (as they aren't available in container test runs) I replaced it with strings. Not the cleanest solution TBH, I intended to make build fixed fast.

PR #1271, opened 2017-11-13
JBPM-6615 - Add WorkItemImpl class to list of XStream allowed types

cristianonicolai:


mswiderski:
@cristianonicolai could we add a test case for it here https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/JobServiceIntegrationTest.java?
It should be enough to create work item and add this to job data

cristianonicolai:
@mswiderski done, althought I believe I only had to include the class on the jaxb marshaller because of the test, in order to serialize the object.

mswiderski:
Jenkins retest this

PR #1274, opened 2017-11-16
JBPM-6625 - Remove advanced queries (BPMQueries) KIE Server extension

mswiderski:
@krisv let me know if we should put this on hold or can we merge it

please note that the client side is preserved (see tests) so you can use type safe query filter spec for process instance and task queries. Though you have the flexibility to define many queries and use this type safe filtering on top of them.

@DuncanDoyle just so you know... look it's not all you have done is removed, the best part is still there


mswiderski:
Jenkins retest this

sutaakar:
@mswiderski Please update artificial dependency in https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-custom-extension/pom.xml to make tests run sequentially.

krisv:
I am fine with merging this on master

mswiderski:
Jenkins retest this

PR #1277, opened 2017-11-17
Override jboss-remoting version for tests to be compatible with jboss…

mbarkley:
…-remote-naming.

### DIFF
--- a/kie-server-parent/kie-server-tests/pom.xml
+++ b/kie-server-parent/kie-server-tests/pom.xml
@@ -52,6 +52,8 @@
     <version.tomcat8>8.5.12</version.tomcat8>
     <version.wildfly10>10.1.0.Final</version.wildfly10>
     <version.wildfly11>11.0.0.Final</version.wildfly11>
+    <!-- Need to override jboss-remoting version to be compatible with jboss-remote-naming -->
+    <version.org.jboss.remoting>4.0.18.Final</version.org.jboss.remoting>
     <!-- The EAP 7 binary can't be anonymously downloaded because of the license. It can be downloaded manually
          and for free (e.g. from http://jbossas.jboss.org/downloads.html) and the zip location needs to be specified
          here or via system property when running the build (don't forget to use the `file://` prefix when
@@ -554,6 +556,11 @@
             <scope>import</scope>
           </dependency>
           <dependency>
+            <groupId>org.jboss.remoting</groupId>
+            <artifactId>jboss-remoting</artifactId>
+            <version>${version.org.jboss.remoting}</version>
+          </dependency>
+          <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>${version.com.google.guava}</version>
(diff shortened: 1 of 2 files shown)


mbarkley:
Related:
https://github.com/errai/errai/pull/307
https://github.com/kiegroup/kie-soup/pull/15
https://github.com/AppFormer/uberfire/pull/916
https://github.com/dashbuilder/dashbuilder/pull/409
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/615
https://github.com/kiegroup/drools-wb/pull/709
https://github.com/kiegroup/droolsjbpm-integration/pull/1277

kie-server-parent/kie-server-tests/pom.xml
```diff
@@ -52,6 +52,8 @@
     <version.tomcat8>8.5.12</version.tomcat8>
     <version.wildfly10>10.1.0.Final</version.wildfly10>
     <version.wildfly11>11.0.0.Final</version.wildfly11>
+    <!-- Need to override jboss-remoting version to be compatible with jboss-remote-naming -->
+    <version.org.jboss.remoting>4.0.18.Final</version.org.jboss.remoting>
```
mbarkley: @mswiderski Since you have lots of recent commits in this directory, would you mind reviewing this? (Or recommend someone else if relevant?)

The ip-bom upgrade in the related PRs updates wildfly-core, which required updating jboss-remoting to version 5.0.3.Final, but unfortunately the most recent jboss-remote-naming release is only compatible with jboss-remoting 4.0.18.Final.

As far as I can tell, jboss-remote-naming is only used in tests as a transitive dependency of wildfly-jms-client-bom, so I think this change. Am I missing anything?

//cc @psiroky @mariofusco @manstis @mbiarnes

mswiderski: @mbarkley as long as it works then it's fine 

@sutaakar could you please have a look when you have chance?

mbarkley: Okay. This change got us from many test failures down to just 2. Looking into those now.

sutaakar: Looks good to me. Those 2 failures seems to be caused by behaviour change caused by dependency updates. Seems that https://github.com/kiegroup/droolsjbpm-integration/blob/494cf10855b338ff2472c17a4cc3e9e902c7920d/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/proxy/KieServerProxyClient.java#L148 needs to be updated to catch this new exception.

mbarkley: @sutaakar thanks for pointing that out. I will update the PR shortly.

mbarkley:
Most recent commit fixes the two test failures (thanks for the help @sutaakar). XNIO now throws an IllegalArgumentException for unresolvable hosts. I don't think this is intentional and have opened an issue [here](https://issues.jboss.org/browse/XNIO-313).

PR #1285, opened 2017-11-23
[RHBRMS-3027] remove old module from KieRepo when the KieContainer is upgraded to a newer version

mariofusco:


mariofusco:

PR #1290, opened 2017-11-26
[JBPM-6647] Jobs: NPE when selecting already deleted job

bsremac:


bsremac:
@mswiderski Added a small fix for one corner use case I've stumbled upon recently.

cristianonicolai:
@bsremac is this case something we could cover with a unit/integration test?

mswiderski:
@bsremac I'd actually throw an exception if job was not found so then the REST endpoint could catch it and produce not found (404) response rather than ok (200) with empty payload.

bsremac:
@mswiderski Is this what you had in mind?

mswiderski:
@bsremac yes, pretty much. If you don't mind could you check if the ExecutorResource will catch it and produce notFound jaxrs response?

bsremac:
@mswiderski Added another ```catch``` block which returns ```notFound``` in this case.

mswiderski:
@bsremac that's exactly what I had in mind, thanks

do you plan to add a test case for it too as commented by @cristianonicolai ?

bsremac:
@mswiderski @cristianonicolai Yes, I will definitely try to cover this with a test in ``jbpm-wb```. 
I wrote "try" as I'm not sure at the moment what type of test is an option (unit or integration). But will figure it out eventually :)

sutaakar:
@bsremac You may implement it as an integration test here in Kie server test suite. See for example https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/JobServiceIntegrationTest.java

bsremac:
@cristianonicolai @mswiderski Working on the integration test as suggested by @sutaakar. Will add it to this PR once I finish

bsremac:
@mswiderski @cristianonicolai Added an integration test.
Initially I tried to replicate the whole use case, but introducing log cleanup and waiting for the executor service to execute the command seemed like a lot of overhead when there are 6 tests to run subsequently.
Instead I went for a simplified version where the exception is thrown when querying for non-existent job id, which is basically the same as deleting a job and then querying for its job id.

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/JobServiceIntegrationTest.java
```diff
@@ -529,4 +529,38 @@ public void testScheduleAndRunJobWithWorkItem() throws Exception {
...
+
+    @Test
+    public void testGetDeletedJob() {
+        final long jobId = 1L;
+        KieServicesException expectedException = new KieServicesException();
+
+        try {
+            jobServicesClient.getRequestById(jobId,
+                                             false,
+                                             false);
```
cristianonicolai: I would suggest adding a "fail" right after the getRequestById call, that way the entire test would fail if the exception is not thrown as expected. You can then also just validate the exception inside the catch statement.

sutaakar:
@bsremac Just a tip, you can use already prepared functionality to invoke failing request and validate the exception - https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/basetests/RestJmsSharedBaseIntegrationTest.java#L146

Example of this method can be seen for example in https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/QueryDataServiceIntegrationTest.java#L399-L403

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/JobServiceIntegrationTest.java
```diff
@@ -529,4 +529,38 @@ public void testScheduleAndRunJobWithWorkItem() throws Exception {
                      jobRequest.getCommandName());
         assertNotNull(jobRequest.getData().get("workItem"));
     }
+
+    @Test
+    public void testGetDeletedJob() {
+        final long jobId = 1L;
```
sutaakar: I guess that this should be updated to represent non-existent job id. Maybe to put there negative value?

bsremac:
@sutaakar I switched to using the existing test API for treating the exception thrown. Thanks for pointing out that piece of code to me.
As for the positive vs. negative value for jobId in the test, I went with a positive value initially as it resembles more to the use case which brought this issue to our attention, nothing else really. Anyway, I changed it to negative value for clarity reasons and renamed the test accordingly.

PR #1292, opened 2017-11-27
JBPM-6243 - Use model classes for serializing collections

cristianonicolai:
@mswiderski changing the collection return types so objects can be properly serialized when using the new websocket client. So far this was the only way I could the entire graph to be serialized/deserialized without marshaling errors.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/src/build/revapi-config.json
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/src/build/revapi-config.json
@@ -14,7 +14,6 @@
       }
     }
   },
-
   "ignores": {
     "revapi": {
       "_comment": "Changes between 7.0.0.Final and the current branch. These changes are desired and thus ignored.",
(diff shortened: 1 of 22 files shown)


cristianonicolai:
Dependent PRs:

- https://github.com/kiegroup/droolsjbpm-integration/pull/1292
- https://github.com/kiegroup/jbpm-wb/pull/951
- https://github.com/kiegroup/kie-wb-common/pull/1305

mswiderski:
@cristianonicolai what was the issue with collections there? Main concern here is that  the content sent over the wire no changes and thus might impact people using controller REST endpoints. For instance we have kie server maven plugin that uses it.

cristianonicolai:
@mswiderski I couldn't get the marshaller to deserialize the content using generic collections (ex: ArrayList), the domain objects inside the collection ( ex: ServerTemplateKey ) would always be loaded as a hashmap instead. It seems that without the specific type, the marshaller doesn't know the proper object type to instantiate. 
Sure, I wasn't happy about changing the return types either, so I decided to open a specific PR for review and discuss any possible alternative.

cristianonicolai:

cristianonicolai:

PR #1294, opened 2017-11-28
RHBPMS-4819: renamed groupId batik to org.apache.xmlgraphics

mbiarnes:


jbpm-process-svg/pom.xml
```diff
@@ -17,7 +17,7 @@
 
   <dependencies>
     <dependency>
-      <groupId>batik</groupId>
+      <groupId>org.apache.xmlgraphics</groupId>
       <artifactId>batik-svg-dom</artifactId>
       <exclusions>
         <exclusion>
```
mareknovotny: I think this exclusion is not useful anymore as it is not in dependency tree, but there is commons-logging which is banned

```
] +- org.apache.xmlgraphics:batik-svg-dom:jar:1.9.1:compile
[INFO] |  +- org.apache.xmlgraphics:batik-awt-util:jar:1.9.1:compile
[INFO] |  |  \- org.apache.xmlgraphics:xmlgraphics-commons:jar:2.2:compile
[INFO] |  |     +- commons-io:commons-io:jar:2.5:compile
[INFO] |  |     \- commons-logging:commons-logging:jar:1.1.1:compile
```

mbiarnes: @mareknovotny this means to remove the exclusion of
        <exclusion>
          <artifactId>xercesImpl</artifactId>
          <groupId>xerces</groupId>
</exclusion>
and add
        <exclusion>
          <artifactId>commons-logging</artifactId>
          <groupId>commons-logging</groupId>
</exclusion> ???

mareknovotny: yes, if the ban of the `commons-logging` is correct and appropriate, I can't say that, but from discussion with @psiroky you should exclude it and use slfj bridge for logging

psiroky: Correct, as per https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L430, `commons-logging` is banned and should be replaced by `jcl-over-slf4j`

jbpm-process-svg/pom.xml
```diff
@@ -17,17 +17,28 @@
...
     <dependency>
-      <groupId>batik</groupId>
+      <groupId>org.slf4j</groupId>
+      <artifactId>jcl-over-slf4j</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore</artifactId>
+    </dependency>
```
psiroky: This is not needed, we just need the `jcl-over-slf4j` dep. Sorry, I should have been more clear.

PR #1298, opened 2017-11-30
Add JMS support for WildFly11 tests

sutaakar:
Nonbreaking changes to add support for WildFly 11 JMS tests - so WildFly 10 test are still working.

To make WildFly 11 working the `version.org.jboss.remoting` property needs to be deleted and `<version.org.jboss.marshalling>2.0.2.Final</version.org.jboss.marshalling>` needs to be added.
Also https://github.com/kiegroup/droolsjbpm-integration/pull/1243 needs to be applied to fix jacson issues.

### DIFF
--- a/kie-server-parent/kie-server-tests/pom.xml
+++ b/kie-server-parent/kie-server-tests/pom.xml
@@ -649,10 +649,9 @@
     <profile>
       <id>wildfly11</id>
       <properties>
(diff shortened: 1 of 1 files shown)


mswiderski:
@sutaakar I included these changes in this PR https://github.com/kiegroup/droolsjbpm-integration/pull/1299
It has both profile changes and json marshaller updates

PR #1299, opened 2017-11-30
RHBPMS-5026 and RHBRMS-3033 - Version difference of jackson artifacts…

mswiderski:
… - support for WildFly 11 and EAP 7.1

this provides support for WildFly 11 and EAp 7.1 which requires upgrade of jackson library (which is not backward compatible in terms what is used in KIE Server's JSON Marshaller).

At the same time it removes wildfly10 profile and moves EAP 7 to use WildFly 11 instead of 10.

@sutaakar could you please review?

It needs to be merged with https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/583

PR #1301, opened 2017-12-07
RHBRMS-3033: trying to use only bundled com.fasterxml.jackson libs

mareknovotny:
https://issues.jboss.org/browse/RHBRMS-3033

This is one point to fix kie-server, I am not sure if there are more such deployment to need exclusions like this

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/shared-ee6-ee7-resources/WEB-INF/jboss-deployment-structure.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/shared-ee6-ee7-resources/WEB-INF/jboss-deployment-structure.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ JBoss, Home of Professional Open Source
+  ~ Copyright 2011 Red Hat Inc. and/or its affiliates and other contributors
+  ~ as indicated by the @author tags. All rights reserved.
+  ~ See the copyright.txt in the distribution for a
+  ~ full listing of individual contributors.
+  ~
+  ~ This copyrighted material is made available to anyone wishing to use,
+  ~ modify, copy, or redistribute it subject to the terms and conditions
+  ~ of the GNU Lesser General Public License, v. 2.1.
+  ~ This program is distributed in the hope that it will be useful, but WITHOUT A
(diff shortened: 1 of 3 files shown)


mareknovotny:
@psiroky of course, I did a safe tryout ;)
I will clean them

psiroky:
@mswiderski could you please take a look? I believe this should be ready to go.

mswiderski:
this might fix kie server itself but won't fix controller/workbench which the jira is actually about see this stack trace https://issues.jboss.org/secure/attachment/12426417/stacktrace.txt

So please make sure that it will work with workbench... which I believe it won't as by excluding jackson from EAP/WildFly module it will essentially break workbench REST API. But I might be wrong...

mareknovotny:
@mswiderski is there a integration test for that? I ran the mentioned kie-server-integ-tests-common module against EAP7.1.

And excluding of jackson is just the module, but there is bundled one in the app(s). I hope it is in controller/workbench too, but I can be wrong ;)

sutaakar:
@mareknovotny Please try to run integration tests in kie-server-integ-tests-controller module. This module contains integration tests for standalone controller with Kie server.

mswiderski:
@mareknovotny I don't think we have integration test that tests it with workbench. 

The issue I see is that workbench has REST api to for example create org unit, new project, compile it etc. (https://github.com/kiegroup/appformer/blob/master/uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java#L490) And it by defaults completely rely on application server JAX-RS features meaning it will expect payload to be transformed to given object before it actually reaches the application code. So if we exclude it (jackson) on workbench side I believe it will not find the right jars or it might find those that are not compatible anyway.

So the way is to test this with workbench to be completely sure it does work. Which I think it won't as this exclusion is only on kie server side.

mareknovotny:
@sutaakar that is ran in the full test build on the PR verification or not?

mareknovotny:
@mswiderski I will try to test it manually then. Thanks for the head up

sutaakar:
@mareknovotny Yes, it is. Just PR build runs against WildFly10/EAP7.0 , it would be good to run the module also with WildFly11/EAP7.1

mareknovotny:
closing as we will use what @mswiderski sent in   https://github.com/kiegroup/droolsjbpm-integration/pull/1299

PR #1302, opened 2017-12-08
JBPM-6243 - Handle ContainerSpec and KieServerControllerDescriptorCom…

cristianonicolai:
…mand serialization

PR #1304, opened 2017-12-11
JBPM-6243 - New Kie Server Controller Management WebSocket API

cristianonicolai:


cristianonicolai:
Still contains changes from https://github.com/kiegroup/droolsjbpm-integration/pull/1302, it probably needs rebasing after that PR is merged.

kie-server-parent/kie-server-controller/kie-server-controller-websocket/src/main/java/org/kie/server/controller/websocket/management/KieServerMgmtCommandServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+            LOGGER.debug("Handler {} returned response {}",
+                         handler,
+                         result);
+            // return successful result
+            return new KieServerControllerServiceResponse(ResponseType.SUCCESS,
+                                                          "",
+                                                          result);
+        } catch (InvocationTargetException e) {
+            e.printStackTrace();
+            LOGGER.error("Failed to invoke service method",
```
mswiderski: this will print the stack trace twice, isn't it?

kie-server-parent/kie-server-controller/kie-server-controller-websocket/src/main/java/org/kie/server/controller/websocket/management/KieServerMgmtCommandServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                         e);
+            return new KieServerControllerServiceResponse(ResponseType.FAILURE,
+                                                          e.getTargetException().getMessage());
+        } catch (Throwable e) {
+            e.printStackTrace();
+            LOGGER.error("Error while processing {} command",
```
mswiderski: same here - this will print the stack trace twice, isn't it?

kie-server-parent/kie-server-controller/kie-server-controller-websocket-common/src/main/java/org/kie/server/controller/websocket/common/config/WebSocketClientConfigurationImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private String controllerUrl;
+
+    private String userName;
+
+    private String token;
+
+    private String password;
+
+    private Long maxSessionIdleTimeout = 0l;
```
sutaakar: Please use 0L for better readability. 0l can be easily misread as 01.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/resources/logback-test.xml
```diff
@@ -8,7 +8,9 @@
     </encoder>
   </appender>
 
-  <logger name="org.kie.server" level="WARN" />
```
sutaakar: Please return logging levels back to original values to reduce log size.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/RestKieControllerManagementIntegrationTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    protected void assertBadRequestException(KieServerControllerHTTPClientException e) {
+        assertEquals(400, e.getResponseCode());
+        assertNotNull(e.getMessage());
+    }
+
+    @Test
+    public void testCreateDuplicateServerTemplate() {
```
sutaakar: Is there any reason why these tests are executed just against REST endpoint? I would expect REST and WS API to behave the same.

kie-server-parent/kie-server-controller/kie-server-controller-websocket/src/test/java/org/kie/server/controller/websocket/management/KieServerMgmtCommandServiceImplTest.java
```diff
@@ -0,0 +1,219 @@
...
+        assertNotNull(response.getResult());
+        assertEquals(serverTemplate,
+                     response.getResult());
+    }
+
+    @Test
+    public void testCommandWithDomainArguments() {
+        final ContainerSpec containerSpec = new ContainerSpec();
+        containerSpec.setId("templateId");
+        containerSpec.setContainerName("containerId");
```
jakubschwan: Container id is a "templateId" and container name is "containerId"?
Use there String constants for container ID and container name.

cristianonicolai:
@mswiderski @sutaakar @jakubschwan updated PR based on your comments, please take another look.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/RestKieControllerManagementIntegrationTest.java
cristianonicolai: @sutaakar these are behaviors that currently apply only to the REST service, see _RestSpecManagementServiceImpl_ class, which does some extra validations on top of _SpecManagementServiceImpl_ service. @mswiderski wdyt, should these be moved into base service class?

mswiderski: @cristianonicolai yes, I think we need to have that validation moved to base classes to make sure it consistent and work the same way and the only thing that is different is the protocol.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
```diff
@@ -108,10 +108,6 @@ public Response saveServerTemplate(@Context HttpHeaders headers, @PathParam("id"
 
         String contentType = getContentType(headers);
         try {
-            if (super.getServerTemplate(serverTemplateId) != null) {
```
cristianonicolai: @mswiderski in this case specifically I'm changing the REST API behavior to match the base service, where the server template is updated instead of error when a server with the same id already exists.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/RestKieControllerManagementIntegrationTest.java
cristianonicolai: @sutaakar @mswiderski sure, updated PR so that both protocols have similar behavior.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
mswiderski: @cristianonicolai fine with me

PR #1308, opened 2017-12-11
RHBRMS-3044: Store template map when template file doesn't exist

sutaakar:
Even Files.write() generates 2 file events when modifying content. I suppose the first event is caused by deleting old content, second contains new content.
Adjusted loadTemplateMapsFromFile() to store actual content just in case of missing file, in case of malformed file it will just show warning.


sutaakar:
@mswiderski Updated commit message with JIRA.
Will check Kie server router watcher.

sutaakar:
@mswiderski Kie server router seems to work correctly. 2 modify file events are generated there, first one already contains completely changed file. In case of corrupted file router just logs marshalling error and waits for another file event.

mswiderski:
@sutaakar good to hear, thanks

once build is over will merge it

PR #1310, opened 2017-12-14
Exec model

mariofusco:


kie-maven-plugin-example/src/main/resources/FireAlarmKBase/alarm.drl
```diff
@@ -15,4 +15,4 @@
 
 package org.kie.sample.model
 declare Alarm
-end
\ No newline at end of file
+end;
```
tarilabs: ??

tarilabs: we can deal with this later I think

kie-maven-project-example/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+  <parent>
+    <groupId>org.drools</groupId>
+    <artifactId>droolsjbpm-integration</artifactId>
+    <version>7.6.0-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.kie</groupId>
+  <artifactId>kie-maven-project-example</artifactId>
+
+  <dependencyManagement>
```
tarilabs: Also we can deal with this later, but this I think can be omitted as it inherits from the parent which I suppose contains the BOMs

PR #1313, opened 2017-12-19
JBPM-6710 - Create comment operation should directly return comment id

mswiderski:
depends on https://github.com/kiegroup/jbpm/pull/1103


mswiderski:
Jenkins retest this

PR #1317, opened 2017-12-23
JBPM-6243 - Update WebSocket javax API version

cristianonicolai:

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-client/pom.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-client/pom.xml
@@ -40,7 +40,7 @@
 
     <dependency>
       <groupId>org.jboss.spec.javax.websocket</groupId>
-      <artifactId>jboss-websocket-api_1.0_spec</artifactId>
+      <artifactId>jboss-websocket-api_1.1_spec</artifactId>
       <scope>provided</scope>
     </dependency>
 
--- a/kie-server-parent/kie-server-controller/kie-server-controller-websocket-client/pom.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-websocket-client/pom.xml
@@ -43,7 +43,7 @@
 
     <dependency>
       <groupId>org.jboss.spec.javax.websocket</groupId>
-      <artifactId>jboss-websocket-api_1.0_spec</artifactId>
+      <artifactId>jboss-websocket-api_1.1_spec</artifactId>
       <scope>provided</scope>
     </dependency>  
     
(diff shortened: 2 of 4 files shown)


cristianonicolai:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/654

cristianonicolai:

kie-server-parent/kie-server-controller/kie-server-controller-client/pom.xml
```diff
@@ -66,7 +72,7 @@
     <dependency>
       <groupId>org.jboss.resteasy</groupId>
       <artifactId>resteasy-client</artifactId>
-      <scope>test</scope>
+      <scope>runtime</scope>
```
mswiderski: Why is this needed to be on runtime? Won't it cause issues on other app server like WAS or WLS?

cristianonicolai: @mswiderski definitely, when running inside a container must likely people would need to exclude the default implementations we provide here. On the other hand, I'm trying to make the client API self-contained, meaning that you only need to depend on the _kie-server-controller-client_ to use it.

mswiderski: @cristianonicolai and why not using jax-rs client instead that is dynamically located on the container it is running?

cristianonicolai: @mswiderski  I believe that would be the case if this dependency is excluded. I guess there are two approaches to this, we either provide all libraries needed for running our client API and in case your runtime already has some of it, people would need to exclude it via pom or we assume people are running it in an EE7 container and instruct people which libraries they should include in case they deploy using a different scenario. I opted for the later because it seemed to me that the usage is really straightforward and give us a little more control over the client implementation that is in use.

mswiderski: do we plan to include this in anything that is currently deployable to WAS or WLS? If so then we should change it to use jax-rs client as it will come back. Anyway including JEE bits in the app is usually problematic long run unless we don't run it on JEE container.

cristianonicolai: no worries, I'll revert changes in here to cover only the websocket version upgrade. Will also add instructions in the docs for standalone usage.

mswiderski: thanks, once that is done I think it's good to be merged.

cristianonicolai:

cristianonicolai:
@mswiderski updated PR to only change the Web Socket API version as well added dependency info in the docs:
- https://github.com/kiegroup/kie-docs/pull/565
- https://github.com/kiegroup/kie-docs/pull/564

mswiderski:
Jenkins retest this

PR #1325, opened 2018-01-09
[Reproducer] Test for retrieving process definitions from duplicated container

sutaakar:


mswiderski:
Jenkins retest this

mswiderski:
@sutaakar do you think we should merge it?

sutaakar:
I think it isn't needed as we have this covered in Case management tests.

mswiderski:
ok, closing then

PR #1328, opened 2018-01-11
[7.5.x] RHDM-232 - ClassNotFoundException when invoking image related operation, RHDM-226 - Print request payload into the log 

mswiderski:
@sutaakar here is back port of #1327 

@etirelli it includes another jira that just adds logging of payload for rules execution that was requested by CEE. I believe it's worth adding to RHDM 7


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1329, opened 2018-01-11
JBPM-6729: Generate correct base-path when context-root is empty (for example wh…

DuncanDoyle:
…en we deploy KIE-Server as ROOT.war, which we do in our OpenShift images).

Fixes JBPM-6729

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-swagger/src/test/java/org/kie/server/services/swagger/SwaggerKierServerExtensionTest.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-swagger/src/test/java/org/kie/server/services/swagger/SwaggerKierServerExtensionTest.java
@@ -1,5 +1,3 @@
-package org.kie.server.services.swagger;
-
 /*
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-services/kie-server-services-swagger/src/test/java/org/kie/server/services/swagger/SwaggerKierServerExtensionTest.java
```diff
@@ -0,0 +1,51 @@
+package org.kie.server.services.swagger;
```
mswiderski: missing license header

mswiderski:
@DuncanDoyle and commit message with JIRA number please :)

DuncanDoyle:
Should be good now.

mswiderski:
@DuncanDoyle license header should be first thing in the file not after package... I think so :)

DuncanDoyle:
Djeeezzz, that happens when you do 20 things at the same time ...... Fixed now.

PR #1330, opened 2018-01-11
DROOLS-2234 DMN JSON payload with decision-name or decision-id null causes NPE

tarilabs:
/cc @baldimir @etirelli @kurobako

PR #1331, opened 2018-01-12
RHDM-266 DROOLS-2234 DMN JSON payload with decision-name or ...

tarilabs:
... decision-id null causes NPE (#1330)

cherry picked from fe8a0707b392622d2707c9acd9dbae21cd039f40

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/src/test/java/org/kie/server/integrationtests/dmn/DMNInputDataStringIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/src/test/java/org/kie/server/integrationtests/dmn/DMNInputDataStringIntegrationTest.java
@@ -15,24 +15,27 @@
 
 package org.kie.server.integrationtests.dmn;
 
+import java.util.Map;
+
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.kie.dmn.api.core.DMNContext;
 import org.kie.dmn.api.core.DMNDecisionResult;
 import org.kie.dmn.api.core.DMNResult;
+import org.kie.server.api.model.KieServiceResponse.ResponseType;
 import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.ServiceResponse;
-import org.kie.server.api.model.KieServiceResponse.ResponseType;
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/src/test/java/org/kie/server/integrationtests/dmn/DMNInputDataStringIntegrationTest.java
```diff
@@ -185,4 +188,22 @@ public void testInputStringEvaluateDecisionById() {
         }
     }
     
+    @Test
+    public void testDROOLS2234() {
```
sutaakar: I think it would be better to create a test directly calling REST endpoint with malformed payload instead of bending of client. That way you can directly replicate JIRA issue.
You may get inspiration for example in https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/jbpm/rest/FormServiceRestOnlyIntegrationTest.java#L86

tarilabs: Thanks for the input, but I respectfully disagree.

I'm not bending the client, is simply a downcast to use a specific method. The end user would not see it anyway, as from the public API is always be returned with the upper interface (which does not contain the specific method used here). More generally this is what happens with the framework, and shows this should not be a supported mechanism. In this case it's a customer reported problem, where we are adding a safeguard to avoid the server to throw exception, when the request is kind of malformed --the nullification of list shown in the test.

On other side, crafting the request manually as per your link would require that IF we change the API, then all the manual crafted requests would need to be aligned, manually --isn't there a better way, provided we have now Swagger/OpenAPI definition file on the KieServer?

Given this is customer reported I would opt to go ahead with the backport as a priority, and we can follow-up on this topic with @etirelli or @mswiderski separately, what do you think @sutaakar ?

sutaakar: Ok, can go ahead for now. I can look if there is a way to use Swagger for such direct tests.

PR #1332, opened 2018-01-12
JBPM-6740 - Expose readiness and liveness checks in KIE Server

mswiderski:
@etirelli @ge0ffrey @tarilabs @sutaakar 

here is the implementation that aims at providing basic readiness and liveness checks for KIE Server and all extensions activated. 
In general: 
- readiness will either respond with 200 (OK) when it's actually ready or with 503 (Service Unavailable) when it's still booting/deploying containers/waiting for controller.
- liveness (aka health check) will perform following:
    - will check readiness
    - will check for failed kie containers
    - will ask each active extension to health check itself
response codes for health check are same as for readiness. Meaning that any error found will result in response 503, regardless if that is failed container, failed extension or not ready yet

Health check can be invoked in two modes: 
- basic - that will return status only 200 or 503
- report - will respond with both status and report in response body that will provide info like below (body can be XML or JSON)

Now I did implement some checks in extensions but not sure if the make sense or more stuff should be added or not. So please take a moment and let me know what you would like to health check in extension you deal with.

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<list-type>
    <items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>KIE Server 'managed-server' is ready to serve requests true</content>
            <content>Server is up for 0 days, 0 hours, 2 minutes, 31 seconds</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.375+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>Drools is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.375+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>jBPM is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>Case-Mgmt is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>jBPM-UI is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>OptaPlanner is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>DMN is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>Swagger is alive</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
        <items xsi:type="kie-message" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <content>Health check done in 3 ms</content>
            <severity>INFO</severity>
            <timestamp>2018-01-12T10:40:34.378+01:00</timestamp>
        </items>
    </items>
</list-type>
```

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
@@ -18,9 +18,13 @@ package org.kie.server.remote.rest.common.resource;
 import static org.kie.server.remote.rest.common.util.RestUtils.buildConversationIdHeader;
 import static org.kie.server.remote.rest.common.util.RestUtils.createCorrectVariant;
 import static org.kie.server.remote.rest.common.util.RestUtils.getContentType;
+import static org.kie.server.remote.rest.common.util.RestUtils.serviceUnavailable;
+
+import java.util.List;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
@@ -38,9 +42,11 @@ import org.kie.server.api.model.KieContainerResource;
 import org.kie.server.api.model.KieContainerResourceFilter;
 import org.kie.server.api.model.KieContainerStatusFilter;
 import org.kie.server.api.model.KieScannerResource;
+import org.kie.server.api.model.Message;
 import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.ReleaseIdFilter;
 import org.kie.server.api.model.ServiceResponse;
+import org.kie.server.api.model.Severity;
 import org.kie.server.remote.rest.common.Header;
 import org.kie.server.services.impl.KieServerImpl;
 import org.kie.server.services.impl.KieServerLocator;
@@ -224,6 +230,45 @@ public class KieServerRestImpl {
     @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
     public Response getServerState(@Context HttpHeaders headers) {
         return createCorrectVariant(server.getServerState(), headers);
+    }
+    
+    
+    
+    @ApiOperation(value="Readiness check for KIE Server that indicates that server is fully booted and ready to accept requests",
+            response=Void.class, code=200)
+    @ApiResponses(value = { @ApiResponse(code = 503, message = "Service not yet available") })
+    @GET
+    @Path("readycheck")
+    @Produces({MediaType.TEXT_PLAIN})
+    public Response readycheck(@Context HttpHeaders headers) { 
+        if (server.isKieServerReady()) {
+            return Response.status(Response.Status.OK).build();
+        }
+        return serviceUnavailable();
+    }
+    
+    @ApiOperation(value="Liveness check for KIE Server that validates both kie server and all extensions, optionally produces report",
+            response=Message.class, code=200, responseContainer="List")
+    @ApiResponses(value = { @ApiResponse(code = 503, message = "If any of the checks failed") })
+    @GET
+    @Path("healthcheck")
+    @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+    public Response healthcheck(@Context HttpHeaders headers,
+                             @ApiParam(value = "optional report flag to return detailed report of the check, defaults to false", required = false) @QueryParam("report")  @DefaultValue("false") boolean report) { 
+        List<Message> healthMessages = server.healthCheck(report);
+        
+        boolean anyfailures = healthMessages.stream().anyMatch(msg -> msg.getSeverity().equals(Severity.ERROR));
+        if (anyfailures) {
+            if (report) {
+                return createCorrectVariant(healthMessages, headers, Response.Status.SERVICE_UNAVAILABLE);
+            }
+            
+            return serviceUnavailable();
+        }
+        if (report) {
+            return createCorrectVariant(healthMessages, headers, Response.Status.OK);
+        }
+        return Response.status(Response.Status.OK).build();
     }
 
 }
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
@@ -29,8 +29,10 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
+
 import javax.naming.InitialContext;
 
+import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.appformer.maven.support.DependencyFilter;
 import org.drools.core.impl.InternalKieContainer;
 import org.kie.api.KieServices;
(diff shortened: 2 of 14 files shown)


kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
```diff
@@ -918,7 +931,81 @@ public void addServerStatusMessage(KieServerInfo kieServerInfo) {
...
+        if (report) {           
+            Message footer = new Message(Severity.INFO, "Health check done in " + (System.currentTimeMillis() - start) + " ms");
+            healthMessages.add(footer);
+        }
+        
+        return healthMessages;
+ 
+    }
+    
+    private String calculateUptime() {
```
sutaakar: Did you consider using java.time.Duration for these calculations?

mswiderski: I actually did and the problem is that it does not take into account already used parts of it, see example:

```
0 days, 0 hours, 30 minutes, 1800 seconds
0 days, 0 hours, 30 minutes, 0 seconds
```
first one is done with duration.toMinutes and duration.toMillis()/1000 as there seems to be missing toSeconds method.

Or did you have something else in mind?

sutaakar: I see. Was thinking about using some formatter to convert Duration to some string representation, but it seems that there isn't any out of the box.
Current approach seems ok.

MarianMacik: You can use plain toString method on a duration, it will generate for example PT41M29.064S. It is ISO 8601 format. In fact, Duration's toString method does it manually with div and mod too.

mswiderski: @MarianMacik correct, but it's way less readable in my opinion, especially for those that are not familiar with ISO date format.

jhrcek: There are also easy to use things [DurationFormatUtils](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/time/DurationFormatUtils.html) in commons-lang3

mswiderski: @jhrcek you rock! replaced with DurationFormatUtils, thanks!

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
```diff
@@ -225,5 +231,44 @@ public Response updateReleaseId( @Context HttpHeaders headers,
...
         return createCorrectVariant(server.getServerState(), headers);
     }
+    
+    
+    
+    @ApiOperation(value="Readiness check for KIE Server that indicates that server is fully booted and ready to accept requests",
+            response=Void.class, code=200)
+    @ApiResponses(value = { @ApiResponse(code = 503, message = "Service not yet available") })
+    @GET
+    @Path("readiness")
```
tsurdilo: Would rename this to "readycheck" (same for method name)

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
```diff
@@ -225,5 +231,44 @@ public Response updateReleaseId( @Context HttpHeaders headers,
...
+            return Response.status(Response.Status.OK).build();
+        }
+        return serviceUnavailable();
+    }
+    
+    @ApiOperation(value="Liveness check for KIE Server that validates both kie server and all extensions, optionally produces report",
+            response=Void.class, code=200)
+    @ApiResponses(value = { @ApiResponse(code = 503, message = "If any of the checks failed") })
+    @GET
+    @Path("health")
```
tsurdilo: Would rename this to "healthcheck" (same for method name)

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
```diff
@@ -225,5 +231,44 @@ public Response updateReleaseId( @Context HttpHeaders headers,
...
+            response=Void.class, code=200)
+    @ApiResponses(value = { @ApiResponse(code = 503, message = "Service not yet available") })
+    @GET
+    @Path("readiness")
+    @Produces({MediaType.TEXT_PLAIN})
+    public Response readiness(@Context HttpHeaders headers) { 
+        if (server.isKieServerReady()) {
+            return Response.status(Response.Status.OK).build();
+        }
+        return serviceUnavailable();
```
tsurdilo: I think all "check/monitoring" methods should have always OK response status and show issues as messages to user in case of error. Might be better if users build monitoring apps against these checks

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
mswiderski: done

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
mswiderski: done

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/resource/KieServerRestImpl.java
mswiderski: these endpoints are meant for machine reading rather than human - so the most important is to communicate clearly over status codes with minimal payload that's why by default they always return just status and no payload. When user wants to see what is not healthy then there is ?report=true query param that will return payload with messages regardless of the status code (both 200 and 503)

PR #1334, opened 2018-01-12
[7.5.x] JBPM-6729: Generate correct base-path when context-root is empty (for…

mswiderski:
… example wh… (#1329)

@etirelli up to you if this should be merged for 7.0 of DM. @DuncanDoyle can provide justification if needed ;)


etirelli:
@mswiderski yes, we should merge this, just can you please have a cloned ticket into RHDM project so that we can get the acks?

mswiderski:
@etirelli done https://issues.jboss.org/browse/RHDM-268

PR #1343, opened 2018-01-17
RHDM-283 - Use storage and notification service from CDI if available

cristianonicolai:


kie-server-parent/kie-server-controller/kie-server-controller-websocket/src/main/java/org/kie/server/controller/websocket/management/WebSocketKieServerMgmtControllerImpl.java
```diff
@@ -35,11 +39,27 @@
...
+    private Instance<KieServerTemplateStorage> templateStorage;
+
+    @Inject
+    private Instance<NotificationService> notificationService;
 
     @PostConstruct
     public void configure() {
         LOGGER.info("Kie Server Controller Management WebSocket service initialized");
+        try {
+            commandService.setTemplateStorage(templateStorage.get());
```
mswiderski: @cristianonicolai will this also work on tomcat for standalone controller?
tomcat does support websocket out of the box but not CDI and for standalone controller this might be an issue.... or will it simply fall back to ServiceLoader setup then?

cristianonicolai: @mswiderski ServiceLoader still applies in the command service constructor, https://github.com/kiegroup/droolsjbpm-integration/pull/1343/files#diff-99e76eadcc84b2616607c230ddbeab24R56.
In here, is just for when CDI is available, it overrides the template storage.

mswiderski: so we just need to include cdi-api in tomcat's standalone controller and should be all good, right?

cristianonicolai: @mswiderski correct and it should be in place already because this same strategy is already in place here: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-websocket/src/main/java/org/kie/server/controller/websocket/WebSocketKieServerControllerImpl.java#L50

PR #1348, opened 2018-01-22
[Do not MERGE yet] RHDM-251 Wildfly 11

mareknovotny:


mswiderski:
@mareknovotny is this going to be included in 7.6.0 community?

mareknovotny:
@mswiderski we talked with @krisv and @etirelli before and that was postponed, but now with speeding up for community release I am not sure.

I am good to go, but we need to merge all related PRs look at https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/682#issuecomment-359461205 and there is also some additional PR at https://github.com/kiegroup/jbpm/pull/1113 too.

mswiderski:
@mareknovotny I'd vote for including it if possible so we don't diverge from 7.5.x which already has it if I am not mistaken

krisv:
Reading this slightly late, but I guess we didn't plan to hold up the 7.6.0.Final for this.  It is true that this means we are not yet including the EAP 7.1 fix yet on master (like on 7.5.x) but the decision was to keep master on WildFly10 + EAP 7.0 until we had a solution for both.  It's ok to merge these PRs (if it means we can upgrade all repos to WF11 + EAP 7.1), it just didn't make 7.6.0.Final.  Or is there a reason why this should block 7.6.0.Final? @mareknovotny @mswiderski

mareknovotny:
@krisv no reason to block it, my impression was just to wait until 7.6.0.Final is done and we can merge the whole set.

PR #1353, opened 2018-01-24
DROOLS-2243_7_5_x

lanceleverich:
Backport of kie-pmml code to allow building from pmml resources in workbench


mswiderski:
@lanceleverich is this still valid or can it be closed?

PR #1371, opened 2018-02-15
RHDM-387 - Accessing DMN container info is not possible by container …

mswiderski:
…alias

@etirelli @tarilabs could you please have a look?


etirelli:
Thank you @mswiderski !

PR #1374, opened 2018-02-16
JBPM-6842 - Redesign jBPM executor to replace polling mechanism with …

mswiderski:
…more efficient approach


mswiderski:
depends on https://github.com/kiegroup/jbpm/pull/1135

PR #1377, opened 2018-02-25
JBPM-6890: Adding the customTaskID as a unique identifier for an active

mcivantos-tribalyte:
Added processInstanceDescription as a businessId

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskWithProcessDescription.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskWithProcessDescription.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,16 +18,18 @@ package org.kie.server.api.model.instance;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlElementWrapper;
 import javax.xml.bind.annotation.XmlRootElement;
 
-@XmlAccessorType(XmlAccessType.FIELD)
-@XmlRootElement(name = "task-instance")
-public class TaskInstance {
 
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlRootElement(name = "task-instance-custom")
+public class TaskWithProcessDescription {
+    
     @XmlElement(name="task-id")
     private Long id;
     @XmlElement(name="task-priority")
@@ -54,10 +56,6 @@ public class TaskInstance {
     private Date activationTime;
     @XmlElement(name="task-expiration-time")
     private Date expirationDate;
-    @XmlElement(name="task-skippable")
-    private Boolean skipable;
-    @XmlElement(name="task-workitem-id")
-    private Long workItemId;
     @XmlElement(name="task-process-instance-id")
     private Long processInstanceId;
     @XmlElement(name="task-parent-id")
@@ -66,6 +64,14 @@ public class TaskInstance {
     private String processId;
     @XmlElement(name="task-container-id")
     private String containerId;
+    @XmlElement(name="process-instance-desc")
+    private String processInstanceDescription;
+    @XmlElement(name="task-last-modification-user")
+    private String lastModificationUser;
+    @XmlElement(name="task-last-modificaiton-date")
+    private Date lastModificationDate;
+    @XmlElement(name="task-correlation-key")
+    private String correlationKey;
 
     @XmlElementWrapper(name="potential-owners")
     @XmlElement(name="task-pot-owners")
@@ -85,7 +91,7 @@ public class TaskInstance {
     @XmlElement(name="task-output-data")
     private Map<String, Object> outputData;
 
-    public TaskInstance() {
+    public TaskWithProcessDescription() {
     }
 
     public static Builder builder() {
@@ -196,22 +202,6 @@ public class TaskInstance {
         this.expirationDate = expirationDate;
     }
 
-    public Boolean getSkipable() {
-        return skipable;
-    }
-
-    public void setSkipable(Boolean skipable) {
-        this.skipable = skipable;
-    }
-
-    public Long getWorkItemId() {
-        return workItemId;
-    }
-
-    public void setWorkItemId(Long workItemId) {
-        this.workItemId = workItemId;
-    }
-
     public Long getProcessInstanceId() {
         return processInstanceId;
     }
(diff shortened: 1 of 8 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskInstance.java
```diff
@@ -64,6 +64,8 @@
     private Long parentId;
     @XmlElement(name="task-process-id")
     private String processId;
+    @XmlElement(name="process-instance-desc")
+    private String processInstanceDescription;
```
mswiderski: I don't like this one... it essentially means always a join to have this information which does not make much sense in my opinion. In other words process instance description does not belong to task instance so having it on TaskInstance type is misleading.

mcivantos-tribalyte: One of the custom queries register for the customer using the processInstanceDescription as a businessId and it's needed to find all the task related to this businesId (which is shared between several processes related to the same contract). If I don't add the processInstanceDescription to the TaskInstance model, how can we get the processInstanceDescription using the mapper I implemented and to which I added it?  (https://github.com/kiegroup/jbpm/pull/1146)

mswiderski: as described, you either create new type (TaskWithProcessInstance most likely better name would need to be used) and then map it here when using your additional jbpm service mappers. Or use the Raw mapper that will return array of arrays. Raw mapper is fine when KIE Server client is not used.

mcivantos-tribalyte: @mswiderski  I added a new type with the custom data used in my addition jbpm service mapper. Wdyt?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskWithProcessDescription.java
```diff
@@ -0,0 +1,462 @@
...
+        return outputData;
+    }
+
+    public void setOutputData(Map<String, Object> outputData) {
+        this.outputData = outputData;
+    }
+
+    @Override
+    public String toString() {
+        return "TaskInstance{" +
```
mswiderski: might be worth altering so it's not getting confused with the main TaskInstance

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/QueryDataServiceBaseTest.java
```diff
@@ -40,15 +40,15 @@
 
     @Test
     public void testLoadDefaultQueryDefinitions() {
-       Date today = new Date();
-       KieServerEnvironment.setServerId(UUID.randomUUID().toString());
+        Date today = new Date();
+        KieServerEnvironment.setServerId(UUID.randomUUID().toString());
```
mswiderski: looks like some formatting issues ...

mswiderski:
ok to test

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/QueryDataServiceIntegrationTest.java
```diff
@@ -814,21 +814,24 @@ public void testGetTaskInstancesUsingUserTaskWithModificationMapper() throws Exc
...
+                "inner join PROCESSINSTANCELOG p on t.processinstanceid = p.processinstanceid " +
+                "inner join TASKVARIABLEIMPL d on t.id=d.taskid");
         query.setTarget("CUSTOM");
         
         try {
 
             queryClient.registerQuery(query);
 
-            List<TaskInstance> tasks = queryClient.query(query.getName(), QueryServicesClient.QUERY_MAP_TASK_WITH_MODIF, 0, 10, TaskInstance.class);
+            List<TaskWithProcessDescription> tasks = queryClient.query(query.getName(), QueryServicesClient.QUERY_MAP_TASK_WITH_MODIF, 0, 10, TaskWithProcessDescription.class);
```
MarianMacik: Don't we lose coverage of a scenario with TaskInstance this way?

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
```diff
@@ -376,23 +378,23 @@ public static TaskInstance convertToTask(UserTaskInstanceDesc userTask) {
         return instance;
     }
     
-    public static TaskInstanceList convertToTaskInstanceListPO(Collection<UserTaskInstanceWithPotOwnerDesc> instances) {
+    public static TaskWithProcessDescriptionList convertToTaskInstanceListPO(Collection<UserTaskInstanceWithPotOwnerDesc> instances) {
```
MarianMacik: TaskInstanceList will be no longer used?

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/QueryDataServiceIntegrationTest.java
MarianMacik: Do we test scenario with TaskInstance entity somewhere else? Wouldn't it be better to have a separate test for TaskWithProcessDescription?

mswiderski:
Jenkins retest this

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/QueryDataServiceIntegrationTest.java
mswiderski: no, it's for special type of mapper so it's expected to use this new type

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/ConvertUtils.java
mswiderski: it's used for all other types but not the PO customised one

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/QueryDataServiceIntegrationTest.java
mswiderski: yes, TaskInstance is used whenever task are returned from other methods. Moreover, it's really driven by the last argument in the query method

MarianMacik: Thanks for all answers.

mswiderski:
@mcivantos-tribalyte please have a look at failures as they seem to be related to changes

mcivantos-tribalyte:
Take into account that this PR is related to https://github.com/kiegroup/jbpm/pull/1148

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1378, opened 2018-02-27
JBPM-6894 - User should be able to define case, process, activity lev…

mswiderski:
…el SLA

depends on https://github.com/kiegroup/jbpm/pull/1147


mswiderski:
Jenkins retest this

PR #1387, opened 2018-03-07
RHBA-365: Reproducer for Kie server

sutaakar:
@mswiderski Please take a look when you will have some time.
I was trying to reproduce issue described in RHBA-365.


mswiderski:
replaced with https://github.com/kiegroup/droolsjbpm-integration/pull/1389

PR #1389, opened 2018-03-08
JBPM-7009 - Pagination displays incorrect amount of tasks when custom…

mswiderski:
… advanced query returns duplicate records - too much filtering - related to RHBA-365

depends on https://github.com/kiegroup/jbpm/pull/1152


mswiderski:
@sutaakar could you please have a look + the one (actual fix) in jbpm repo

PR #1391, opened 2018-03-09
RHDM-478

lanceleverich:
* Added code that is necessary to have everything generated properly when a PMML asset is included in the project
* Updated kie-plugins-testing to add kie.conf telling which PMMLCompiler to use
* Updated simple-pmml.pmml to have a model that is one of the currently working models


lanceleverich:
Jenkins retest this

lanceleverich:
Jenkins retest this

mswiderski:
@mariofusco @desmax74 guys could you please have a look at this PR as it's currently droolsjbpm-integration build is broken due to this one being already merged https://github.com/kiegroup/drools/pull/1807

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -158,19 +208,53 @@ private void buildDrl() throws MojoFailureException {
...
-            KieRepository kr = ks.getRepository();
-            InternalKieModule kModule = (InternalKieModule) kr.addKieModule(ks.getResources().newFileSystemResource(sourceFolder));
-            for (InternalKieModule kmoduleDep : kmoduleDeps) {
-                kModule.addKieDependency(kmoduleDep);
-            }
+            List<File> resources = getResourceFiles(sourceFolder);
+            ResultsImpl messages = null;
+            InternalKieModule kModule = null;
+            KieProject kieProject = null;
+            if (containsPMML) {
```
mariofusco: @lanceleverich can you please clarify why we need 2 different types of build whether the kjar contains pmml or not? Can't we unify this?

lanceleverich: The primary reason that it done this way is that the BuildMojo did not make use of the KieBuilder, when compiling assets. The code that generates and compiles the Java classes, for PMML assets, is contained in the KieBuilderImpl. As we have talked about, that will change when we refactor the PMML into a service, similar to the approach taken by the DMN modules.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -158,19 +208,53 @@ private void buildDrl() throws MojoFailureException {
...
+                final KieBuilderImpl kieBuilder = new KieBuilderImpl(kfs);
+                kieBuilder.buildAll();
+                messages = (ResultsImpl)kieBuilder.getResults();
+                kModule = (InternalKieModule)kieBuilder.getKieModule();
+                List<String> classFiles = kModule.getFileNames()
+                        .stream()
+                        .filter(name -> name.endsWith(".class"))
+                        .collect(Collectors.toList());
+
+                MemoryFileSystem mfs = kModule instanceof CanonicalKieModule ?
```
mariofusco: @lanceleverich what in my opinion is not entirely correct is that you pasted parts of the executable model mojo here. I don't think this makes sense even because in case of a generated model this mojo will be completely skipped (see line 110). I'm still hoping that we could find a way to unify the build with and without pmml. Can we please have a call this afternoon (your morning) to discuss this?

mariofusco:
Replaced by https://github.com/kiegroup/droolsjbpm-integration/pull/1393

PR #1399, opened 2018-03-19
Test stabilization and adjustments

sutaakar:
The MySQL database can be started for example by using docker:
`docker run --name mysql57 -e MYSQL_ROOT_PASSWORD=my-secret-pw -e MYSQL_USER=sa -e MYSQL_PASSWORD=sa -e MYSQL_DATABASE=mydb -p 3306:3306 -d mysql:5.7 --ssl=0 --lower_case_table_names=1`


sutaakar:
Jenkins retest this

sutaakar:
@mswiderski  Can you please take a look?

sutaakar:
@mswiderski  The reason is that default JNDI name for datasource is already used by default H2 datasource. To be able to use custom datasources I need to point to different JNDI.

sutaakar:
@mswiderski It is ready for merge, can be squashed.

PR #1418, opened 2018-04-06
[do not merge yet] Skip unstable tests on Jenkins PR builder

sutaakar:
@mswiderski So far I wasn't able to figure out the cause of test failures, skipping it for PR builder.
From what I discovered the issue seems related to missing event listener when processing signal event [1]. It seems like the event listener isn't added in [2]. Also the issue seems to happen right after ProcessServiceIntegrationTest.testProcessInstanceWithTimer() [3]. Do you have any idea what can cause such behaviour?

In case we won't be able to find out the cause we can skip the test for PR builder (using this PR).

[1] https://github.com/kiegroup/jbpm/blob/f70a465dd2621f5598f043631b8eb991142e05ba/jbpm-flow/src/main/java/org/jbpm/process/instance/event/DefaultSignalManager.java#L85
[2] https://github.com/kiegroup/jbpm/blob/f70a465dd2621f5598f043631b8eb991142e05ba/jbpm-flow/src/main/java/org/jbpm/process/instance/event/DefaultSignalManager.java#L53
[3] https://github.com/kiegroup/droolsjbpm-integration/blob/6d4caaf430835e9f862e67cea4737a07ab049d2f/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/ProcessServiceIntegrationTest.java#L720

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/category/UnstableOnJenkinsPrBuilder.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/category/UnstableOnJenkinsPrBuilder.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2018 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mswiderski:
@sutaakar to be honest I don't see anything that might be related and cause it. Since it is jpa based tests we could check if the JpaSignalManager is actually finding anything for given signal - as this should find an entry and then load process instance, have you already checked this?

mswiderski:
@sutaakar how are things with it? Shall merge this PR or ignore these tests?

sutaakar:
@mswiderski Can be merged for now to make PR build more stable. I will keep trying to figure out what is wrong in the meantime.
I would like to keep those tests as they have a meaning and they are executed locally and in downstream tests.

PR #1436, opened 2018-04-17
[7.7.x] RHPAM-479 - WebSocket controller client method scan now not working

cristianonicolai:


cristianonicolai:
@jakubschwan for this one in particular, scanNow works fine as the container is upgraded, what might be an issue is that once the upgrade is performed, the container release id is overwritten to 1.0.1 instead of remaining as LATEST and only changing the resolved released id. @mswiderski might be able to comment on that but I believe best would be to handle it on a separated JIRA as it sounds like something to be fixed at the Drools level.

PR #1440, opened 2018-04-18
RHPAM-579 - Add method to retrieve all containers in a server template

cristianonicolai:


kie-server-parent/kie-server-controller/kie-server-controller-client/src/main/java/org/kie/server/controller/client/rest/RestKieServerControllerClient.java
```diff
@@ -197,6 +197,12 @@ public ContainerList getContainers(ServerInstanceKey serverInstanceKey) {
         return throwUnsupportedException();
     }
 
+    @Override
+    public ContainerList getContainers(ServerTemplate serverTemplate,
+                                       ContainerSpec containerSpec) {
+        return throwUnsupportedException();
```
mswiderski: @cristianonicolai the only question/concern is why we diverge in implementations of the clients - REST vs Websocket... do you believe REST client should not support it?

cristianonicolai: @mswiderski I agree, we should support both clients, it is on my TODO to review the REST implementation that was left behind.

cristianonicolai:
jenkins please retest this

PR #1461, opened 2018-05-04
Fix for FormServiceIntegrationTest.testGetTaskFormWithoutPermissioneV…

sutaakar:
…iaUIClientTest()

The MessageFormat adds a comma as a separator between digit groups. This makes the test failing for JMS (which returns number without digit separators) in case task id is bigger than 1000 (REST is not affected as it uses MessageFormat to generate error message).

toString() casts the Long value to the String which is used in MessageFormat without changes.


sutaakar:
jenkins retest this

PR #1462, opened 2018-05-07
Kie server test stabilizations

sutaakar:


MarianMacik:
@sutaakar maybe it would be good to backport these stabilizations to 7.7.x branch as well.

PR #1470, opened 2018-05-16
JBPM-7280: Skip removing of container info if not existing

sutaakar:


sutaakar:
@mswiderski Will ask.

PR #1473, opened 2018-05-18
RHPAM-1044 - Smart router nodes in OpenShift don't share data

mswiderski:


kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/ControllerStorageFileWatcher.java
```diff
@@ -35,12 +30,16 @@
 public class ControllerStorageFileWatcher implements Runnable {
     
     private static final Logger log = LoggerFactory.getLogger(ControllerStorageFileWatcher.class);
+    public static final String STORAGE_FILE_WATCHER_INTERVAL = "org.kie.server.controller.templatefile.watcher.interval";
+    
+    private long sleepTime = Long.parseLong(System.getProperty(STORAGE_FILE_WATCHER_INTERVAL, "30000"));
```
jhrcek: What if user makes a mistake and specifies value which is not valid long?
To make it more robuse we should probably try-catch parse exception, log warning and use default in such cases. Wdyt?

mswiderski: this should fail fast as soon as user makes a mistake and give unexpected value and thus it will fail to boot

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/storage/ControllerStorageFileWatcher.java
```diff
@@ -35,12 +30,16 @@
 public class ControllerStorageFileWatcher implements Runnable {
     
     private static final Logger log = LoggerFactory.getLogger(ControllerStorageFileWatcher.class);
+    public static final String STORAGE_FILE_WATCHER_INTERVAL = "org.kie.server.controller.templatefile.watcher.interval";
```
sutaakar: Just a nitpick, please move this constant to KieServerConstants to have them all in one place

mswiderski: let's first see if this works and then we can fine tune it :)

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1476, opened 2018-05-23
Fix CaseQueryResource.getCaseInstances API description typo

ron-murhammer:
Fix CaseQueryResource.getCaseInstances API description typo


mswiderski:
@ron-murhammer can we please combine all of these description fixes into single PR?

ron-murhammer:
@mswiderski I can though its much easier to just submit a PR as I find them. Otherwise there's a good chance I forget about them.

mswiderski:
It is. Now each of the PR needs to be executed by Jenkins with all tests which takes both resources and time. Plus in case we should back port them we have to do many and thus increase chance to forget about one etc. so I would really appreciate combining them to one

ron-murhammer:
@mswiderski Ok, I'll try to merge them all together later today.

PR #1482, opened 2018-05-28
[RHDM-594] Remove drools-pmml dependency

mariofusco:


sutaakar:
The failure seems to be related by duplicate classes. The kie-pmml artifact brings transitive dependency on ch.qos.logback:logback-classic:jar containing same classes as are in org.slf4j:slf4j-jdk14:jar used in kie-server pom.

mariofusco:
ok to test

mariofusco:
ok to test

mbiarnes:
Jenkins please retests this

PR #1483, opened 2018-06-04
Fix kie-spring and kie-aries-blueprint and kie-camel tests

baldimir:
When Spring boots, it searches for a KieModule for the test scenario it runs in. Some tests doesn't have it, so it used kie-spring module for that, which was wrong.  

@mariofusco could you please review?


baldimir:
I checked the failures, they were unrelated to this, however it looked like kie-camel needed similar fix, so I added it.

baldimir:
jenkins retest this

baldimir:
Failures not related to the changes.

PR #1486, opened 2018-06-05
RHPAM-1029: Smart Router Servers view error

karreiro:
See: https://issues.jboss.org/browse/RHPAM-1029

Before:
_before

After:
_after

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/KieServerInstanceManagerTest.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/KieServerInstanceManagerTest.java
@@ -18,6 +18,7 @@ package org.kie.server.controller.impl;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -35,6 +36,7 @@ import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.client.KieServicesClient;
 import org.kie.server.controller.api.model.runtime.Container;
+import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
 import org.kie.server.controller.api.model.spec.Capability;
 import org.kie.server.controller.api.model.spec.ContainerSpec;
 import org.kie.server.controller.api.model.spec.ProcessConfig;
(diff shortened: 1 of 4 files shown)


karreiro:
Jenkins please retest this.

karreiro:
@mswiderski I'm still checking the broken tests. But, I would really appreciate your feedback.

Thanks.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/impl/KieServerInstanceManagerTest.java
```diff
@@ -470,6 +483,58 @@ public void testUpgradeAndStartContainer() {
...
+        final ServerInstanceKey serverInstanceKey = mock(ServerInstanceKey.class);
+        final Collection<ServerInstanceKey> serverInstanceKeys = Collections.singletonList(serverInstanceKey);
+
+        when(serverTemplate.getServerInstanceKeys()).thenReturn(serverInstanceKeys);
+
+        doReturn(client).when(instanceManager).getClient(any());
+
+        final List<Container> containers = instanceManager.callRemoteKieServerOperation(serverTemplate, containerSpec, operation);
+
+        assertFalse(containers.isEmpty());
```
jomarko: I think we should assert precise size of the `containers` collection.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/test/java/org/kie/server/router/KieServerRouterTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void setup() {
+
+        when(blockingHttpExchange.getSender()).thenReturn(sender);
+        when(serverConnection.getByteBufferPool()).thenReturn(bufferPool);
+
+        kieServerRouter = new KieServerRouter();
+    }
+
+    @Test
+    public void testMakePathHandler() throws Exception {
```
jomarko: We miss negative case test method:
```java
@Test
public void testMakePathHandler_Invalid() throws Exception {
    assertInvalidPath("/@#$#");
    // ...
    // ...
}
```

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -141,6 +129,24 @@ public void start(String host, Integer port, ConfigurationListener ...listeners)
 
     }
 
+    PathHandler makePathHandler(final AdminHttpHandler adminHandler,
```
mswiderski: I guess this is more of refactoring rather fix for the issue, right?

karreiro:
Thanks for the reviews, everyone.

@mswiderski @jhrcek When I was checking the broken tests, I noticed that my previous changes had some impact in other APIs. Thus, I've changed my approach to avoid regressions.

I could fix the router API and get the same result by changing the `DefaultRestrictionPolicy.java` class:
_demo

@jomarko Changes applied.

Thanks.

karreiro:
Unrelated test failures.

PR #1489, opened 2018-06-07
Skip several tests on Jenkins as they tend to be unstable there

sutaakar:
Based on test failure analysis.
For some unknown reason these tests cause random database transaction issues when executed in community Jenkins.
@mswiderski Could it be possible to add "jenkins-pr-builder" profile to Jenkins downstream tests so these tests are skipped for downstream builds too?

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/SLAComplianceIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/SLAComplianceIntegrationTest.java
@@ -30,11 +30,13 @@ import java.util.stream.Collectors;
 
 import org.junit.BeforeClass;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 import org.kie.api.KieServices;
 import org.kie.server.api.model.ReleaseId;
 import org.kie.server.api.model.instance.NodeInstance;
 import org.kie.server.api.model.instance.ProcessInstance;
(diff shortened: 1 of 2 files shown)


baldimir:
Let's wait for this PR [1] to finish. I asked @mbiarnes  to remove -T1C from droolsjbpm-integration PR job, so we have no parallelism there for now. 

[1] https://github.com/kiegroup/jbpm/pull/1243

baldimir:
Ok so it is still failing also without T1C.

mswiderski:
@sutaakar fine with me to mark them as unstable on jenkins at the moment so at least we have more stable overall builds of this repo. Would you mind to do it?

sutaakar:
@mswiderski Well, they are marked so in this PR :)

mswiderski:
I know but is this enough for the downstream as well? or there is a need to make changes in the job configuration itself...

sutaakar:
@mswiderski The profile excluding the unstable tests rely on property "ghprbGhRepository" being a parameter of the build. This is true for PR builder builds invoked in droolsjbpm-integration repo, but downstream builds don't contain this parameter. So the option is to invoke the profile "jenkins-pr-builder" in maven command.
The purpose of all this is to skip unstable tests just on community Jenkins while keeping these tests available in local build and for QE.

mswiderski:
so this means a need to change community jenkins job configuration, and that's what I meant by asking if you could take care of it :) Or did I still misunderstand it...

sutaakar:
@mswiderski  Ok, I will take care of the change, will just figure out where to do it.....

mswiderski:
@sutaakar can we merge this or shall we wait for jenkins to be updated?

PR #1492, opened 2018-06-11
RHPAM-1149 - Synchronization issue for concurrent registrations in Sm…

mswiderski:
…art router


sutaakar:
Tried in OpenShift (2 Smart router instances and 3 Kie server instances) and it seems working, Smart router status is updated correctly on both pods when a container is added/removed to Kie servers.

mswiderski:
@sutaakar thanks for confirming, merging this one and will prepare 7.7.x branch backport

PR #1493, opened 2018-06-13
JBPM-7360 - Extend Kie Server Controller REST endpoints for RuntimeMa…

cristianonicolai:
…nagementService

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.controller.rest;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.kie.server.controller.api.KieServerControllerException;
+import org.kie.server.controller.api.KieServerControllerIllegalArgumentException;
+import org.kie.server.controller.api.model.runtime.ContainerList;
+import org.kie.server.controller.api.model.runtime.ServerInstanceKeyList;
+import org.kie.server.controller.impl.service.RuntimeManagementServiceImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.kie.server.controller.rest.ControllerUtils.*;
+
+@Path("/controller/runtime")
+public class RestRuntimeManagementServiceImpl extends RuntimeManagementServiceImpl {
+
+    private static final Logger logger = LoggerFactory.getLogger(RestRuntimeManagementServiceImpl.class);
+    private static final String REQUEST_FAILED_TOBE_PROCESSED = "Request failed to be processed due to: ";
+
+    @GET
+    @Path("servers/{id}/instances")
+    @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+    public Response getServerInstances(@Context HttpHeaders headers, @PathParam("id") String serverTemplateId) {
+        String contentType = getContentType(headers);
+        try {
+            logger.debug("Received get server template with id {}", serverTemplateId);
+            final ServerInstanceKeyList instances = super.getServerInstances(serverTemplateId);
+            String response = marshal(contentType, instances);
+            logger.debug("Returning response for get server instance with server template id '{}': {}", serverTemplateId, response);
+
+            return createCorrectVariant(response, headers, Response.Status.OK);
+        } catch (KieServerControllerIllegalArgumentException e) {
+            return createCorrectVariant(e.getMessage(), headers, Response.Status.NOT_FOUND);
+        } catch (KieServerControllerException e){
+            return createCorrectVariant(REQUEST_FAILED_TOBE_PROCESSED + e.getMessage(), headers, Response.Status.BAD_REQUEST);
+        } catch (Exception e) {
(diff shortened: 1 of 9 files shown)


kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.kie.server.controller.rest.ControllerUtils.*;
+
+@Path("/controller/runtime")
+public class RestRuntimeManagementServiceImpl extends RuntimeManagementServiceImpl {
+
+    private static final Logger logger = LoggerFactory.getLogger(RestRuntimeManagementServiceImpl.class);
+    private static final String REQUEST_FAILED_TOBE_PROCESSED = "Request failed to be processed due to: ";
+
+    @GET
+    @Path("servers/{id}")
```
mswiderski: this endpoint should be servers/{id}/instances as this is what it actually returns - server instances for given template. At the moment it misleads that it will return server template identified by id

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @GET
+    @Path("servers/{id}/containers/{containerId}")
+    @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+    public Response getServerTemplateContainers(@Context HttpHeaders headers, @PathParam("id") String serverTemplateId, @PathParam("containerId") String containerId) {
+        String contentType = getContentType(headers);
+        try {
+            logger.debug("Received get container {} for server template with id {}", containerId, serverTemplateId);
+
+            ContainerList containers = super.getServerTemplateContainers(serverTemplateId, containerId);
```
mswiderski: why this returns a list of containers? shouldn't there be only one container with given id in server template?

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+@Path("/controller/runtime")
+public class RestRuntimeManagementServiceImpl extends RuntimeManagementServiceImpl {
+
+    private static final Logger logger = LoggerFactory.getLogger(RestRuntimeManagementServiceImpl.class);
+    private static final String REQUEST_FAILED_TOBE_PROCESSED = "Request failed to be processed due to: ";
+
+    @GET
+    @Path("servers/{id}")
+    @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+    public Response getServerInstances(@Context HttpHeaders headers, @PathParam("id") String serverTemplateId) {
```
jakubschwan: From my point of view it's not clear enough, what this endpoint return. I think it should be better to rename this REST endpoint to "servers/{id}/instances".

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
```diff
@@ -0,0 +1,110 @@
...
+        } catch (Exception e) {
+            logger.error("Get containers using server template id {} and instance id {} failed due to {}", serverTemplateId, instanceId, e.getMessage(), e);
+            return createCorrectVariant("Unknown error " + e.getMessage(), headers, Response.Status.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    @GET
+    @Path("servers/{id}/containers/{containerId}")
+    @Produces({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
+    public Response getServerTemplateContainers(@Context HttpHeaders headers, @PathParam("id") String serverTemplateId, @PathParam("containerId") String containerId) {
```
jakubschwan: Same problem here. It is not clear what this endpoint return. For me it looks like, that this endpoint should return container details. Am I right?

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
cristianonicolai: Sure.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
cristianonicolai: I'm not quite sure, I thought this would be the list of containers with the same id in all server instances related to a server template. It really just replicates/exposes the following method: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/service/RuntimeManagementService.java#L28

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
cristianonicolai: Correct, it does return a list of containers. Perhaps update it to: _servers/{id}/containers/instances/{containerId}_ ?

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestRuntimeManagementServiceImpl.java
mswiderski: in that case the endpoint should be renamed to something that would make it clearer like
servers/{id}/containers/{containerId}/instances
as this essentially returns list of actual instances running, isn't it?

cristianonicolai: @mswiderski I was considering using _servers/{id}/instances/containers/{containerId}_ to keep consistency with the other endpoints. The difference here is that there is no filter regarding a particular server instance. wdty?

mswiderski: the problem I see with it is that ending it with {containerId} indicates there will be single resource returned while here we get a list so having this end with /instances makes more sense to me at least.

cristianonicolai:
@jakubschwan @mswiderski updated PR based on the last comments, please take another look.

PR #1497, opened 2018-06-14
[AF-1224] Execution UI should point to swagger, instead of endpoint

elguardian:
added capabilities information to server instance key so it can
be retrieved by the use and act acordingly

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/Capability.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/Capability.java
@@ -17,5 +17,5 @@ package org.kie.server.controller.api.model.spec;
 
 
 public enum Capability {
-    PROCESS, RULE, PLANNING
+    PROCESS, RULE, PLANNING, SWAGGER
 }
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java
@@ -179,6 +179,9 @@ public abstract class KieServerControllerImpl implements KieServerController {
             if (serverInfo.getCapabilities().contains(KieServerConstants.CAPABILITY_BRP)) {
                 capabilities.add(Capability.PLANNING.toString());
             }
+            if (serverInfo.getCapabilities().contains(KieServerConstants.CAPABILITY_SWAGGER)) {
+                capabilities.add(Capability.SWAGGER.toString());
+            }
             serverTemplate.setCapabilities(capabilities);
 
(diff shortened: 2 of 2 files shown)


mswiderski:
@elguardian why do you need this on server instance? capabilities are already available on server template and should be used from there

elguardian:
@mswiderski aren't they filtered regarding a specfic set ?
https://github.com/elguardian/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java#L173

The data sent back to the client 
https://github.com/elguardian/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java#L227

is not the template thought; so I thought those two arguments were strong enough to add the capabilities in there.

mswiderski:
@elguardian they are at the moment but this seems like a good reason to extend it and add swagger/docs capability there too. That does not have to mean it needs to be presented on UI but since you already require this information to show the link or not it might be worth to use that instead of duplicating it on each server instance

elguardian:
ok to test

elguardian:
@mswiderski do you mind to check whethere the pr is correct or not ? 
https://github.com/kiegroup/kie-wb-common/pull/1900

mswiderski:
@elguardian looks ok though I'd question if the url change is actually a good idea. Main reason is that swagger docs are built as extension that can be easily disabled and by that break the link completely - while the server endpoint always exists. 

so what's the reason for this URL change?

elguardian:
@mswiderski the link to the docs is only shown if the swagger functionality is active and sent by the server. That is the main reason to put this part of the code here, so the UI can decide whether to show the link to the docs or not. 

https://github.com/kiegroup/kie-wb-common/pull/1900/files#diff-1946eafb6f0f208021cf921b473abff7R162

It makes sense to me in the sense that the user is in a UI so you are jumping from UI -> to swagger (which is easier to manipulate from the UI point of view rather getting machine output only useful for rest clients more likely) IMHO.

elguardian:
I guess @porcelli   has something to say as he is the reporter of https://issues.jboss.org/browse/RHPAM-1146

mswiderski:
@elguardian @porcelli frankly speaking I am not convinced this change (of the URL) makes sense - it will cause inconsistence and make people confused as it will behave differently based on the extensions enabled. So I'd opt to stick to what we have at the moment. 
What's the reason to change this URL?

porcelli:
@mswiderski this request came up during enablement as a UX improvement. as, if enabled, the link of Swagger docs makes more sense then just the REST endpoint itself.

With that said, maybe ask for UX support to address @mswiderski concern on how display:
 - just the endpoint URI
 - both endpoint URI and Swagger

mswiderski:
@porcelli thanks and I agree we should first check with UX team to make sure this is not opposite of improvement

porcelli:
@wrnkld could you help us here?

wrnkld:
Added some comments on RHPAM-1146.

PR #1498, opened 2018-06-18
JBPM-7340: Added Swagger Info to our extension. Info field (actually …

DuncanDoyle:
…the title) is a required field in a Swagger definition.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-swagger/src/main/java/org/kie/server/services/swagger/SwaggerKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-swagger/src/main/java/org/kie/server/services/swagger/SwaggerKieServerExtension.java
@@ -38,6 +38,7 @@ import io.swagger.jaxrs.config.DefaultJaxrsScanner;
 import io.swagger.jaxrs.config.JaxrsScanner;
 import io.swagger.jaxrs.config.SwaggerContextService;
 import io.swagger.jaxrs.config.SwaggerScannerLocator;
+import io.swagger.models.Info;
 
 public class SwaggerKieServerExtension implements KieServerExtension {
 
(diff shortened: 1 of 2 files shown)


mswiderski:
I'd actually skip version completely as we do not version the API at all

kie-server-parent/kie-server-services/kie-server-services-swagger/src/main/java/org/kie/server/services/swagger/SwaggerKieServerExtension.java
```diff
@@ -69,27 +70,34 @@ public void init(KieServerImpl kieServer, KieServerRegistry registry) {
...
+    	beanConfig.getSwagger().setInfo(getInfo());
+    	beanConfig.setScan(true);
+		
     	initialized = true;
     }
+    
+    private Info getInfo() {
+    	Info info = new Info();
+    	info.setTitle("KIE-Server API");
+    	info.setVersion("7.8.0");
```
mswiderski: can we remove the version completely?

MarianMacik:
jenkins retest this

kie-server-parent/kie-server-services/kie-server-services-swagger/src/main/java/org/kie/server/services/swagger/SwaggerKieServerExtension.java
DuncanDoyle: Not really .... it's a required field in the specification: https://swagger.io/specification/#infoObject

The whole point of adding the "Info" object was to be spec compliant ... so removing the version field would defy the purpose.

Also, I still think this should define the version of the API, not the version of KIE-Server. It should indicate versioning of the RESTful interface. So, if we patch stuff -> minor version, if we add stuff or make backwards compatible changes -> minor, and backwards incompatible changes -> major.

Would this be something we could infer from the JAX-RS annotations? For example during build time?

mswiderski: in that case I'd say we should use major version of the project/product as the version of the API - in this case version would be set to 7. wdyt?

MarianMacik: @DuncanDoyle We use revapi for notifications of API changes for example so we can infer it from there. But in general API versioning should be the same as project one I think. All minor releases should or even must be backward-compatible, majors don't need to. So if we stick to this, I agree with @mswiderski.

mswiderski: changed the version to be 7.0 - in general refers to major version of the project but uses minor to accommodate unexpected situation

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

MarianMacik:
Jenkins retest this

mswiderski:
Jenkins retest this

zyfAnd:
@mswiderski Could you provide method jBPM project how to call external Restful API service(eg: springboot Restful API )? thankls a lot

PR #1505, opened 2018-07-03
RHPAM-1332: Bad link for subprocess instances in Business Central pro…

jesuino:
RHPAM-1332: Bad link for subprocess instances in Business Central process SVG


mswiderski:
@jesuino could you please add a test case?

mswiderski:
replaced with https://github.com/kiegroup/droolsjbpm-integration/pull/1562

PR #1509, opened 2018-07-11
JBPM-7068: Stabilize RuntimeDataServiceIntegrationTest after the sort…

sutaakar:
…ing changes

Also make the tests more resilient.


sutaakar:
Updated process names to distinguish SLA processes and make sorting reliable.

mswiderski:
failures not related and should be addressed with https://github.com/kiegroup/droolsjbpm-integration/pull/1511

PR #1512, opened 2018-07-11
RHPAM-1240 - Aborting process in not existing container

mswiderski:
@sutaakar here is the fix and your test case


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1515, opened 2018-07-12
AF-799

desmax74:
Contains the new Takari version and some cleanup in code used with the new compiler,  the code is only on the maven compiler and not the integration of the workbench

It depends on:
1)https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/781
2)https://github.com/kiegroup/appformer/pull/417


desmax74:
@porcelli @jhrcek

kie-takari-plugin/pom.xml
```diff
@@ -26,6 +26,7 @@
       <plugin>
         <groupId>io.takari.maven.plugins</groupId>
         <artifactId>takari-lifecycle-plugin</artifactId>
+        <version>${version.io.takari.maven.plugins}</version>
```
porcelli: not needed

kie-maven-plugin/pom.xml
```diff
@@ -26,6 +26,7 @@
       <plugin>
         <groupId>io.takari.maven.plugins</groupId>
         <artifactId>takari-lifecycle-plugin</artifactId>
+        <version>${version.io.takari.maven.plugins}</version>
```
porcelli: not needed

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -247,72 +251,73 @@ private void saveFile(MemoryFileSystem mfs, String fileName) throws MojoFailureE
         }
     }
 
-    private void shareKieObjectsWithMap(InternalKieModule kModule) {
-        Optional<Map<String, Object>> optionalKieMap = getKieMap();
-        if (optionalKieMap.isPresent()) {
-            KieMetaInfoBuilder builder = new KieMetaInfoBuilder(kModule);
-            KieModuleMetaInfo modelMetaInfo = builder.getKieModuleMetaInfo();
+    private String getCompilationID(Map<String, Object> optionalKieMap) {
```
desmax74: @jhrcek we can't test the kie-map here (in this prj) because is always null when the kie-maven-plugin is invoked without the compiler in the kie-wb-common

jhrcek: @desmax74 The reason why it can't be tested here is because it's part of one big monolithic function.
You can extract the logic that modifies the input kieMap into separate package private method and test that method separately. Something like
```
               if (optionalKieMap.isPresent()) {
                    Map<String, Object> kieMap = optionalKieMap.get();
                    addStuffToKieMap(kieMap, kModule);
                } else {
                    getLog().info("Kie Map not present");
                }
...

   // create new package private so it can be tested separately
    void addStuffToKieMap(...) {
                    String compilationID = getCompilationID(kieMap);
                    shareKieObjectsWithMap(kModule, compilationID, kieMap);
                    shareStoreWithMap(kModule.getModuleClassLoader(), compilationID, kieMap);
                    shareTypesMetaInfoWithMap(kModule, compilationID, kieMap);
     }
```
There is a lot of logic and branching in this parameter-processing code and it should be unit tested. This way we'll cover all possible cases easily in fast-running tests, which you be much harder to achieve in integration tests.

desmax74: @jhrcek Moved the logic used by the compiler into CompilerHelper class

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -180,20 +175,30 @@ private void buildDrl() throws MojoFailureException {
             ResultsImpl messages = (ResultsImpl)kieBuilder.getResults();
 
             List<Message> errors = messages != null ? messages.filterMessages( Message.Level.ERROR): Collections.emptyList();
+
+            if (container != null) {
+                Optional<Map<String, Object>> optionalKieMap = getKieMap();
```
porcelli: Don't use Optional for Collections, reference: https://stackoverflow.com/a/26328555

kie-maven-plugin/src/main/java/org/kie/maven/plugin/CompilerHelper.java
```diff
@@ -0,0 +1,92 @@
...
+import org.drools.compiler.kie.builder.impl.FileKieModule;
+import org.drools.compiler.kie.builder.impl.InternalKieModule;
+import org.drools.compiler.kie.builder.impl.KieMetaInfoBuilder;
+import org.drools.core.common.ProjectClassLoader;
+import org.drools.core.rule.KieModuleMetaInfo;
+import org.drools.core.rule.TypeMetaInfo;
+
+//Contains the logic used by the Maven compiler in the BuildMojo to permits better tests
+
+class CompilerHelper {
```
porcelli: tests?

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
desmax74: Fixed

kie-maven-plugin/src/main/java/org/kie/maven/plugin/CompilerHelper.java
desmax74: Here : https://github.com/kiegroup/droolsjbpm-integration/pull/1515/files#diff-eb21c97169b7b9c513a6e385d5968df5

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -247,76 +275,21 @@ private void saveFile(MemoryFileSystem mfs, String fileName) throws MojoFailureE
...
              */
             Map<String, Object> kieMap = (Map) container.lookup(Map.class,
-                    "java.util.HashMap",
-                    "kieMap");
-            return Optional.of(kieMap);
+                                                                "java.util.HashMap",
+                                                                "kieMap");
+            return Optional.ofNullable(kieMap).orElse(Collections.emptyMap());
         } catch (ComponentLookupException cle) {
             getLog().info("kieMap not present with compilationID and container present");
```
jhrcek: Shouldn't this be warning instead of info message?
The message "kieMap not present with compilationID and container present" is a bit confusing
Is the kie map not present? Or is it present but doesn't contain complationID? 
Imagine that you are a user who sees this log message. If the message is good he'll immediately know what's wrong and what and where should be fixed. Otherwise he'll have to look at the source code to understand what's wrong :(

Also there is probably duplication of log messages on line 194: `getLog().info("Kie Map not present");` - we should probably only log this at one place.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/CompilerHelper.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.maven.plugin.logging.Log;
+import org.drools.compiler.kie.builder.impl.FileKieModule;
+import org.drools.compiler.kie.builder.impl.InternalKieModule;
+import org.drools.core.common.ProjectClassLoader;
+import org.drools.core.rule.KieModuleMetaInfo;
+import org.drools.core.rule.TypeMetaInfo;
+
+class CompilerHelper {
+
+    public String getCompilationID(Map<String, Object> optionalKieMap, Log log) {
```
jhrcek: Nitpick: parameter name - after the refactoring, the map is no longer optional :)

kie-plugins-testing/src/test/java/org/kie/maven/plugin/CompilerHelperTest.java
```diff
@@ -0,0 +1,518 @@
...
+
+        KieBaseModel kieBaseModel2 = kproj.newKieBaseModel(namespace + ".KBase2")
+                .setEqualsBehavior(EqualityBehaviorOption.IDENTITY)
+                .addPackage(namespace + ".KBase2")
+                .setEventProcessingMode(EventProcessingOption.CLOUD);
+
+        kieBaseModel2.newKieSessionModel(namespace + ".KSession3")
+                .setType(KieSessionModel.KieSessionType.STATEFUL)
+                .setClockType(ClockTypeOption.get("pseudo"));
+/*
```
jhrcek: Let's not merge commented-out code, please.

kie-plugins-testing/src/test/java/org/kie/maven/plugin/CompilerHelperTest.java
```diff
@@ -0,0 +1,518 @@
...
+        kieMap.get(sbTypes.toString());
+        Map store = (Map) kieMap.get(sbTypes.toString());
+        assertThat(store).isNull();
+    }
+
+    @Test
+    public void shareWrongStoreWithMapTest() throws Exception {
+        //prepare
+        CompilerHelper helper = new CompilerHelper();
+        Map<String, Object> kieMap = new HashMap<>();
```
jhrcek: Just so you know: you can create a map with single key-value pair using single statement
```
Map<String,Object> kieMap = Collections.singletonMap("compilation.ID", compilationID);
``` 
to remove some boilerplate..

kie-plugins-testing/src/test/java/org/kie/maven/plugin/CompilerHelperTest.java
```diff
@@ -0,0 +1,518 @@
...
+import org.kie.api.definition.type.Role;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.StatelessKieSession;
+import org.kie.api.runtime.conf.ClockTypeOption;
+
+import static ch.qos.logback.core.encoder.ByteArrayUtil.hexStringToByteArray;
+import static ch.qos.logback.core.encoder.ByteArrayUtil.toHexString;
+import static org.assertj.core.api.Assertions.*;
+
+public class CompilerHelperTest {
```
jhrcek: Nice coverage :+1:

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
desmax74: I'll change with more clear messages

kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
```diff
@@ -180,27 +175,60 @@ private void buildDrl() throws MojoFailureException {
             ResultsImpl messages = (ResultsImpl)kieBuilder.getResults();
 
             List<Message> errors = messages != null ? messages.filterMessages( Message.Level.ERROR): Collections.emptyList();
+
+            if (container != null) {
+                Map<String, Object> kieMap = getKieMap();
+                if (!kieMap.isEmpty()) {
+                    CompilerHelper helper = new CompilerHelper();
+                    String compilationID = helper.getCompilationID(kieMap, getLog());
+                    shareKieObjects(kModule,
```
porcelli: the use of internal `share*` methods and the use of helper.share* is confuse here. Why not improve and do something like:
helper.shareData(this)

desmax74: Ok, was an attempt to preserve the infos with methodNames about the object shared

kie-plugins-testing/src/test/java/org/kie/maven/plugin/CompilerHelperTest.java
```diff
@@ -0,0 +1,500 @@
...
+                                 compilationID,
+                                 kieMap,
+                                 log);
+
+        //verify
+        StringBuilder sbTypes = new StringBuilder(compilationID).append(".").append("ProjectClassloaderStore");
+        kieMap.get(sbTypes.toString());
+        Map store = (Map) kieMap.get(sbTypes.toString());
+        assertThat(store).isNotNull();
+        assertThat(store.size()).isEqualTo(1);
```
jakubschwan: Can be used `assertThat(store).hasSize(1);`

desmax74: Fixed

PR #1517, opened 2018-07-13
RHPAM-1382: Reproducer for config file watcher issue

sutaakar:
@mswiderski AFAIK it could be fixed by moving catch block from https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/repository/ConfigFileWatcher.java#L87 into the while loop, however it makes me thinking if we want to handle all IOException like this or just the NoSuchFileException? What do you think about it?


sutaakar:
Jenkins retest this

mswiderski:
closing this one as the reproducer is included in https://github.com/kiegroup/droolsjbpm-integration/pull/1523#pullrequestreview-139369060

PR #1518, opened 2018-07-16
Increase version of pax-exam

jiripetrlik:
It is necessary to use new version of pax-exam because Fuse 7.0 does not run with the older version.

### DIFF
--- a/kie-osgi/kie-karaf-itests/pom.xml
+++ b/kie-osgi/kie-karaf-itests/pom.xml
@@ -12,6 +12,7 @@
   <name>KIE :: Karaf Integration Tests</name>
 
   <properties>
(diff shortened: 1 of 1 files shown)


jiripetrlik:
@mariofusco 
Can you please merge this PR. It seems that Fuse 7.0 does not run with older version of pax exam. It fails with following error: "Error: Could not find or load main class org.apache.karaf.main.Main".

mariofusco:
@jiripetrlik it seems that there are some test failures and I think that they could be fixed with the work I started doing on this (see "Fuse integration challenges" thread). Tomorrow I'll restart working on the osgi integration following the advices sent by Guillaume. I'll keep you updated and I'll eventually ping you on IRC if I'll have any doubt. Thanks.

jiripetrlik:
Fixed as part of ENTESB-7450. I'm closing this PR.

PR #1519, opened 2018-07-16
[7.7.x] Increase version of pax-exam

jiripetrlik:
It is necessary to use new version of pax-exam because Fuse 7.0 does not run with the older version.


jiripetrlik:
Fixed as part of ENTESB-7450. I'm closing this PR.

PR #1522, opened 2018-07-18
JBPM-7454 - Can't use custom jpa entity and jpa repository when using…

mswiderski:
… kie-server-spring-boot-starter

@tsurdilo @MarianMacik  could you please have a look guys?

### DIFF
--- a/kie-spring-boot/kie-spring-boot-samples/jbpm-spring-boot-sample-basic/src/test/java/org/jbpm/springboot/samples/ExtendedEMFBusinessProcessTest.java
+++ b/kie-spring-boot/kie-spring-boot-samples/jbpm-spring-boot-sample-basic/src/test/java/org/jbpm/springboot/samples/ExtendedEMFBusinessProcessTest.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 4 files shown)


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

kie-spring-boot/kie-spring-boot-samples/jbpm-spring-boot-sample-basic/src/test/java/org/jbpm/springboot/samples/ExtendedEMFBusinessProcessTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @After
+    public void cleanup() {
+
+        deploymentService.undeploy(unit);
+    }
+ 
+    @Test
+    public void testProcessStartAndAbort() {
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("employee", "john");
```
MarianMacik: Just to clarify. If we wanted to use Person entity here (as a process parameter), would we have to include the Person.java class in evaluation kjar as well?

mswiderski: no, it's not needed, since this is a custom app and single classpath process can get access to that entity via application classpath and same entity manager can be used without the need to setup new entity manager factory in the jpa marshalling strategy

MarianMacik: OK, so when is the classpath separate? Only in KIE Server scenario since with KIE Server you communicate remotely? With plain engine (including jUnit engine tests) it is not needed?

mswiderski:
Jenkins retest this

kie-spring-boot/kie-spring-boot-samples/jbpm-spring-boot-sample-basic/src/test/java/org/jbpm/springboot/samples/ExtendedEMFBusinessProcessTest.java
mswiderski: it's only needed if you don't put this on kie server classpath which usually you don't want to do as this requires unpacking kie server war file.

PR #1523, opened 2018-07-20
[RHPAM-1382] File watcher doesn't work in case of missing watched file

elguardian:
Using watchService to improve configuration change detection, creation and deletion


mswiderski:
failure not related

PR #1526, opened 2018-07-23
[7.7.x] [RHPAM-1382] File watcher doesn't work in case of missing watched file

elguardian:
allow the inexistance of configuration file during bootstrap, and update polling
allow some io errors during configuration file update


mswiderski:
failures not related

PR #1534, opened 2018-07-27
[RHPAM-1354] Kie server loses connection to the Business Central

elguardian:
adding the proper message handler during reconnection

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-websocket-client/src/main/java/org/kie/server/controller/websocket/client/WebSocketKieServerControllerImpl.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-websocket-client/src/main/java/org/kie/server/controller/websocket/client/WebSocketKieServerControllerImpl.java
@@ -58,14 +58,11 @@ public class WebSocketKieServerControllerImpl implements KieServerController, Ki
 
     public WebSocketKieServerControllerImpl() {
         this.marshaller = MarshallerFactory.getMarshaller(MarshallingFormat.JSON, this.getClass().getClassLoader());
-        
+
         this.client = new KieServerMessageHandlerWebSocketClient((WebSocketClient client) -> {
             try {
                 ((KieServerMessageHandlerWebSocketClient) client).sendTextWithInternalHandler(serialize(serverInfo),
-                                                                                              message -> {
-                                                                                                  logger.info("Successfully reconnected");
-                                                                                                  return null;
-                                                                                              });
+                        new WebSocketKieServerControllerReconnectHandler(context));
             } catch (IOException e) {
                 logger.warn("Error when trying to reconnect to Web Socket server - {}", e.getMessage());
             }
@@ -105,7 +102,6 @@ public class WebSocketKieServerControllerImpl implements KieServerController, Ki
                             if (received && kieServerSetup.getContainers() != null) {
                                 // once there is non null list let's return it
                                 return kieServerSetup;
-    
                             }
     
                             break;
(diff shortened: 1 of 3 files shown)


elguardian:
@mswiderski 
During startup the kieServerInfo is passed to the 
container manager, but in case of reconnection not really sure what to do. wdyt ?

mswiderski:
@elguardian I'd say that we should simply ignore the the reply as long as it's successful response for the reconnect request. Server is still fully functional and already configured so there is no need to perform any operations on kie server side.

@cristianonicolai would you mind to take a look too?

kie-server-parent/kie-server-controller/kie-server-controller-websocket-client/src/main/java/org/kie/server/controller/websocket/client/WebSocketKieServerControllerImpl.java
```diff
@@ -61,12 +61,16 @@ public WebSocketKieServerControllerImpl() {
         
         this.client = new KieServerMessageHandlerWebSocketClient((WebSocketClient client) -> {
             try {
+                KieServerSetup kieServerSetup = new KieServerSetup();
+                CountDownLatch waitLatch = new CountDownLatch(1);
                 ((KieServerMessageHandlerWebSocketClient) client).sendTextWithInternalHandler(serialize(serverInfo),
-                                                                                              message -> {
-                                                                                                  logger.info("Successfully reconnected");
-                                                                                                  return null;
```
cristianonicolai: @elguardian I think you can probably skip the setup handler and in here just return `return new CommandScriptMessageHandler(context)` that way there is a handler in place capable of handling future requests.

elguardian:
@cristianonicolai just about creating the integration test where there is a connection controller <-> kie server with web sockets I had a look at 
kie-server-integ-tests-controller
but don't see which one I should based for that, some guidance ?
I also need to understand TestConfig.isLocal

cristianonicolai:
@elguardian you could try the KieControllerStartupIntegrationTest class and make use of the controller `.startKieController();`  `.stopKieController();` operations. @sutaakar  might be able to give some insights too.

sutaakar:
@cristianonicolai @elguardian Unfortunately Kie server integration tests aren't suited well for tests requiring staring/stopping of deployables. This behaviour can be better tested in cloud test suite using OpenShift. There is already a test simulating this behaviour, currently skipped because of this issue - https://github.com/kiegroup/kie-cloud-tests/blob/master/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/persistence/WorkbenchPersistenceIntegrationTest.java.
I think it could be enough to test the behaviour here just using unit tests.

elguardian:
@sutaakar I don't think is suitable a unit test because you really need to open/close the websocket to check the proper behaviour. You need to check that the current protocol / code works properly in case of reconnection and it is not broken in the future. Don't think you can do that with a unit test.

sutaakar:
The fix itself works (tried locally on EAP).
However the test is failing - test suite is currently not capable of stopping controller and starting it again.

elguardian:
@sutaakar command you are using for running the integration test?

sutaakar:
@elguardian mvn clean install -Pwildfly11
Possibly add -Dit.test=... when you want to run single test class

elguardian:
the integration test cannot be done this way. I removed the integration test from the commit.

sutaakar:
@elguardian Is there something blocking this PR to be ready?

elguardian:
@sutaakar no AFAIK

sutaakar:
@cristianonicolai Is this PR ok for you?

elguardian:
added the unit test @sutaakar  @cristianonicolai

PR #1540, opened 2018-08-09
[RHPAM-1230] [DROOLS-2886] Add KIE features for Camel to KIE features file

jiripetrlik:


jiripetrlik:
@mswiderski Yes, you are right. It may cause failing workitem test in PR build. It would be also good to consider changing of GAV. See RHPAM-1456.

PR #1561, opened 2018-09-06
JBPM-7724 - Unable to find JTA transaction when running LogCleanupCom…

mswiderski:
…mand

depends on 
https://github.com/kiegroup/jbpm/pull/1319

@sutaakar here is a test case for just the basic execution of LogCleanupCommand in spring boot


mswiderski:
Jenkins retest this

sutaakar:
@mswiderski I have a suspicion that the failure is related to container creation timeout, I think the default timeout (60s) is too small for Optaplanner kjar.

PR #1564, opened 2018-09-06
JBPM-7727 - SecurityException when SLA is triggered in case

mswiderski:
@sutaakar I believe that is what we need to get aligned


mswiderski:
Jenkins retest this

PR #1575, opened 2018-09-18
[JBPM-7762] Process Instance Migration integration

ruromero:
Fix https://issues.jboss.org/browse/JBPM-7762

Things to consider:
* Overridden mysql version due to an initialisation problem
* Tests are failing, need some advice. Shall we use arquillian?


kiereleaseuser3:

kiereleaseuser3:

kiereleaseuser2:

kiereleaseuser3:

ruromero:
@mswiderski  @MarianMacik I added the README.md file

process-migration-service/src/test/java/org/kie/processmigration/service/util/ConfigurationValueProducer.java
```diff
@@ -0,0 +1,28 @@
...
+import org.yaml.snakeyaml.Yaml;
+
+public class ConfigurationValueProducer {
+
+    Map<String, String> configuration;
+
+    @SuppressWarnings("unchecked")
+    public ConfigurationValueProducer() throws FileNotFoundException {
+        Yaml yaml = new Yaml();
+        configuration = yaml.loadAs(this.getClass().getClassLoader().getResourceAsStream("project-defaults.yml"), Map.class);
```
MarianMacik: When I tried to run this and print out the map, I found out that it is actually a LinkedHashMap with values of LinkedHashMap etc. so values cannot be typed to String in runtime. This should be probably done differently.

ruromero: Yeah, the tests are totally screwed up. I will come back to this as soon as possible

process-migration-service/src/main/java/org/kie/processmigration/rest/MigrationResource.java
```diff
@@ -0,0 +1,121 @@
...
+        } else {
+            return Response.ok(migrationService.submit(migration, credentials)).build();
+        }
+    }
+
+    @PUT
+    @Path("/{id}")
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response update(@PathParam("id") Long id, MigrationDefinition migrationDefinition) {
+        // TODO: Support reschedule migrations
```
bmozaffa: Just making sure the TODO is on your radar here

process-migration-service/src/main/java/org/kie/processmigration/serializer/DateDeserializer.java
```diff
@@ -0,0 +1,46 @@
...
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+
+public class DateDeserializer extends StdDeserializer<Date> {
```
bmozaffa: Any reason you wouldn't use the newer and superior java.time package?

process-migration-service/src/main/java/org/kie/processmigration/service/CredentialsProviderFactory.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        if (authHeader.startsWith(CredentialsProvider.BASIC_AUTH_PREFIX)) {
+            logger.debug("Extracting username/password for Scheme: Basic");
+            String encoded = authHeader.substring(CredentialsProvider.BASIC_AUTH_PREFIX.length());
+            String credentials = new String(Base64.getDecoder().decode(encoded));
+            if (isBlank(credentials)) {
+                return null;
+            }
+            String[] parts = credentials.split(CREDENTIALS_SEPARATOR);
+            if (parts != null && parts.length == 2) {
```
bmozaffa: parts can't be null

process-migration-service/src/test/java/org/kie/processmigration/service/util/ConfigurationValueProducer.java
ruromero: > Regarding the authentication, the way I understand it is that we just store credentials on the migration service side and then we will forward them to KIE Server. So plans, migration and migrationdefinition can be created without any authentication? And currently it can connect to just one KIE Server, right?

Yes, I don't consider authentication is required at the Migration service level to avoid re-implementing all the auth functionalities already in place for the workbench and Kie server until a common SSO functionality is provided. WDYT?

process-migration-service/src/main/java/org/kie/processmigration/model/Execution.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.kie.processmigration.model;
+
+import java.net.URI;
+import java.time.LocalDateTime;
```
bmozaffa: If you care about timezones, the risk of pods running in different regions and your logic being sensitive to the actual timestamp, consider using DateTime instead of LocalDateTime.

ruromero:
@MarianMacik I fixed the tests. Is there anything else that requires changes?
Thanks

process-migration-service/src/main/java/org/kie/processmigration/rest/MigrationResource.java
```diff
@@ -0,0 +1,131 @@
...
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response submit(@Context HttpHeaders headers, MigrationDefinition migration) {
+        Credentials credentials = CredentialsProviderFactory.getCredentials(headers.getHeaderString(HttpHeaders.AUTHORIZATION));
+        if (credentials == null) {
+            return Response.status(Status.UNAUTHORIZED).header(HttpHeaders.WWW_AUTHENTICATE, "Basic").build();
+        }
+        if (ExecutionType.ASYNC.equals(migration.getExecution().getType())) {
+            return Response.accepted(migrationService.submit(migration, credentials)).build();
+        } else {
+            return Response.ok(migrationService.submit(migration, credentials)).build();
```
mswiderski: can we move up 

```
migrationService.submit(migration, credentials)
```

 to be before the if else of execution type as it seems to be doing exactly same call but it's really not visible when it's done as part of building response ...and does not really differ if it is async or sync

process-migration-service/src/main/java/org/kie/processmigration/rest/MigrationResource.java
```diff
@@ -0,0 +1,131 @@
...
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response update(@Context HttpHeaders headers, @PathParam("id") Long id, MigrationDefinition migrationDefinition) {
+        Credentials credentials = CredentialsProviderFactory.getCredentials(headers.getHeaderString(HttpHeaders.AUTHORIZATION));
+        if (credentials == null) {
+            return Response.status(Status.UNAUTHORIZED).header(HttpHeaders.WWW_AUTHENTICATE, "Basic").build();
+        }
+        try {
+            Migration migration = migrationService.update(id, migrationDefinition, credentials);
+            if (migration == null) {
+                return getMigrationNotFound(id);
```
mswiderski: why not throwing this migration not found directly from the service?

process-migration-service/src/main/java/org/kie/processmigration/rest/MigrationResource.java
```diff
@@ -0,0 +1,131 @@
...
+
+    }
+
+    @DELETE
+    @Path("/{id}")
+    @Consumes({MediaType.APPLICATION_JSON})
+    public Response delete(@PathParam("id") Long id) {
+        Migration migration = migrationService.delete(id);
+        if (migration == null) {
+            return getMigrationNotFound(id);
```
mswiderski: same here, should be thrown from within the service in my opinion

process-migration-service/src/main/java/org/kie/processmigration/rest/PlanResource.java
```diff
@@ -0,0 +1,102 @@
...
+        return Response.ok(planService.findAll()).build();
+    }
+
+    @GET
+    @Path("/{id}")
+    @ApiOperation(value = "Finds a migration plan by the given plan Id")
+    public Response get(@ApiParam(value = "Plan Id") @PathParam("id") Long id) {
+        Plan plan = planService.get(id);
+        if (plan == null) {
+            return getPlanNotFound(id);
```
mswiderski: similar as for migration service ... throw the not found exception from within the service

process-migration-service/src/main/java/org/kie/processmigration/rest/PlanResource.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @PUT
+    @Path("/{id}")
+    @ApiOperation(value = "Save a migration plan")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response save(@ApiParam(value = "Plan Id to update") @PathParam("id") Long id,
+                         @ApiParam(value = "Plan") Plan plan) {
+        if (planService.get(id) == null) {
+            return getPlanNotFound(id);
```
mswiderski: same here

process-migration-service/src/main/java/org/kie/processmigration/rest/PlanResource.java
```diff
@@ -0,0 +1,102 @@
...
+        return Response.ok(planService.save(plan)).build();
+    }
+
+    @DELETE
+    @Path("/{id}")
+    @ApiOperation(value = "Delete an existing migration plan")
+    public Response delete(@ApiParam(value = "Plan Id to update") @PathParam("id") Long id) {
+        Plan plan = planService.delete(id);
+        if (plan == null) {
+            return getPlanNotFound(id);
```
mswiderski: and again

process-migration-service/src/main/java/org/kie/processmigration/rest/provider/ObjectMapperContextResolver.java
```diff
@@ -0,0 +1,49 @@
...
+@Provider
+public class ObjectMapperContextResolver implements ContextResolver<ObjectMapper> {
+
+    private final ObjectMapper mapper;
+
+    public ObjectMapperContextResolver() {
+        mapper = new ObjectMapper();
+        SimpleModule module = new SimpleModule();
+        module.addDeserializer(Instant.class, new InstantDeserializer());
+        module.addSerializer(new InstantSerializer());
```
mswiderski: any reason why not to use standard Jackson implementation for java time?

https://github.com/FasterXML/jackson-modules-java8

process-migration-service/src/main/java/org/kie/processmigration/service/CredentialsProviderFactory.java
```diff
@@ -0,0 +1,64 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/CredentialsService.java
```diff
@@ -0,0 +1,13 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/KieService.java
```diff
@@ -0,0 +1,17 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/MigrationService.java
```diff
@@ -0,0 +1,27 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/PlanService.java
```diff
@@ -0,0 +1,17 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/SchedulerService.java
```diff
@@ -0,0 +1,12 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/impl/CredentialsServiceImpl.java
```diff
@@ -0,0 +1,48 @@
+package org.kie.processmigration.service.impl;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/impl/CredentialsServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.NoResultException;
+import javax.persistence.PersistenceContext;
+import javax.transaction.Transactional;
+
+import org.kie.processmigration.model.Credentials;
+import org.kie.processmigration.service.CredentialsService;
+
+@ApplicationScoped
+public class CredentialsServiceImpl implements CredentialsService {
```
mswiderski: in my opinion this should be removed as jBPM should not store any user credentials. Instead it should be configured with admin kind of user that can talk to kie servers and that credentials should come from keystore

process-migration-service/src/main/java/org/kie/processmigration/service/impl/EjbTimerSchedulerServiceImpl.java
```diff
@@ -0,0 +1,79 @@
+package org.kie.processmigration.service.impl;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/impl/EjbTimerSchedulerServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+            timer.get().cancel();
+            scheduleMigration(migration, credentials);
+        }
+    }
+
+    @PreDestroy
+    public void stopAll() {
+        for (Timer timer : timerService.getTimers()) {
+            logger.info("Stopping timer: {}", timer.getInfo());
+            timer.cancel();
```
mswiderski: wound't this essentially cancel all scheduled migrations? Meaning if server goes down all scheduled migrations are cancelled and thus after server starts again nothing will be scheduled... or am I missing something?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,48 @@
+package org.kie.processmigration.service.impl;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        } catch (Exception e) {
+            ServiceResponse<KieServerStateInfo> response = new ServiceResponse<>();
+            response.setType(ResponseType.FAILURE);
+            response.setMsg(e.getMessage());
+            return response;
+        }
+    }
+
+    public ProcessAdminServicesClient createProcessAdminServicesClient(CredentialsProvider credentialsProvider) {
+        KieServicesConfiguration kieServicesConfig = new KieServicesConfigurationImpl(config.getUrl(), credentialsProvider);
```
mswiderski: why not using org.kie.server.client.KieServicesFactory instead of directly impl classes?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        ProcessAdminServicesClientImpl processAdminServicesClient = new ProcessAdminServicesClientImpl(kieServicesConfig);
```
mswiderski: ProcessAdminServicesClient should be taken from kieServicesClient and not instantiated manually

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        processAdminServicesClient.setOwner((KieServicesClientImpl) createKieServicesClient(credentialsProvider));
+        return processAdminServicesClient;
+    }
+
+    public KieServicesClient createKieServicesClient(CredentialsProvider credentialsProvider) {
+        KieServicesConfiguration kieServicesConfig = new KieServicesConfigurationImpl(config.getUrl(), credentialsProvider);
+        return new KieServicesClientImpl(kieServicesConfig);
```
mswiderski: same here use factory classes to create KieServicesClient instances

process-migration-service/src/main/java/org/kie/processmigration/rest/HealthCheckResource.java
```diff
@@ -0,0 +1,73 @@
...
+    public Response checkReadiness() {
+        return Response.ok(new HealthStatus().up()).build();
+    }
+
+    @GET
+    @Path("/liveness")
+    @Health
+    public Response checkLiveness(@Context HttpHeaders headers) {
+        CredentialsProvider credProvider = CredentialsProviderFactory.getProvider(headers.getHeaderString(HttpHeaders.AUTHORIZATION));
+        ServiceResponse<KieServerStateInfo> serverState = kieService.getServerState(credProvider);
```
mswiderski: don't really understand this liveness check ... first of all why liveness check of the migration service depends on availability of kie servers?
second, it calls single kie server and what if there are many kie servers that are target for scheduled migration?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,200 @@
+package org.kie.processmigration.service.impl;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+
+    @Override
+    @Transactional
+    public Migration update(Long id, MigrationDefinition definition, Credentials credentials) throws ReScheduleException {
+        Migration migration = get(id);
+        if (migration == null) {
+            return null;
+        }
+        if (!ExecutionStatus.SCHEDULED.equals(migration.getStatus())) {
+            throw new ReScheduleException("The migration is already started and cannot be re-scheduled");
```
mswiderski: maybe change the message as I find it bit confusing

```
The migration is  not scheduled and cannot be re-scheduled
```

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+            em.persist(migration.start());
+            if (credentials == null) {
+                throw new CredentialNotFoundException();
+            }
+            AtomicBoolean hasErrors = new AtomicBoolean(false);
+            //each instance id will spawn its own request to KIE server for migration.
+            List<Long> instancesIdList = migration.getDefinition().getProcessInstanceIds();
+            for (Long instanceId : instancesIdList) {
+                MigrationReportInstance reportInstance = kieService
+                                                                   .createProcessAdminServicesClient(CredentialsProviderFactory.getProvider(credentials))
```
mswiderski: since the work ins done in a synchronous loop (no multi threading) I'd create ProcessAdminServicesClient once and reuse it

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private void doCallback(Migration migration) {
+        URI callbackURI = null;
+        try {
+            callbackURI = migration.getDefinition().getExecution().getCallbackUrl();
+            Response response = ClientBuilder.newClient()
+                                             .target(callbackURI)
+                                             .request(MediaType.APPLICATION_JSON)
+                                             .buildPost(Entity.json(migration))
+                                             .invoke();
```
mswiderski: missing authentication support ... so all callbacks must be open endpoints

process-migration-service/src/main/java/org/kie/processmigration/service/impl/PlanServiceImpl.java
```diff
@@ -0,0 +1,52 @@
+package org.kie.processmigration.service.impl;
```
mswiderski: missing license header

process-migration-service/src/main/java/org/kie/processmigration/util/DateUtils.java
```diff
@@ -0,0 +1,37 @@
+package org.kie.processmigration.util;
```
mswiderski: missing license header

process-migration-service/src/main/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<persistence version="2.1"
+	xmlns="http://xmlns.jcp.org/xml/ns/persistence"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd">
+	<persistence-unit name="migration-unit"
+		transaction-type="JTA">
+		<provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
+		<!--it seems defines the jta-data-source or not it still works, so comment 
+			it out for now <jta-data-source>java:jboss/datasources/ExampleDS</jta-data-source> -->
```
mswiderski: this could be due to default data source for jpa configuration but I guess we should support custom data sources to be defined outside for other db so having more meaningful data source name than ExampleDS might be useful

process-migration-service/src/main/resources/project-defaults.yml
```diff
@@ -0,0 +1,23 @@
...
+        application-path: /
+  logging:
+    org.kie.processmigration:
+      level: INFO
+  datasources:
+    data-sources:
+      ExampleDS:
+        driver-name: mysql
+#add the ?useUnicode=true&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC to avoid timezone related exception        
+        connection-url: jdbc:mysql://localhost:3306/rhpam?useUnicode=true&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC
```
mswiderski: any reason why mysql is default? in most (if not all) cases we use H2 as default to it can be tested as part of builds

process-migration-service/src/test/java/org/kie/processmigration/jpa/converter/CryptoConverterTest.java
```diff
@@ -0,0 +1,35 @@
+package org.kie.processmigration.jpa.converter;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/serializer/InstantSerializerTest.java
```diff
@@ -0,0 +1,51 @@
+package org.kie.processmigration.serializer;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/service/CredentialsProviderFactoryTest.java
```diff
@@ -0,0 +1,88 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/service/CredentialsServiceImplTest.java
```diff
@@ -0,0 +1,89 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/service/PersistenceTest.java
```diff
@@ -0,0 +1,40 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/service/PlanServiceImplTest.java
```diff
@@ -0,0 +1,52 @@
+package org.kie.processmigration.service;
```
mswiderski: missing license header

process-migration-service/src/test/java/org/kie/processmigration/util/DateUtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.time.ZoneOffset;
+import java.time.format.DateTimeParseException;
+import java.util.Calendar;
+import java.util.Date;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
```
mswiderski: why mockito runner is needed here?

process-migration-service/src/test/java/org/kie/processmigration/util/DateUtilsTest.java
```diff
@@ -0,0 +1,54 @@
+package org.kie.processmigration.util;
```
mswiderski: missing license header

mswiderski:
one more thing is about clustered ejb timer service. How will this be configured?

process-migration-service/src/main/java/org/kie/processmigration/model/Credentials.java
```diff
@@ -0,0 +1,73 @@
...
+ * Stores the credentials to use for communicating against KIE Server.
+ * These credentials must be safely stored in the database.
+ * The migrationId is indirectly referenced to avoid dependencies
+ */
+@Entity
+@Table(name = "credentials")
+@NamedQueries({
+               @NamedQuery(name = "Credentials.findByMigrationId", query = "SELECT p FROM Credentials p WHERE p.migrationId = :id")
+})
+public class Credentials {
```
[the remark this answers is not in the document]
ruromero: It also supports tokens but yes. Do you have an alternative? If we want to interact asynchronously with kieserver we would need SSO capabilities. I know you're working on it already so we could move to that model once it is in place

process-migration-service/src/main/java/org/kie/processmigration/rest/PlanResource.java
ruromero: Same. It's not an exception but a Response object.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/CredentialsServiceImpl.java
ruromero: Then we wouldn't know who did what. Is that acceptable?

process-migration-service/src/test/java/org/kie/processmigration/util/DateUtilsTest.java
ruromero: No needed at all. It slipped

process-migration-service/src/main/resources/project-defaults.yml
ruromero: No reason at all, we just wanted to provide an example of use for mysql. For H2 there's the test-defaults.yml. In any case a user would want to provide a completely different configuration file. Maybe I should remove this file completely or move it to the test folder with more examples.

process-migration-service/src/main/resources/META-INF/persistence.xml
ruromero: It can indeed be done in the configuration file, see the examples. I will remove this comment.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
ruromero: Do you suggest to add support to basic auth?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
ruromero: Will do, thanks

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
ruromero: The `setOwner` method is not declared in the interface.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/EjbTimerSchedulerServiceImpl.java
ruromero: No, your assumption is correct. I have removed it.

process-migration-service/src/main/java/org/kie/processmigration/rest/provider/ObjectMapperContextResolver.java
ruromero: out of ignorance :grin:

process-migration-service/src/main/java/org/kie/processmigration/model/Migration.java
```diff
@@ -0,0 +1,190 @@
...
+@NamedQueries({
+               @NamedQuery(name = "Migration.findAll", query = "SELECT p FROM Migration p"),
+               @NamedQuery(name = "Migration.findById", query = "SELECT p FROM Migration p WHERE p.id = :id")
+})
+public class Migration implements Serializable {
+
+    private static final long serialVersionUID = 7212317252498596171L;
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
```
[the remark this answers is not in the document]
ruromero: Changing it to AUTO

process-migration-service/pom.xml
```diff
@@ -0,0 +1,236 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.drools</groupId>
```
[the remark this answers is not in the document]
ruromero: Fixed

process-migration-service/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Is the execution of a defined plan, applied to a set of process instances. These are the attributes that define a Migration:
+
+* Migration ID (Generated)
+* Plan ID
+* Instances to migrate. Either referenced directly or through a query (TBD).
+* Execution type
+    * Sync or Async
+    * Scheduled start time
+    * Callback URL
```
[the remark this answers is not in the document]
ruromero: The Migration object, including the migration definition and the results. The same as if you did a get to this migration resource.

process-migration-service/src/main/java/org/kie/processmigration/model/Credentials.java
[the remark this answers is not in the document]
mswiderski: kie server client has CredentialProvider interface that could be used for plugging in different implementations so depending on what we need. I'd start with keystore based password or token, same we kie server does it right now when talking to controller.

process-migration-service/src/main/java/org/kie/processmigration/model/Migration.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @JsonInclude(Include.NON_NULL)
+    @JsonProperty("cancelled_at")
+    @Column(name = "cancelled_at")
+    private Instant cancelledAt;
+
+    @JsonInclude(Include.NON_NULL)
+    @JsonProperty("error_message")
+    @Column(name = "error_message", columnDefinition = "TEXT")
+    private String errorMessage;
```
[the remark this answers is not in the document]
mswiderski: hmm this sounds bit dangerous to me as it is actually more of DDL syntax that might differ on various databases, so I'd suggest to rely on JPA annotations ... like @Log or so if we need to specify that the error message should/could be quite big

process-migration-service/src/main/java/org/kie/processmigration/rest/MigrationResource.java
ruromero: Indeed, working on it

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
ruromero: The callbackUri can include a (unique?) token e.g. http://somehost.example.com/callback?token=1sdf2342fsf

kiereleaseuser2:

kiereleaseuser2:

process-migration-service/src/main/java/org/kie/processmigration/model/KieServerConfig.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public String getUrl() {
+        return new StringBuilder(protocol)
+                                          .append("://")
+                                          .append(host)
+                                          .append(":")
+                                          .append(port)
+                                          .append(contextRoot)
+                                          .append("/services/rest/server")
```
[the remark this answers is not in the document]
ruromero: done

bmozaffa:
@mswiderski What are the required steps to get this merged?

process-migration-service/src/main/java/org/kie/processmigration/model/KieServerConfig.java
```diff
@@ -0,0 +1,111 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class KieServerConfig {
+
+    private String id;
+    private String host;
+    private Integer port;
+    @JsonProperty("context_root")
```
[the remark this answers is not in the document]
ruromero: I thought it was the "right" way to write json properties, I just found out there's no convention and the only agreement I found comes from the [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format), so moving to camelCase.

process-migration-service/src/test/resources/project-kieservers.yml
```diff
@@ -0,0 +1,8 @@
+kieservers:
+  - host: kieserver.example.com
+    port: 8080
+    protocol: http
+    contextRoot: /
```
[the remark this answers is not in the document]
ruromero: This is an example and in this example I was pointing to a containeraized kie-server deployed on `/`

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public QueryServicesClient getQueryServicesClient(String kieServerId) throws InvalidKieServerException {
+        if (!kieServices.containsKey(kieServerId)) {
+            throw new InvalidKieServerException(kieServerId);
+        }
+        return (QueryServicesClient) kieServices.get(kieServerId).getServicesClient(QueryServicesClient.class);
+    }
+
+    private void loadConfig(Map<String, String> config) {
+        KieServerConfig kieConfig = new KieServerConfig();
```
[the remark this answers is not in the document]
ruromero: Yes, in the configuration file

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private Map<String, KieServerConfig> configs = new HashMap<>();
+    private Map<String, KieServicesClient> kieServices = new HashMap<>();
+
+    @Inject
+    private ConfigView configView;
+
+    @SuppressWarnings("unchecked")
+    public void loadConfigs(@Observes @Initialized(ApplicationScoped.class) Object event) {
+        List<Map<String, String>> value = configView.resolve("kieservers").as(List.class).getValue();
```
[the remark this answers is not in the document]
ruromero: the annotations? I can change them by the `@Startup` and `@PostConstruct`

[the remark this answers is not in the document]
mswiderski: no, I asked how this config is injected here ... but that looks specific way how horntail does it, right? Can such file be outside of the fat jar?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
mswiderski: so that means to add another kie server we need to bounce the application so the config is reloaded, right? Just wondering how we can prevent users from defining migration plans for particular server that might not exist...

process-migration-service/README.md
```diff
@@ -0,0 +1,58 @@
+# Process Instance Migration Service
+
+This service is intended to be deployed as a standalone application. It can be integrated with KIE Process Server API to execute migrations remotely.
```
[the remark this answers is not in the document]
mswiderski: but that would only apply to migration plans and not migrations itself and the app is called Process Instance Migration which would indicate it will migrate and not just store plans.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
ruromero: Yes, this is specific to Thorntail. The file can (and will be) outside of the fat-jar.
With Thorntail you can provide a configuration file and/or multiple profiles (a.k.a. stages) to compose or override configurations. See multiple project-*.yml files in the project.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
ruromero: Yes, a configuration change would require the application to restart. We can work on some kind of discovery service using K8S API but this can be added later on.
In the UI it "should" be easy as the user will only be able to select from the existing instances.
Additionally maybe a "validate" or "dry-run" operations can help with that. WDYT?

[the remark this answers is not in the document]
mswiderski: I'd say the service should reject migration plans (or put them in some sort of locked state) for kie servers it does not manage

process-migration-service/README.md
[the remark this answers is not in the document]
ruromero: You have a point. Let's make it a requirement in the README.md. However, from the application point of view I will just raise a Warning if no endpoints were configured.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
ruromero: We can reject migration plans at creation/update and reject migrations when they're submitted in case the process existed when the plan was created but it was removed afterwards.

process-migration-service/src/main/java/org/kie/processmigration/service/RecoveryService.java
```diff
@@ -0,0 +1,64 @@
...
+                                                                                ExecutionStatus.CREATED);
+
+    @PersistenceContext
+    private EntityManager em;
+
+    @Inject
+    MigrationService migrationService;
+
+    @PostConstruct
+    public void resumeMigrations() {
```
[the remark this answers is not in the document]
ruromero: Will do and ignore those instances that doesn't exist in the sourceContainer

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
mswiderski: I'd say we need to make this at both places - first at creation/update time so that essentially will discard particular migration plan and on execution time it should fail the migration with proper message.

process-migration-service/src/test/resources/project-kieservers.yml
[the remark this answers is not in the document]
mswiderski: ok, is this standard practice in containerised kie servers to be deployed to root context path?

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
mswiderski: would be good to have a complete sample in the readme to let people quickly get started with the service. So we have the command to start it and maybe an sample config file with kie server running on 8080 or so.

process-migration-service/README.md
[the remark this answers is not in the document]
mswiderski: fair enough

process-migration-service/src/test/resources/project-kieservers.yml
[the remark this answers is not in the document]
bmozaffa: Normally, yes.

process-migration-service/README.md
```diff
@@ -0,0 +1,61 @@
...
+* Source and target container IDs
+* Node mappings (if any)
+
+## Migration
+
+Is the execution of a defined plan, applied to a set of process instances. These are the attributes that define a Migration:
+
+* Migration ID (Generated)
+* Plan ID
+* Instances to migrate. Either referenced directly or through a query (TBD).
```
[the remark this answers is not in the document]
ruromero: Removed

process-migration-service/README.md
```diff
@@ -0,0 +1,58 @@
...
+It is a [Thorntail 2.x](http://docs.wildfly-swarm.io/2.2.1.Final/) application
+
+```
+$ java -jar target/process-migration-thorntail.jar
+```
+
+You can provide your custom configuration file. Check [project-defaults.yml](./src/main/resources/project-defaults.yml) to see an example
+
+```
+$ java -jar target/process-migration-thorntail.jar -s./myconfig.yml
```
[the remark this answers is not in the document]
ruromero: No, it is like this. https://docs.thorntail.io/2.2.1.Final/#non-default-thorntail-yaml-configuration-files_thorntail

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+            throw new InvalidMigrationException("The Migration Definition must not be null");
+        }
+        if (definition.getPlanId() == null) {
+            throw new InvalidMigrationException("The Plan ID is mandatory");
+        }
+        if (StringUtils.isBlank(definition.getKieserverId())) {
+            throw new InvalidMigrationException("The KIE Server ID is mandatory");
+        }
+        if (!kieService.getConfigs().containsKey(definition.getKieserverId())) {
+            throw new InvalidKieServerException(definition.getKieserverId());
```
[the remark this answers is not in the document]
ruromero: It is a subclass of the InvalidMigrationException but adds information about the target KieServer used.

process-migration-service/src/test/resources/META-INF/jboss-deployment-structure.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jboss-deployment-structure>
+    <deployment>
+        <dependencies>
+            <module name="org.dom4j"/>
```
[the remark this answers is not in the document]
ruromero: This helped me avoid this error during integration tests

```
2018-11-27 23:00:17,843 ERROR [stderr] (main) Caused by: org.wildfly.swarm.container.DeploymentException: THORN0004: Deployment failed: {"WFLYCTL0080: Failed services" => {"jboss.persistenceunit.\"ProcessMigrationIT.war#migration-unit\"" => "java.lang.ClassCastException: org.dom4j.DocumentFactory cannot be cast to org.dom4j.DocumentFactory
```

process-migration-service/src/test/java/org/kie/processmigration/it/ProcessMigrationIT.java
```diff
@@ -0,0 +1,193 @@
...
+import org.kie.server.api.model.instance.ProcessInstance;
+import org.kie.server.client.KieServicesClient;
+import org.kie.server.client.KieServicesConfiguration;
+import org.kie.server.client.KieServicesFactory;
+import org.kie.server.client.ProcessServicesClient;
+import org.wildfly.swarm.arquillian.DefaultDeployment;
+import org.wildfly.swarm.arquillian.DefaultDeployment.Type;
+
+@RunWith(Arquillian.class)
+@DefaultDeployment(type = Type.WAR)
```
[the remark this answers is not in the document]
ruromero: I think so but somehow I'm having ClassNotFoundException for Migration.class during integration testing

process-migration-service/src/main/java/org/kie/processmigration/model/Execution.java
```diff
@@ -0,0 +1,85 @@
...
+    public enum ExecutionStatus {
+        SCHEDULED,
+        STARTED,
+        COMPLETED,
+        FAILED,
+        CANCELLED,
+        CREATED
+    }
+
+    public static final Execution SYNC = new Execution().setType(ExecutionType.SYNC);
```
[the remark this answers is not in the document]
ruromero: Unfortunately the `columnDefinition = "TEXT"` doesn't seem to work well with Oracle. Moving to `@Lob`

process-migration-service/pom.xml
```diff
@@ -0,0 +1,506 @@
...
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.cargo</groupId>
+                <artifactId>cargo-maven2-plugin</artifactId>
+                <configuration>
+                    <container>
+                        <containerId>${cargo.container.id}</containerId>
+                        <systemProperties>
+                            <kie.maven.settings.custom>${kie.server.server.deployment.settings.xml}</kie.maven.settings.custom>
```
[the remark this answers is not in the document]
ruromero: Thanks! I've done a clean up of the properties and configuration

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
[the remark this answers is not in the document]
ruromero: I have done the validation at the time of creating an async migration and when starting the migration. 
As plans are not related to any kieserver, the validation cannot be done.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+        Migration migration = get(id);
+        if (migration == null) {
+            return null;
+        }
+        if (!ExecutionStatus.SCHEDULED.equals(migration.getStatus())) {
+            throw new ReScheduleException("The migration is not scheduled and cannot be re-scheduled");
+        }
+        if (ExecutionType.SYNC.equals(definition.getExecution().getType())) {
+            throw new ReScheduleException("The migration execution type MUST be ASYNC");
+        }
```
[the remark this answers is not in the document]
ruromero: That's why we need to throw an exception. This is an update and maybe the user would try to change the type from ASYNC to SYNC when it is in the SCHEDULED state

process-migration-service/README.md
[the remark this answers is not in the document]
MarianMacik: OK, my bad for not reading the manual :)

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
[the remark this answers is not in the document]
MarianMacik: Yes, you are right.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+        migration.setDefinition(definition);
+        em.persist(migration);
```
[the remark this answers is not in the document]
MarianMacik: As we discussed on the chat, this can be removed completely since `migration` is already managed because it was fetched from DB in `get()` method. So call to `setDefinition()` will be propagated to DB when committing.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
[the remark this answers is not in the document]
MarianMacik: True, makes sense.

process-migration-service/src/test/resources/META-INF/jboss-deployment-structure.xml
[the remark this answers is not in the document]
MarianMacik: OK.

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+        return migration;
+    }
+
+    @Override
+    @Transactional
+    public Migration migrate(Migration migration) throws InvalidMigrationException {
+        try {
+            Plan plan = planService.get(migration.getDefinition().getPlanId());
+            if (ExecutionStatus.CREATED.equals(migration.getStatus()) || ExecutionStatus.SCHEDULED.equals(migration.getStatus())) {
+                em.persist(migration.start());
```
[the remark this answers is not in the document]
ruromero: As discussed online, it won't

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+                if (instances.size() < QUERY_PAGE_SIZE) {
+                    allFetched = true;
+                }
+            }
+        }
+        if (migration.getReports() == null || migration.getReports().isEmpty()) {
+            return instanceIds;
+        }
+        List<Long> migratedInstances = migration.getReports().stream().map(r -> r.getMigrationId()).collect(Collectors.toList());
+        return instanceIds.stream().filter(id -> !migratedInstances.contains(id)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
ruromero: I'm working on your suggestion. It is definitely better to use individual transactions for each process instance migration.
Thanks for noticing the `getProcessInstanceId()`

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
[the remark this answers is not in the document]
MarianMacik: By looking at it once again I think that this if statement is not needed at all. `migrate()` is called only when:
* Recovery is called - and then the migration is already managed since it comes straight from DB
* `@Timeout` occurred - in this case it is again already managed, since it is fetched from DB
* We migrate it right after `submit()` method is called which has, again, already fetched it from DB

So I think this if can be omitted completely.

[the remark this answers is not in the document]
ruromero: If the migration is recovered, the status is already `STARTED` and the `startedAt` field is set, if we omit this condition, the date will be reset.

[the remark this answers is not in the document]
MarianMacik: OK, but then it is sufficient to just call `migration.start()`.

ruromero:
@MarianMacik @mswiderski
I think I have implemented all the suggested changes. Can you please give a final review?
In order to have a separate transaction for each migration I implemented a TransactionHelper, do you think it's a good approach?
Thanks for all the effort so far :muscle:

ruromero:
In the IntegrationTest I'm facing a `ClassNotFoundException` issue which I don't know how to fix. I'm afraid it can be related to this https://issues.jboss.org/browse/THORN-767
Maybe @Ladicek might know better.

Ladicek:
Sorry, I have absolutely no idea what this is about, and the PR has 60 files! I'm pretty sure that you're not hitting https://issues.jboss.org/browse/THORN-767, because you're not using `@CreateSwarm` (which is deprecated and buggy as hell).

ruromero:
@Ladicek if this Jira is only related to `@CreateSwarm` then I'll continue looking. Thanks!

process-migration-service/pom.xml
```diff
@@ -0,0 +1,439 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.drools</groupId>
+        <artifactId>droolsjbpm-integration</artifactId>
+        <version>7.15.0-SNAPSHOT</version>
```
mswiderski: 7.16.0-SNAPSHOT on master already

process-migration-service/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The right way to configure the connection to one or more KIE Servers in order to perform the migrations, a list of kieservers should exist in the configuration file. 
+
+```
+kieservers:
+  - host: kieserver1.example.com
+    port: 8080
+    protocol: http
+    contextRoot: /kie-server
+    path: /services/rest/server
```
mswiderski: @ruromero is there any reason to keep all these elements (protocol, port, host, etc) as separate entries instead of single one e.g. url?

process-migration-service/src/main/java/org/kie/processmigration/service/RecoveryService.java
```diff
@@ -0,0 +1,64 @@
...
+        logger.info("Resuming ongoing migrations ...");
+        TypedQuery<Migration> query = em.createNamedQuery("Migration.findByStatus", Migration.class);
+        query.setParameter("statuses", PENDING_STATUSES);
+        query.getResultList().stream().forEach(m -> {
```
mswiderski: you could skip the stream() as the list itself has the forEach method already

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,307 @@
...
+                List<ProcessInstance> instances = kieService.getQueryServicesClient(migration.getDefinition().getKieserverId())
+                                                            .findProcessInstancesByContainerId(plan.getSourceContainerId(), QUERY_PROCESS_INSTANCE_STATUSES, page++, QUERY_PAGE_SIZE);
+
+                instances.stream().forEach(p -> instanceIds.add(p.getId()));
+                if (instances.size() < QUERY_PAGE_SIZE) {
+                    allFetched = true;
+                }
+            }
+        }
+        if (migration.getReports() == null || migration.getReports().isEmpty()) {
```
mswiderski: Shouldn't this if be at the beginning of the method? that way we don't have to load instances if report is empty...

process-migration-service/src/test/java/org/kie/processmigration/it/ProcessMigrationIT.java
```diff
@@ -0,0 +1,193 @@
...
+        pids.add(client.startProcess(SOURCE_CONTAINER_ID, PROCESS_ID));
+        return pids;
+    }
+
+    private void deployProcesses(KieServicesClient kieServicesClient) throws IOException {
+        KieServices ks = KieServices.Factory.get();
+        MavenRepository repo = MavenRepository.getMavenRepository();
+        for (String version : Arrays.asList("1.0.0", "2.0.0")) {
+            org.kie.api.builder.ReleaseId builderReleaseId = ks.newReleaseId(GROUP_ID, ARTIFACT_ID, version);
+            File kjar = readFile(CONTAINER_ID + "-" + version + ".jar");
```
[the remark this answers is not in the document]
mswiderski: we can start with jar right now and maybe move (and by that make the test setup more complex) later on

process-migration-service/README.md
ruromero: Each kie endpoint might have different configuration (containerised vs non-containeraised)
```
- host: localhost
  port: 8080
  protocol: http
  contextRoot: /kie-server
  path:/services/rest/server
- host: localhost
  port: 8280
  protocol: http
  contextRoot: /kie-server
  path:/services/rest/server
- host: kieserver2.example.com
  port: 80
  protocol: http
  contextRoot: /
  path:/services/rest/server
```

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
ruromero: Good catch!

process-migration-service/README.md
mswiderski: that's clear, what I was asking is 

```
- host: localhost
  port: 8080
  protocol: http
  contextRoot: /kie-server
  path:/services/rest/server
```

compared with 

```
- host: http:://localhost:8080/kie-server/services/rest/server
```

as this will give you same result but much less verbose, wdyt?

process-migration-service/README.md
```diff
@@ -0,0 +1,150 @@
+# Process Instance Migration Service
```
mswiderski: we need to add a section on how to use it. I tested base migration for evaluation process with following which could be the base for the docs. This will give users really easy way to make use of the features

1. Define plan without any node mapping

```
http://localhost:8180/plans

{
    "name": "Test plan",
    "description": "Evaluation Process Test Plan",
    "sourceContainerId": "evaluation_1.0",
    "targetProcessId": "evaluation",
    "targetContainerId": "evaluation_1.1"
}
```

2. Start Kie Server and deploy two versions of evaluation project (evaluation_1.0 and evaluation_1.1)
3. Start one instance of evaluation process from project (evaluation_1.0)
4. Create sync migration

```
http://localhost:8180/migrations

{
    "planId": 1,
    "processInstanceIds": [1],
    "kieserverId": "sample-server",
    "execution": {
   	 "type": "SYNC"
    }
}
```

5. View migration details
http://localhost:8180/migrations/1
http://localhost:8180/migrations/1/results

6. Create two additional process instances of evaluation process from project (evaluation-1.0)
7. Run migration as scheduled async operation for both instances

```
http://localhost:8180/migrations

{
    "planId": 1,
    "processInstanceIds": [3,4],
    "kieserverId": "sample-server",
    "execution": {
   	 "type": "ASYNC",
   	 "scheduledStartTime": "2018-12-11T12:35:00.000Z"
    }
}
```

process-migration-service/src/test/java/org/kie/processmigration/utils/ContextUtils.java
```diff
@@ -0,0 +1,27 @@
+package org.kie.processmigration.utils;
```
mswiderski: missing license header

process-migration-service/src/test/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,24 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    version="2.1"
+    xmlns="http://xmlns.jcp.org/xml/ns/persistence"
+    xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd">
+    <persistence-unit name="migration-test" transaction-type="RESOURCE_LOCAL">
+        <provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
+        <class>org.kie.processmigration.model.Plan</class>
+        <class>org.kie.processmigration.model.Migration</class>
+        <class>org.kie.processmigration.model.MigrationReport</class>
+        <class>org.kie.processmigration.model.Credentials</class>
```
mswiderski: why Credentials is here and not in the main persistence.xml? See 
https://github.com/kiegroup/droolsjbpm-integration/pull/1575/files#diff-c414854ad24e6abc450694bc39c11fc8

process-migration-service/src/test/resources/kjars/test-1.0.0.pom
```diff
@@ -0,0 +1,59 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.myspace</groupId>
+  <artifactId>test</artifactId>
+  <version>1.0.0</version>
+  <packaging>kjar</packaging>
+  <name>test</name>
+  <description></description>
+  <dependencies>
+    <dependency>
```
mswiderski: I don't think we need any of the dependencies here so feel free to remove them. Same goes for kie-maven-plugin as the jar is already build. So remove that too and change packaging to jar - that way we get rid of all references for versions which could be problematic on some environments.

process-migration-service/src/test/resources/kjars/test-2.0.0.pom
```diff
@@ -0,0 +1,59 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.myspace</groupId>
+  <artifactId>test</artifactId>
+  <version>2.0.0</version>
+  <packaging>kjar</packaging>
+  <name>test</name>
+  <description></description>
+  <dependencies>
+    <dependency>
```
mswiderski: same cleanup here

process-migration-service/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+It is possible to override or extend the provided configuration.
+
+### Defining KIE Servers
+
+The right way to configure the connection to one or more KIE Servers in order to perform the migrations, a list of kieservers should exist in the configuration file. 
+
+```
+kieservers:
+  - host: kieserver1.example.com
```
mswiderski: Kie Server id is required for running migration but it is not specified in the Kie Server configuration. Though it is expected to match org.kie.server.id of the target Kie Server. Maybe we should make this more visible here? Meaning users would specify the id here and we document that it must match the server id on the other side...

process-migration-service/src/test/resources/META-INF/persistence.xml
ruromero: Legacy from old implementation. Removing it.

process-migration-service/README.md
ruromero: During start-up the kieserver-id will be retrieved from the remote kie-server instance. That way we leverage the users from providing this.
https://github.com/kiegroup/droolsjbpm-integration/pull/1575/files#diff-b6e3969848272b6dc3b5be066b90be4bR107

process-migration-service/README.md
ruromero: Why not? :smile:

process-migration-service/README.md
ruromero: Done

mswiderski:
ok to test

mswiderski:
Jenkins retest this

process-migration-service/src/test/resources/project-test.yml
```diff
@@ -0,0 +1,30 @@
+thorntail:
```
MarianMacik: Just out of curiosity. Do we have a test stage or -Stest configuration anywhere? I cannot see it. What is the difference between stages and -S switch?

process-migration-service/src/test/java/org/kie/processmigration/utils/ContextUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import org.wildfly.swarm.container.config.ConfigViewFactory;
+import org.wildfly.swarm.container.config.ConfigViewImpl;
+import org.wildfly.swarm.spi.api.config.ConfigView;
+
+@ApplicationScoped
+public class ContextUtils {
+
+    @Produces
+    private ConfigView createConfigView() throws IOException {
+        URL url = getClass().getClassLoader().getResource("h2-defaults.yml");
```
MarianMacik: I guess `h2-defaults.yml` is available inside thorntail by default?

ruromero: The whole class is not needed anymore. Thanks for noticing

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,307 @@
...
+    private void validatePlanExecution(MigrationDefinition definition) throws InvalidMigrationException {
+        try {
+            validatePlanExecution(definition, planService.get(definition.getPlanId()));
+        } catch (PlanNotFoundException e) {
+            throw new InvalidMigrationException("Plan not found with ID: " + definition.getPlanId());
+        }
+    }
+
+    private void validatePlanExecution(MigrationDefinition definition, Plan plan) throws InvalidMigrationException {
+        if (!kieService.existsProcessDefinition(plan.getSourceContainerId(), plan.getTargetProcessId(), definition.getKieserverId())) {
```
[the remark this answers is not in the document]
mswiderski: @ruromero there is no need to check if there target process id exists in source container. Actually in most of the cases it won't exists as recommendation for process instance migration is to use unique process ids

process-migration-service/src/main/java/org/kie/processmigration/service/impl/MigrationServiceImpl.java
```diff
@@ -0,0 +1,307 @@
...
+                                                            .findProcessInstancesByContainerId(plan.getSourceContainerId(), QUERY_PROCESS_INSTANCE_STATUSES, page++, QUERY_PAGE_SIZE);
+
+                instances.stream().forEach(p -> instanceIds.add(p.getId()));
+                if (instances.size() < QUERY_PAGE_SIZE) {
+                    allFetched = true;
+                }
+            }
+        }
+        return instanceIds;
+    }
```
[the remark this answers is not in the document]
ruromero: You are right, I've done a small refactor

mswiderski:
Jenkins retest this

PR #1576, opened 2018-09-20
JBPM-7777 - Stabilize OptaplannerIntegrationTest

sutaakar:

PR #1578, opened 2018-09-24
Fix test failure in WebSocketKieControllerRuleCapabilitiesIntegration…

sutaakar:
…Test

Some recent change (didn't identify it) caused failures in KieControllerRuleCapabilitiesIntegrationTest.testScanNow(). In this test class there are several test methods updating kjar and container from version 1.0.0 to 1.0.1. Cleanup was solved by deleting 1.0.1 kjar from Kie server local repo. It used to be sufficient as when container with LATEST version was created the Kie server seemed to check just its own local repo for kjars with appropriate versions.

Now the behavior is changed and Kie server checks also remote repos for latest kjar in case LATEST container is created. However deleting kjar with version 1.0.1 in remote repo is not enough as the versions are also stored in maven-metadata....xml and they are used by Kie server when resolving the version. Rather then deleting maven-metadata....xml I decided to completely delete repos as it is cleaner solution (Maven java client doesn't support artifact removal and I want to keep environment as consistent as possible).


sutaakar:
Test failure unrelated, caused by https://github.com/kiegroup/droolsjbpm-integration/pull/1576

PR #1580, opened 2018-09-24
JBPM-7797 - ContainerAliasResolver requires KieServer to be successfu…

mswiderski:
…lly autowired


tsurdilo:
I tested this locally with services app and it works great . Now you can do 

@Autowired
    private ContainerAliasResolver containerAliasResolver;

without having to prior do

@Autowired
    private KieServer kieServer;

thank you @mswiderski

PR #1586, opened 2018-09-26
[DROOLS-2810] Add test for installation of kie-dmn

jiripetrlik:


mareknovotny:
do we need this still @jiripetrlik  ?

jiripetrlik:
@mareknovotny 
Hello I think it would be good to merge despite it is pending quite a long here. It is not absolutely necessary because we have tests for kie-dmn in KieDMNKarafIntegrationTest, but it would be good to have test which only installs this feature same way as we do with other features.

mareknovotny:
jenkins retest this please

PR #1590, opened 2018-09-27
PLANNER-614: Rename getContainers() to getSolverInstances()

yurloc:
I _believe_ this doesn't break any user-facing API:

1. The method name doesn't affect marshalling output. JSON, JAXB and XStream formats are unchanged after the rename. Just for the record:

## JSON
```json
{
  "solver" : [ {
    "container-id" : "employeerostering_1.0.0-SNAPSHOT",
    "solver-id" : "solver1",
    "solver-config-file" : "employeerostering/employeerostering/EmployeeRosteringSolverConfig.solver.xml",
    "status" : "NOT_SOLVING",
    "score" : {
      "value" : null,
      "scoreClass" : null
    },
    "best-solution" : null
  } ]
}
```
## JAXB
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<solvers>
    <solver>
        <container-id>employeerostering_1.0.0-SNAPSHOT</container-id>
        <solver-id>solver1</solver-id>
        <solver-config-file>employeerostering/employeerostering/EmployeeRosteringSolverConfig.solver.xml</solver-config-file>
        <status>NOT_SOLVING</status>
        <score/>
    </solver>
</solvers>
```
## XStream
```xml
<org.kie.server.api.model.instance.SolverInstanceList>
  <solvers>
    <solver-instance>
      <container-id>employeerostering_1.0.0-SNAPSHOT</container-id>
      <solver-id>solver1</solver-id>
      <solver-config-file>employeerostering/employeerostering/EmployeeRosteringSolverConfig.solver.xml</solver-config-file>
      <status>NOT_SOLVING</status>
      <score/>
    </solver-instance>
  </solvers>
</org.kie.server.api.model.instance.SolverInstanceList>
```
2. Remote Java client is unaffected because the renamed method is not exposed in its API.

### DIFF
--- a/kie-server-parent/kie-server-api/src/build/revapi-config.json
+++ b/kie-server-parent/kie-server-api/src/build/revapi-config.json
@@ -203,6 +203,24 @@
           "methodName": "setLastModificationUser",
           "elementKind": "method",
           "justification": "JBPM-6890: Adding the customTaskID as a unique identifier for an active"
+        },
+        {
+          "code": "java.method.removed",
+          "old": "method java.util.List<org.kie.server.api.model.instance.SolverInstance> org.kie.server.api.model.instance.SolverInstanceList::getContainers()",
+          "package": "org.kie.server.api.model.instance",
+          "classSimpleName": "SolverInstanceList",
+          "methodName": "getContainers",
+          "elementKind": "method",
+          "justification": "PLANNER-614: Fix method name. Method renamed to getSolverInstances()"
+        },
+        {
+          "code": "java.method.removed",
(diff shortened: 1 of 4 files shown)


sutaakar:
@yurloc Well, the change can affect customers in case they use async response (as shown in the test change). I think it would be good to at least document it.

yurloc:
@sutaakar Good catch, Karel. I didn't realize that `SolverInstanceList` is accessible through the async API! My assumption was that we can fix this without affecting public API. Now I understand it's not true. Since this is nearly a trivial problem I'll [defer it to 8.0](https://issues.jboss.org/browse/PLANNER-750).

PR #1596, opened 2018-10-04
[7.11.x] Stabilize SpringBoot tests

MarianMacik:
Small refinements as discussed + springboot test jar now has a name ending in -SNAPSHOT instead of a volatile timestamp, so it can be automatically tested with the latest SNAPSHOT version (before it couldn't locate the correct jar file).


MarianMacik:
Failures not related.

MarianMacik:
jenkins retest this

PR #1606, opened 2018-10-17
JBPM-7872 - DataSetLookupException when Kie server is initialized

mswiderski:
@sutaakar here is the fix, could you please have a look? 

It is a blocker for 7.13 so the sooner we merge it the better


mswiderski:
Jenkins retest this

PR #1607, opened 2018-10-18
Cleanup allocated port when embedded server is turned off

sutaakar:
To avoid rare BindException failures caused by occupied port in case embedded server is used in test.


sutaakar:
Cannot be implemented now - clashing with controller tests which are parametrized (trying to connect to already shut down Kie server). Needs to be implemented in different way.

PR #1625, opened 2018-11-03
Revert "JBPM-7887 - Use enum in ProcessConfig.setRuntimeStrategy() me…

cristianonicolai:
…thod"

This reverts commit 2be9414232e83cc36594f3ffd614782a9b18427f.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/pom.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/pom.xml
@@ -25,11 +25,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.kie</groupId>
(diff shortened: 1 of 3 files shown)


cristianonicolai:

rsynek:
@cristianonicolai Looking at the latest build failure, the FDB for this particular repository is probably wrongly configured - it's missing activation of "wildfly11" profile (although this profile now uses WFLY14).

As the change you're reverting does not seem to be connected with any particular requirement, +1 for reverting, if it unblocks the build.

krisv:
Merging to unblock master.  These test failures however need to be looked at, they do not seem to be false positives (but most likely unrelated to this change).

PR #1639, opened 2018-11-13
BXMSDOC-3365 provide missing content for Drools segments in KIE APIs doc

tarilabs:
@mswiderski in order to complete task BXMSDOC-3365 for @sterobin I had to create some actual working examples, which at this point I believe better to just incorporate in code base. wdyt?


mswiderski:
@tarilabs I believe what you have here is already covered in drools integrations tests so it looks to me as duplicates

tarilabs:
@mswiderski signal command cannot be covered in Drools because (afaik) is a BPMN feature. Also the goal of these is to have the same snippet used from the doc in the codebase, hence I just preferred to map all the missing one as specific example.

mswiderski:
@tarilabs if you want to show jbpm then you should not use commands at all. Commands are only used for drools (and maybe dmn extensions). jBPM uses dedicated endpoints for all operations exposed.

tarilabs:
@mswiderski no I believe there is a misunderstanding here. I have no specific intention besides to address what I was requested as part of BXMSDOC-3365. If Signal command is deprecated we should then simply remove it from the docs. Can you kindly take a look at @sterobin proposal in BXMSDOC-3365 and let me know if the signal command section should be dropped then, please?

tarilabs:
```
Test Name | Duration | Age
-- | -- | --
org.kie.camel.embedded.camel.component.cxf.CxfSoapTest.test1
```

Jenkins test again

mswiderski:
as I said, commands are irrelevant in context of jBPM so they should only be document in context of drools/dmn which I don't know exactly what is used and what is not. part of jBPM is used by drools to call rule flows but again I don't know which one.

tarilabs:
Okay, so to recap.

The intention of BXMSDOC-3365 is to cover Drools command with JSON payload examples.
I understand we can keep all Commands listed in the docs.

This PR intention was to preserve the examples I needed to craft in order to actually be able to provide those payload examples for BXMSDOC-3365, but I understand this is seen as test/code duplication.

I believe there is no easy way to satisfy both requirements, hence I am going to close this PR and if needed I hope JIRA will keep a reference, shall the documentation eventually be refreshed.

I hope there will be a swagger annotation solution also for this problem --rest endpoint accepting String/Object but being a serialization of a well known class, in this case BatchCommand-- going forward so I don't have to rely on examples :)

mswiderski:
@tarilabs if you need to just get the payloads there is no need to use entire kie server client, it's enough to use particular marshaller from kie server api, see this test case https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/model/MarshallingRoundTripCustomClassListTest.java

tarilabs:
> if you need to just get the payloads there is no need to use entire kie server client, it's enough to use particular marshaller from kie server api, see this test case

that looks very helpful, thanks @mswiderski I hope I will remember about that :D
However in this case I needed to provide not only request, but also the related response :) so I indeed needed a working example

mswiderski:
@tarilabs I would argue here ... as you know what will be returned (ServiceResponse with result) you can do the same way with marshaller only too... but I won't argue ;)

PR #1645, opened 2018-11-20
JBPM-7969 - Make use of @Wid annotation to register work item handlers

mswiderski:
@tsurdilo here is the enhancement we talked about


tsurdilo:
Jenkins retest please

PR #1646, opened 2018-11-23
JBPM-7934 - Swagger Controller updates

cristianonicolai:
Review PUT and POST behaviour
Return resource location for POST methods
Migrate Maven Controller to use Controller Client API


cristianonicolai:
@sterobin @mswiderski final changes to this. One thing to notice is that I changed a few methods from `PUT` to `POST` to identify a resource creation and from `POST` to `PUT` to identify an update as it seems a more commonly used approach. That can be considered a breaking change so I'm fine to revert that if you think is not worth.

mswiderski:
@cristianonicolai not sure how commonly used this REST api is but I would be reluctant at changing it ... if we decide to change it please also update maven plugin for kie controller

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
```diff
@@ -148,7 +163,12 @@ public Response saveServerTemplate(@Context HttpHeaders headers,
 
             specManagementService.saveServerTemplate(serverTemplate);
             logger.debug("Returning response for save server template with id '{}': CREATED", serverTemplateId);
-            return createCorrectVariant("", headers, Response.Status.CREATED);
+            UriBuilder builder = uriInfo.getAbsolutePathBuilder();
+            builder.path(serverTemplate.getId());
```
MarianMacik: Maybe I'm missing something, but doesn't the `uriInfo.getAbsolutePathBuilder();` already contain full path of the reuquest including resolved URI path template parameters, e.g. `..../servers/2`? If that's the case, I don't think we need to append yet another ID by calling `builder.path(serverTemplate.getId());`.

cristianonicolai: @MarianMacik good catch, fixed.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
```diff
@@ -69,7 +75,10 @@ public Response saveContainerSpec(@Context HttpHeaders headers,
 
             specManagementService.saveContainerSpec(serverTemplateId, containerSpec);
             logger.debug("Returning response for save container spec request for server template with id '{}': CREATED", serverTemplateId);
-            return createCorrectVariant("", headers, Response.Status.CREATED);
+            String response = marshal(contentType,
+                                      new ServiceResponse(KieServiceResponse.ResponseType.SUCCESS,
+                                                          "KIE container " + containerId + " successfully deployed"));
+            return createCreatedVariant(response, headers, uriInfo.getAbsolutePathBuilder());
```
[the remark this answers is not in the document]
mswiderski: @sterobin I disagree with this statement that every api call should return human-friendly response. Frankly speaking I don't see any need for that as the API docs are meant to provide this human-friendly details that can be easily mapped to response code and potential data returned. Moreover adding additional payload to all request would
- increase time required to receive response as the payload must be marshalled (on both ends)
- increase network use as the number of bytes to be sent will increase
- will in most of the cases be ignored anyway as that is never parsed to present on UI especially when international end users are involved - why would you show English text to Spanish speaking people and since that is technical anyway what's the point anyway? Instead application developers should based their feedback message to UI based on the API docs and response code

You also mentioned that our competitors do this could you please provide examples for it?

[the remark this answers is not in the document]
sterobin: @mswiderski, you make good points, especially about translation and international considerations. Regarding other APIs, Oracle, HelpSpot, Twitter, Facebook, and other larger organizations are the ones I looked at that do a good job with responses and response messages. Requests typically have a response of some kind, if only repeating back to the user the asset data related to the request, except for delete requests, which typically don't have a response. However, to your point, these aren't our competitors, so if I look at Signavio or the [Camunda API](https://docs.camunda.org/manual/7.7/reference/rest/execution/) for example, which are more our competitors, they follow a method more like what you're saying, but they too are inconsistent.

So I can roll with your logic Maciej, but the only problem I still have is that there are fewer cases of no response messages in the KIE Server REST API than there are with response messages. I was just trying to help us to be consistent in our responses, and I applied the same logic to the controller API for this PR. So imo, we need to choose one route or the other. If we want to go no response messages for reasons you've explained, and just rely on codes, then we need to do this universally. Many KIE Server API requests say "X thing successfully ___". I've just tried to fill in the gaps, but if we don't want to, then those messages should likewise be removed. When we're inconsistent, it looks like the ones without responses are a bug.

So for this PR, @cristianonicolai, I leave it to you whether you want response messages or not. But we need to be consistent.

And @mswiderski, just a reminder that [BXMSDOC-3368](https://issues.jboss.org/browse/BXMSDOC-3368) assigned to you and Tiho is the KIE Server counterpart to this PR, for applying improved API endpoint data to the KIE Serer REST API swagger source. You can likewise choose to either add response messages where I've indicated, or remove response messages where they currently exist, for consistency.

kie-server-parent/kie-server-controller-plugin/README.md
```diff
@@ -28,7 +28,7 @@ On pom.xml of local project add plugin:
     <plugin>
         <groupId>org.kie.server</groupId>
         <artifactId>kie-server-controller-plugin</artifactId>
-        <version>7.5.0-SNAPSHOT</version>
+        <version>7.15.0-SNAPSHOT</version>
```
mswiderski: should be 7.16 now or maybe even replace it with ${version.org.kie} just to indicate it should be same version as all other org.kie artefacts

cristianonicolai: changed to ${version.org.kie}

PR #1649, opened 2018-11-26
Remove Narayana version override from kie-spring-boot

MarianMacik:
@mswiderski  7.14.x backport is probably not relevant anymore, right?


mswiderski:
I'd say if might be worth to cherry pick it, @krisv what do you think?

PR #1664, opened 2018-12-05
Reset all allocated ports for Kie server tests in @AfterClass

sutaakar:
@mswiderski Should hopefully help to get rid of port allocation failures.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/config/TestConfig.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/config/TestConfig.java
@@ -195,6 +195,15 @@ public class TestConfig {
     }
 
     /**
+     * Reset all allocated ports so they can be dynamically reallocated again, to avoid BindException: Address already in use (Bind failed).
+     */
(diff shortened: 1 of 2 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/config/TestConfig.java
```diff
@@ -194,6 +194,15 @@ public static Integer getKieServerAllocatedPort() {
         return ALLOCATED_PORT;
     }
 
+    /**
+     * Reset all allocated ports so they can be dynamically reallocated again, to avoid BindException: Address already in use (Bind failed).
+     */
+    public static void resetAllocatedPorts() {
+        ALLOCATED_PORT = null;
+        CONTROLLER_ALLOCATED_PORT = null;
+        ROUTER_ALLOCATED_PORT = null;
```
triceo: This doesn't solve the issue, right? It avoids it and at the same time is making it worse for those who come after.
Shouldn't we rather make sure we actually close those ports?

sutaakar:
It turns out that this is actually a problem with not correctly stopped embedded server, will fix it in another PR.

PR #1681, opened 2019-01-07
JBPM-8106 - Multiple work item handlers with reference to process ser…

mswiderski:
…vice fail at boot with circular dependency exception


mswiderski:
Jenkins retest this

PR #1685, opened 2019-01-09
JBPM-8117 - Springboot doesn't release database connections

MarianMacik:
Hi @mswiderski,

I have found that this line:

https://github.com/kiegroup/droolsjbpm-integration/blob/e82c61c1b9f4acb12870c51c80e8959fa9162383/kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/autoconfigure/JBPMAutoConfiguration.java#L186

added when splitting data source configuration to another module has caused springboot to hold connections and not closing them.

The root cause is that here:
https://github.com/spring-projects/spring-framework/blob/df13e4dd5e85cc1fde12277f17ad7a7f64ce7e3d/spring-orm/src/main/java/org/springframework/orm/jpa/vendor/HibernateJpaVendorAdapter.java#L167
the HibernateJpaVendorAdapter will add this configuration item to the JpaPropertyMap of AbstractEntityManagerFactoryBean if it is not already there:
https://github.com/spring-projects/spring-framework/blob/v4.3.3.RELEASE/spring-orm/src/main/java/org/springframework/orm/jpa/AbstractEntityManagerFactoryBean.java#L330

And since the programmatic properties have priority over persistence.xml ones when creating EMF, it will override our setting which is in jbpm-persistence.xml in jbpm autoconfiguration module as well.

So, we have 2 options:

1. Drop this `factoryBean.setJpaVendorAdapter(new HibernateJpaVendorAdapter());` line, because it seems that it works even without it.

1. Leave it there (probably preferred to have it properly configured), but then we have to add this setting to all application.properties files, which I have already done. I have added it only to non-test application.properties files, since they are used in tests as well according to the Spring config.


MarianMacik:
fails not related
jenkins retest this

mswiderski:
@MarianMacik I opt for option 3 so we stick to what we define in persistence.xml

MarianMacik:
Change applied.

PR #1694, opened 2019-01-22
[JBPM-8153] Container is removed from UI even though it was not possible to stop it.

elguardian:
the status is not properly set in the resource response from the kie server


elguardian:
https://github.com/kiegroup/kie-wb-common/pull/2415 depends on this.

PR #1699, opened 2019-01-25
DROOLS-3543 Raise timeout for BlueprintContainer initialization

baldimir:
Timeout raised because when running with specific databases, it takes a long time to initialize the container. 

@jiripetrlik @mariofusco could you please review?


jiripetrlik:
Looks good to me.

PR #1705, opened 2019-01-29
JBPM-8088 - Create separate Kie Server war files for ee7 and ee8

mswiderski:
depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/917

@sutaakar would you mind to review and maybe try it out on various containers? I tested myself on WildFly 14 with ee8 and on WAS with ee7


sutaakar:
Sure, will take a look.

mswiderski:
Jenkins retest this

kie-server-parent/kie-server-wars/kie-server/src/main/assembly/assembly-ee8-container.xml
```diff
@@ -0,0 +1,116 @@
...
+        <exclude>org.jboss.spec.javax.ws.rs:*</exclude>
+        <exclude>javax.enterprise:cdi-api</exclude>
+        <exclude>javax.inject:javax.inject</exclude>
+        <exclude>javax.validation:validation-api</exclude>
+        <exclude>stax:stax-api</exclude>
+        <exclude>javax.activation:activation</exclude>
+        <exclude>javax.persistence:javax.persistence-api</exclude>
+        <exclude>xml-apis:xml-apis:jar</exclude>
+        <exclude>javax.mail:mail</exclude>
+        <exclude>org.apache.cxf:cxf-rt-ws-policy</exclude>
```
sutaakar: AFAIK this has been excluded due to compatibility issues with WebSphere. As ee8 classifier is used just by WildFly/EAP, would it have sense to comment this out (with description)?

mswiderski:
@sutaakar what would you say as preferred solution - upgrade controller classifier to ee8 or just change the config? Personally I don't see much point in upgrading the classifier as it does not really have anything from ee8...

sutaakar:
@mswiderski There is just a need for separate controller for Tomcat, otherwise it works on all application servers. It may be the best to adjust classifier just to 'ee' and use it for all JEE servers.

sutaakar:
Or just to keep 'ee7' is ok too.

sutaakar:
Controller adjustments look good. Controller tests passed for all application servers.

kie-server-parent/kie-server-tests/pom.xml
```diff
@@ -26,9 +26,10 @@
     <kie.server.context>kie-server-services</kie.server.context>
     <cargo.container.id>valid-id-needs-to-be-specified-in-appropriate-container-profile</cargo.container.id>
     <kie.server.controller.context>kie-server-controller-services</kie.server.controller.context>
-    <kie.server.classifier>ee7</kie.server.classifier>
+    <kie.server.classifier>ee8</kie.server.classifier>
+    <kie.server.controller.classifier>ee7</kie.server.controller.classifier>
```
mareknovotny: I think this is clear now, but maybe we could have here a note/comment about why we have these 2 slightly similar properties for classifier?

PR #1719, opened 2019-02-08
Fix tests and Migration update

ruromero:

* Unit tests were failing due to updated hibernate library
* Fixed problem while updating Migration instances

### DIFF
--- a/process-migration-service/src/test/java/org/kie/processmigration/persistence/TestEntityManager.java
+++ b/process-migration-service/src/test/java/org/kie/processmigration/persistence/TestEntityManager.java
@@ -1,0 +1,289 @@
+package org.kie.processmigration.persistence;
+
+import java.util.List;
+import java.util.Map;
+
+import javax.persistence.EntityGraph;
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.EntityTransaction;
+import javax.persistence.FlushModeType;
+import javax.persistence.LockModeType;
+import javax.persistence.Query;
+import javax.persistence.StoredProcedureQuery;
+import javax.persistence.TypedQuery;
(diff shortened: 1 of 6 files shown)


kiereleaseuser3:

ruromero:
CC @MarianMacik

MarianMacik:
ok to test

MarianMacik:
@mswiderski any comments?

process-migration-service/src/test/java/org/kie/processmigration/persistence/TestEntityManager.java
```diff
@@ -0,0 +1,289 @@
+package org.kie.processmigration.persistence;
```
mswiderski: missing license header

MarianMacik:
@mswiderski There is a header template configured in kie-parent: `.*` which fits for everything. Moreover, in kie-parent we have: `<checkstyle.failOnViolation>false</checkstyle.failOnViolation>`. In jBPM we have quite largely overridden configuration so it just checks templates and nothing else. If I turn failOnViolation on, there are ~130 violations just in process-instance-migration module. Finally, I have noticed that process-instance-migration is not built by default, see:
https://github.com/kiegroup/droolsjbpm-integration/blob/947db7e178ed6d6330adb264dd7a0d9093c77a9e/pom.xml#L52-L81

So that's why :) I would turn on failOnViolation and fix the issues in whole repo, but certainly not this week. Wdyt?

mswiderski:
thanks @MarianMacik, that explains why I can't find it in any release in maven repo, +1 at enabling that whenever you have the time.

process-migration-service/src/test/java/org/kie/processmigration/persistence/TestEntityManager.java
ruromero: Done, thanks

PR #1723, opened 2019-02-12
JBPM-8227: Move RestEasy dependencies to the rest-common module

sutaakar:
As there is no reason to have them in services-common.

kie-server-controller-openshift artifact has a dependency on kie-server-services-common which brought unintentional RestEasy dependency.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/pom.xml
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/pom.xml
@@ -40,6 +40,10 @@
     </dependency>
     <dependency>
       <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.server</groupId>
       <artifactId>kie-server-services-common</artifactId>
     </dependency>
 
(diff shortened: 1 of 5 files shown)


jhrcek:
Checked locally and I can confirm this also fixes the 405 controller errors.
Not sure about the PR build test failures though.

sutaakar:
Adjusted exclusions for depending modules.

cristianonicolai:
@sutaakar this also looks good, but now that https://github.com/kiegroup/kie-wb-common/pull/2467 got merged, could you also create a PR reverting that change with the same branch? so we can test the entire fix?

sutaakar:
Unfortunately the RestEasy dependencies are required for RestEasy960Util class, residing in kie-server-common module. The class provides workaround for https://issues.jboss.org/browse/RESTEASY-960 which is still actual for Java 8.
It is required for both kie-server-rest-common and kie-server-controller-rest modules. 

Makes me think about creating a separate module, providing common functionality related to REST for both Kie server and controller, could be called for example kie-server-rest-tools. The module kie-server-common would stay, but there would be just KeyStoreHelperUtil, used both in Kie server services and Kie controller impl. That way we could get rid of RestEasy dependencies in services modules and make them available just for rest modules. What do you think about it?

cristianonicolai:
@sutaakar sounds reasonable. @mswiderski wdyt?

mswiderski:
And why not to reuse https://github.com/kiegroup/droolsjbpm-integration/tree/master/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common?

sutaakar:
My intention was to keep kie-server-controller-rest separated from kie-server-rest-common as right now there is no dependency between them.
But it is an option too.

MarianMacik:
Ignore Spring Boot build please, I am just tuning it.

mareknovotny:
could we  close this PR?

PR #1726, opened 2019-02-13
JBPM-8149: Add missing jaxb-xjc also to Kie server ee8

sutaakar:


sutaakar:
Test failures not related.

PR #1729, opened 2019-02-14
Fix duplicated classes

MarianMacik:
@cimbalek This should fix the build.


MarianMacik:
Duplicated classes check passed.

PR #1731, opened 2019-02-15
Enable Prometheus metrics also for SpringBoot

sutaakar:
And fix related issues.
@lucamolteni @MarianMacik Please take a look.

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/pom.xml
@@ -318,6 +318,7 @@
       <properties>
         <kieserver.drools.enabled>true</kieserver.drools.enabled>
         <kieserver.dmn.enabled>true</kieserver.dmn.enabled>
+        <kieserver.prometheus.enabled>true</kieserver.prometheus.enabled>
       </properties>
     </profile>
   </profiles>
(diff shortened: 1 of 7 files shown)


lucamolteni:
Also putting @vblagoje in the loop

Hey @sutaakar, great job!

TBH I thought the current implementation was already working as I used it in

 https://github.com/lucamolteni/example-prometheus 

Does this explain why I had to pass the system property at the startup?

```
java -Dorg.kie.prometheus.server.ext.disabled=false -jar target/example-prometheus-service-1.0-SNAPSHOT.jar
```

sutaakar:
@lucamolteni Yes, SpringBoot implementation calls initialization code of Kie server. All the extensions are initialized there, including check if they are active, otherwise they are skipped completely.
By default all extensions are active, apart from prometheus.

lucamolteni:
ok @sutaakar thank you

lucamolteni:
Yes @MarianMacik I can see the problem, but how's that different from the Drools extension?

https://github.com/kiegroup/droolsjbpm-integration/blob/56a1ab0b2a64e89c6016935dbcc116148dad0757/kie-spring-boot/kie-spring-boot-autoconfiguration/kie-server-spring-boot-autoconfiguration-drools/src/main/java/org/kie/server/springboot/autoconfiguration/drools/DroolsKieServerAutoConfiguration.java

In SpringBoot is `kieserver.drools.enabled`and in the kie-server it's `org.drools.server.ext.disabled`

MarianMacik:
The issue is that the one for drools is active by default, so in springboot system property is not found, so then it is active (default value) and then you just set Spring Boot property kieserver.drools.enabled to either true/false whether you want to have it run or not.

With Prometheus when you activate only Spring Boot property, the bean wrapping the extension would be created, but the extension won't be active, because the system property is missing and the default is false.

MarianMacik:
I will merge this to have functional tests and we can then unify it in some way.

PR #1733, opened 2019-02-15
Rhpam 1875a

lanceleverich:
Removed a duplicated dependency in kie-maven-plugin/pom.xml


lanceleverich:
Still has wrong dependency in it

PR #1734, opened 2019-02-15
Fix springboot on JDK 11 (jax-ws-api)

MarianMacik:
depends on https://github.com/kiegroup/jbpm/pull/1443

2 changes:
* default application.properties of integ-tests-sample shouldn't use swagger, since it is not configured (dismisses the warning)
* kie-server-spring-boot-sample will get JAX-WS API from jbpm-workitems-bpmn2 (previously it wasn't the root module who needed that dependency)


MarianMacik:
jenkins retest this

MarianMacik:
jenkins retest this

PR #1735, opened 2019-02-19
KieServerInfo has KieServerMode.REGULAR as a default KIE Server mode

MarianMacik:
Default KIE Server mode is KieServerMode.DEVELOPMENT, so we should reflect it in KieServerInfo entity as well.

@mswiderski I have also noticed that although by default DEVELOPMENT mode is activated, which will allow undeployment while there are still active process instances present, it won't abort them by default. Thus leaving the developer in a possibly inconsistent situation. Shouldn't we make this aborting default in DEVELOPMENT mode? Currently it is not very safe. Perhaps explicitly allowing to undeploy without aborting of processes would be better?

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/KieServerInfo.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/KieServerInfo.java
@@ -36,7 +36,7 @@ public class KieServerInfo {
 
     private List<Message> messages;
 
-    private KieServerMode mode = KieServerMode.REGULAR;
+    private KieServerMode mode = KieServerMode.DEVELOPMENT;
     
     public KieServerInfo() {
         super();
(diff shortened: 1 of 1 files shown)


mswiderski:
looks ok, is the test failure related? @MarianMacik 

@pefernan any comments?

MarianMacik:
@jiripetrlik is the one test fail related? I don't think so, but better to double-check.

jiripetrlik:
@MarianMacik 
Seems like a random fail.

pefernan:
@MarianMacik good catch +1 for me, at the begining I thought that REGULAR would be the default and it seems that I missed that change.

I also think that the error is random, but let's trigger the test again and see if it continues appearing.

Regarding aborting instances  by default or not, we had some conversations regarding that. We wanted to offer both options keep instances active or abort them. We preferred not to have redeploy option (update + abort) also for security but we can talk about that and change it.

pefernan:
jenkins retest this please

pefernan:
@MarianMacik @mswiderski green!

PR #1740, opened 2019-02-21
JBPM-7945 - Use Wildfly container for testing controller standalone d…

cristianonicolai:
…istribution


sutaakar:
@cristianonicolai Please update classifier in controller deployable to ee7.

cristianonicolai:
@sutaakar sure, but I would prefer to do a separated PR for that.

PR #1744, opened 2019-02-21
AF-1865: [Streamline Development Lifcycle] Fix backguard compatibilities and improve 'dev mode' deployment policy

pefernan:
This changes makes Kie Server start in regular mode by default, this will avoid backward compatibility issues.

Also allows deploying -SNAPSHOT and non-SNAPSHOT  modules int dev mode kie-servers.

@tomasdavidorg @MarianMacik @mswiderski could you please take a look?


porcelli:
solution has been revisited, a new PR will be sent

PR #1750, opened 2019-02-26
[DROOLS-3700] implement osgi support on constraint-parser

mariofusco:


kie-osgi/kie-karaf-features/pom.xml
```diff
@@ -84,6 +84,8 @@
 
     <!-- Overwrites for Fuse -->
     <fuse.version.com.h2database>1.3.168</fuse.version.com.h2database>
+
+    <karaf.version.com.github.javaparser>3.10.2</karaf.version.com.github.javaparser>
```
jiripetrlik: Is this dependency necessary only in OSGI? If not, can we keep version on some central place in Kie project. If we keep it only in OSGI features it would be easy to forget update.

PR #1765, opened 2019-03-06
[DROOLS-3708] Support Mvel Validator in the kie-maven-plugin

lucamolteni:
.


lucamolteni:
@baldimir @mariofusco 

Please check this: it has two new parameters to enable the mvel validation (and to enable it preserving DRL in the compiled class)
It's an expensive computation as it requires a second compilation which is hard to avoid.

PR #1767, opened 2019-03-08
JBPM-8306: Process diagram images are hardcoded to large sizes

nmirasch:
Added viewBox attribute to svg returned by Process Instance image sevices

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -11,12 +11,11 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
-*/
+ */
 
 package org.kie.server.services.jbpm.ui;
 
 import java.io.ByteArrayInputStream;
-import java.io.UnsupportedEncodingException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -25,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.jbpm.process.svg.SVGImageProcessor;
+import org.jbpm.process.svg.processor.SVGProcessor;
 import org.jbpm.services.api.ProcessInstanceNotFoundException;
 import org.jbpm.services.api.RuntimeDataService;
 import org.jbpm.services.api.model.NodeInstanceDesc;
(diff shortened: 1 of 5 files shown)


nmirasch:
@mswiderski @cristianonicolai @jstastny-cz could you take a look? Thanks!

cristianonicolai:

cristianonicolai:
@nmirasch should width and height attributes be removed then? That might have an impact in BC, as the diagram will be then fitted to the div size.

nmirasch:
@cristianonicolai maybe just keep a boolean parameter to the services to ask for the viewBox addition and remove width and height in that case, keeping the same calls for BC. Is that a good approach?

nmirasch:
@cristianonicolai @mswiderski @jstastny-cz  added that boolean parameter to ask for the svgViewBox attribute addition, that implies the removal of width and height allowing the external resizing. This PR just change the svg treatment in case this new parameter is present and true, otherwise it remains the same behaviour as now.

cristianonicolai:
@nmirasch I'm not sure you actually need to add the new parameter, I was just concerned that the image would now fit in the div area inside Business Central. I think in the we might actually need to enforce that the div has a minimum size instead. Another point, should this changes be done here or at the designer level? @romartin wdyt?

romartin:
@cristianonicolai @nmirasch I think it's fine doing this just on the widget that renders the SVG image generated by the process designer, so it's fine for me going this way... lemme explain:
The designer generates an SVG image which size matches with bounding box generated by its content, so it's up to other components (as your svg renderer for html) to change the viewport on the SVG declaration to achieve different transforms, such as scaling.
On the other hand @tiagodolphine built this and he has more detail on the impl, so will let him to take a look a this and check if the impl looks good or could be simplified in some way. 
Lemme know if any Qs on this guys!
Thanks!

MarianMacik:
Ignore Spring Boot build please.

tiagodolphine:
@romartin @cristianonicolai I agree with the current approach, as Roger mentioned the SVG width/height is set to the bounding box generated by its content on the designer. So it is up to the client displaying the SVG to change the width/height to scale according to its needs.

cristianonicolai:
@romartin @tiagodolphine thanks for the info :)

nmirasch:
@cristianonicolai Your suggestion is just to keep only the viewBox removing width and height and let the BC calculate that minimum size (from the viewBox, calculate the width and height  used to the container zoom calculations)? In this case the rest extra parameter woudn't be needed

cristianonicolai:
@nmirasch correct, that's why I was mentioning that this would have some impact on BC. Removing the width and height attributes you get the svg resized in there.

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
```diff
@@ -92,27 +93,39 @@ public ImageServiceBase(RuntimeDataService dataService, Map<String, ImageReferen
...
         containerId = registry.getContainerId(containerId, ContainerLocatorProvider.get().getLocator());
 
         String imageSVGString = null;
         byte[] imageSVG = getProcessImageAsBytes(containerId, processId);
         if (imageSVG != null) {
-            try {
-                imageSVGString = new String(imageSVG, "UTF-8");
-            } catch (UnsupportedEncodingException e) {
-                logger.debug("UnsupportedEncodingException while building process image due to {}", e.getMessage());
+            if (addSvgViewBox) {
```
jstastny-cz: can't we handle this uniformly? i.e. remove the else branch completely and rely on the addSvgViewBox variable being passed to processor.getSVG(boolean) in both cases? it would help maintain encapsulation ...

nmirasch:
updated, have to be merged at the same time as https://github.com/kiegroup/jbpm-wb/pull/1323

nmirasch:
rebased

cristianonicolai:
@jstastny-cz happy for this to be merged?

PR #1712, opened 2019-02-05
[JBPM-7974] kie-server not removed from the controller view if server  is killed or crashes

elguardian:
added logic to the standalone controller to remove servers that were killed or crashed

PR #1772, opened 2019-03-13
[JBPM-7974] kie-server not removed from the controller view if server is killed or crashes (7.18.x)

elguardian:
added logic to the standalone controller

cherry picked from #1712


elguardian:
@mareknovotny plz remove the do not merge. let see if this can be in 7.3.z

PR #1773, opened 2019-03-13
[JBPM-8348] Tomcat doesn't correctly stop Kie server controller healt hcheck

elguardian:
starting/stopping the executor service in the listener

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/common/KieControllerServletListener.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/common/KieControllerServletListener.java
@@ -16,6 +16,7 @@ package org.kie.server.controller.common;
 
 import java.util.ServiceLoader;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ExecutorService;
 
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/java/org/kie/server/controller/common/KieControllerServletListener.java
```diff
@@ -49,6 +51,7 @@ public void contextDestroyed(ServletContextEvent event) {
             logger.debug("Template storage {} closed successfully", storage);
         }
         healthCheck.stop();
+        executorService.shutdown();
```
sutaakar: Wouldn't it have more sense to use shutdownNow() and not to wait for healthcheck to stop by itself?

elguardian: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#shutdown()

shutdown does not say how it will try to kill a thread. so it is better to terminate the task and then shutdown. IMHO

sutaakar: Yes, but shutdown() doesn't wait until the task is stopped. Considering that it will take up to 5 seconds (for default configuration) for the task to stop, the result is that the application server will stop sooner than the healtcheck thread ends.
Or possibly awaitTermination() can be used to wait until the healthcheck task finishes.

elguardian:
I see your point @sutaakar. Changed

PR #1776, opened 2019-03-18
PLANNER-1443: Monitoring of OptaPlanner on kie-server

vblagoje:
See https://issues.jboss.org/browse/PLANNER-1443 for more details.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
@@ -28,11 +28,15 @@ import org.kie.server.api.model.instance.ScoreWrapper;
 import org.kie.server.api.model.instance.SolverInstance;
 import org.kie.server.api.model.instance.SolverInstanceList;
 import org.kie.server.services.api.KieContainerInstance;
+import org.kie.server.services.api.KieServerExtension;
 import org.kie.server.services.api.KieServerRegistry;
 import org.kie.server.services.impl.KieContainerInstanceImpl;
+import org.kie.server.services.prometheus.PrometheusKieServerExtension;
+import org.kie.server.services.prometheus.PrometheusMetricsSolverListener;
 import org.optaplanner.core.api.score.Score;
 import org.optaplanner.core.api.solver.Solver;
 import org.optaplanner.core.api.solver.SolverFactory;
+import org.optaplanner.core.impl.solver.AbstractSolver;
 import org.optaplanner.core.impl.solver.ProblemFactChange;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -95,6 +99,8 @@ public class SolverServiceBase {
                     Solver<Object> solver = solverFactory.buildSolver();
 
                     sic.setSolver(solver);
+                    registerListener(solver, solverId);
+
                     updateSolverInstance(sic);
 
                     solvers.put(instance.getSolverInstanceKey(),
@@ -502,6 +508,13 @@ public class SolverServiceBase {
                 }
             }
         });
+    }
+
+    private void registerListener(Solver solver, String solverId) {
+        KieServerExtension extension = context.getServerExtension(PrometheusKieServerExtension.EXTENSION_NAME);
+        if (extension != null) {
+            ((AbstractSolver) solver).addPhaseLifecycleListener(new PrometheusMetricsSolverListener(solverId));
+        }
     }
 
     private void terminateSolverEarly(SolverInstanceContext sic) {
--- a/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
@@ -17,7 +17,9 @@ package org.kie.server.services.prometheus;
 import java.util.stream.IntStream;
 
 import io.prometheus.client.Counter;
+import io.prometheus.client.Gauge;
 import io.prometheus.client.Histogram;
+import io.prometheus.client.Summary;
 import org.apache.commons.lang3.ArrayUtils;
 
 public class PrometheusMetrics {
@@ -91,4 +93,33 @@ public class PrometheusMetrics {
     Histogram getDroolsEvaluationTimeHistogram() {
         return droolsEvaluationTimeHistogram;
     }
+
+
+    private static final Summary optaPlannerSolverDuration = Summary.build()
+            .name("solver_duration_seconds")
+            .help("Time in seconds it took solver to solve the constraint problem")
+            .labelNames("solver_id").register();
+
+    Summary getOptaPlannerSolverDuration() {
+        return optaPlannerSolverDuration;
+    }
+
+    private static final Summary optaPlannerSolverScoreCalculationSpeed = Summary.build()
+            .name("solver_score_calculation_speed")
+            .help("Number of moves per second for a particular solver solving the constraint problem")
+            .labelNames("solver_id").register();
+
+    Summary getOptaPlannerSolverScoreCalculationSpeed() {
+        return optaPlannerSolverScoreCalculationSpeed;
+    }
+
+    private static final Gauge optaPlannerSolverCount = Gauge.build()
+            .name("solvers_running")
+            .help("Number of solvers currently running")
+            .register();
+
+    Gauge getOptaPlannerSolverCount() {
+        return optaPlannerSolverCount;
+    }
+
 }
--- a/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetricsSolverListener.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetricsSolverListener.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.server.services.prometheus;
+
+import io.prometheus.client.Summary;
+import org.optaplanner.core.impl.phase.event.PhaseLifecycleListenerAdapter;
+import org.optaplanner.core.impl.solver.scope.DefaultSolverScope;
+
+public class PrometheusMetricsSolverListener extends PhaseLifecycleListenerAdapter<Object> {
+
+    private final PrometheusMetrics metrics;
+    private final String solverId;
+    private Summary.Timer timer;
+
+    public PrometheusMetricsSolverListener(String solverId) {
+        this(solverId, PrometheusKieServerExtension.getMetrics());
+    }
+
+    public PrometheusMetricsSolverListener(String solverId, PrometheusMetrics metrics) {
+        this.solverId = solverId;
+        this.metrics = metrics;
+    }
+
+    @Override
+    public void solvingStarted(DefaultSolverScope solverScope) {
+        metrics.getOptaPlannerSolverCount().inc();
+        timer = metrics.getOptaPlannerSolverDuration().labels(solverId).startTimer();
+    }
+
+    @Override
+    public void solvingEnded(DefaultSolverScope solverScope) {
+        metrics.getOptaPlannerSolverCount().dec();
+        metrics.getOptaPlannerSolverScoreCalculationSpeed().labels(solverId).observe(getScoreCalculationSpeed(solverScope));
+        timer.observeDuration();
+    }
+
+    private double getScoreCalculationSpeed(DefaultSolverScope solverScope) {
+        long timeMillisSpent = solverScope.calculateTimeMillisSpentUpToNow();
+        // Avoid divide by zero exception on a fast CPU
+        return solverScope.getScoreCalculationCount() * 1000L / (timeMillisSpent == 0L ? 1L : timeMillisSpent);
+    }
+}
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
@@ -16,6 +16,7 @@
     <java.module.name>org.kie.server.itests.optaplanner</java.module.name>
     <org.drools.server.ext.disabled>false</org.drools.server.ext.disabled>
     <org.optaplanner.server.ext.disabled>false</org.optaplanner.server.ext.disabled>
+    <org.kie.prometheus.server.ext.disabled>false</org.kie.prometheus.server.ext.disabled>
   </properties>
 
   <dependencies>
@@ -185,6 +186,7 @@
               <org.optaplanner.server.ext.disabled>${org.optaplanner.server.ext.disabled}</org.optaplanner.server.ext.disabled>
               <org.jbpm.case.server.ext.disabled>true</org.jbpm.case.server.ext.disabled>
               <org.kie.dmn.server.ext.disabled>true</org.kie.dmn.server.ext.disabled>
+              <org.kie.prometheus.server.ext.disabled>${org.kie.prometheus.server.ext.disabled}</org.kie.prometheus.server.ext.disabled>
               <kie.server.testing.kjars.build.settings.xml>${kie.server.testing.kjars.build.settings.xml}</kie.server.testing.kjars.build.settings.xml>
             </systemPropertyVariables>
           </configuration>
@@ -226,6 +228,7 @@
               <org.jbpm.case.server.ext.disabled>true</org.jbpm.case.server.ext.disabled>
               <org.kie.dmn.server.ext.disabled>true</org.kie.dmn.server.ext.disabled>
               <org.drools.server.ext.disabled>${org.drools.server.ext.disabled}</org.drools.server.ext.disabled>
+              <org.kie.prometheus.server.ext.disabled>${org.kie.prometheus.server.ext.disabled}</org.kie.prometheus.server.ext.disabled>
             </systemProperties>
           </container>
         </configuration>
@@ -268,6 +271,16 @@
         <dependency>
           <groupId>org.kie.server</groupId>
           <artifactId>kie-server-rest-optaplanner</artifactId>
+          <scope>test</scope>
+        </dependency>
+        <dependency>
+          <groupId>org.kie.server</groupId>
+          <artifactId>kie-server-rest-prometheus</artifactId>
+          <scope>test</scope>
+        </dependency>
+        <dependency>
+          <groupId>org.kie.server</groupId>
+          <artifactId>kie-server-services-prometheus</artifactId>
           <scope>test</scope>
         </dependency>
       </dependencies>
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
@@ -15,13 +15,21 @@
  */
 package org.kie.server.integrationtests.optaplanner;
 
+import java.io.BufferedReader;
+import java.io.StringReader;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.StringTokenizer;
+import java.util.concurrent.TimeUnit;
 
+import org.jboss.resteasy.client.jaxrs.BasicAuthentication;
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.kie.api.KieServices;
@@ -32,13 +40,19 @@ import org.kie.server.api.model.ServiceResponse;
 import org.kie.server.api.model.instance.ScoreWrapper;
 import org.kie.server.api.model.instance.SolverInstance;
 import org.kie.server.client.KieServicesClient;
+import org.kie.server.integrationtests.config.TestConfig;
 import org.kie.server.integrationtests.shared.KieServerAssert;
 import org.kie.server.integrationtests.shared.KieServerDeployer;
 import org.kie.server.integrationtests.shared.KieServerReflections;
 import org.optaplanner.core.api.score.buildin.hardsoft.HardSoftScore;
 import org.optaplanner.core.impl.solver.ProblemFactChange;
 
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
 import static org.junit.Assert.*;
+import org.assertj.core.api.Assertions;
 
 public class OptaplannerIntegrationTest
         extends OptaplannerKieServerBaseIntegrationTest {
@@ -51,6 +65,7 @@ public class OptaplannerIntegrationTest
     private static final String NOT_EXISTING_CONTAINER_ID = "no_container";
     private static final String CONTAINER_1_ID = "cloudbalance";
     private static final String SOLVER_1_ID = "cloudsolver";
+    private static final String SOLVER_PROMETHEUS_ID = "cloudsolver-prometheus";
     private static final String SOLVER_1_CONFIG = "cloudbalance-solver.xml";
     private static final String SOLVER_1_REALTIME_CONFIG = "cloudbalance-realtime-planning-solver.xml";
     private static final String SOLVER_1_REALTIME_DAEMON_CONFIG = "cloudbalance-realtime-planning-daemon-solver.xml";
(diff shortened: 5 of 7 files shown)


kie-ci7:

vblagoje:
@yurloc could you please give it a look and lmk if I am going in the right direction :-)

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
```diff
@@ -87,4 +89,24 @@ Counter getDMNNumberOfEvaluationFailed() {
     Histogram getDroolsEvaluationTimeHistogram() {
         return droolsEvaluationTimeHistogram;
     }
+
+
+    private static final Summary optaPlannerSolverDuration = Summary.build()
```
cristianonicolai: @vblagoje how about defining buckets, don't you need to add them beforehand?

vblagoje: @cristianonicolai Yes, I would love to define buckets but it's hard as we don't know how long will solvers run. It could be anywhere from seconds to hours. If we had some data about the problem a customer is trying to solve and some beforehand knowledge about their duration then yes histograms would make sense. But in this case, I think summaries make more sense and then we just pick the standard quantiles/percentiles so whatever solver durations are - we are ok. I am still open to fine-tuning this idea and to see how actual metrics data comes out so we can pick the right metric type.

vblagoje: I picked summary after I read this [page](https://prometheus.io/docs/practices/histograms/) perhaps more than 10 times to understand the difference between histograms and summaries :-) LMK if you think otherwise.

vblagoje:
@lucamolteni I tried to follow your design pattern where it comes to listeners and structure overall. Give it a quick look once you have a free moment

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
lucamolteni: @vblagoje I feel your pain, I also read that page many times and I had the same problems with the buckets defined in Drools. 
One way could be let the user specify the buckets in the configuration but I haven't explored that possibility yet.

lucamolteni:
> @lucamolteni I tried to follow your design pattern when it comes to listeners and structure overall. Give it a quick look once you have a free moment

It seems fine to me, great job!

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
cristianonicolai: @vblagoje @lucamolteni exactly the same problem I'm facing, that's why I was curious about your approach. Good to know that we're all on the same boat :)

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/SolverServiceBase.java
```diff
@@ -504,6 +510,13 @@ private void solvePlanningProblem(final SolverInstanceContext sic,
         });
     }
 
+    private void registerListener(Solver solver, String solverId) {
+        KieServerExtension extension = context.getServerExtension(PrometheusKieServerExtension.EXTENSION_NAME);
+        if (extension != null) {
+            ((AbstractSolver) solver).addPhaseLifecycleListener(new PrometheusMetricsSolverListener(solverId));
```
yurloc: raw type -> unchecked call.

vblagoje:
> @vblagoje I'd say it makes a good sense. I think it would also make sense to move the listener to optaplanner extension. I think it would be nice if prometheus extension didn't have to depend on optaplanner-core.

Right, I thought about it. But then I have to introduce Prometheus dependencies into optplanner kie services. This way we keep them only in prometheus services project

yurloc:
@vblagoje I see your point. That can be solved adding a new listener interface specific to the optaplanner extension service. It could cover some of the events from PhaseLifeCycleListener (solving started/ended) but also some other events (solver created/destroyed). Moreover, these new type of events could easily include additional information like `solverId` that's need by Prometheus (or more, like container ID if needed). It's important to consider pros and cons and avoid unnecessary complexity but in this case what I propose would make the code more loosely coupled and easier to test and extend.

I think you'll need to introduce some kind of layer of indirection anyway in order to be able to unit test things like:
- When solver is created, then a Prometheus listener is registered.
- Events have the expected solver ID.

vblagoje:
@yurloc I added a test that's scraping Prometheus metrics endpoint. It verifies that Optaplanner metrics attributes are published.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +628,42 @@ public void testTerminateEarly() throws Exception {
                                    SOLVER_1_ID);
     }
 
+    protected WebTarget newRequest(String uriString) {
+
+        ResteasyClient httpClient = new ResteasyClientBuilder()
+                    .establishConnectionTimeout(10, TimeUnit.SECONDS)
```
yurloc: Deprecated. Please use `connectTimeout()` (https://github.com/resteasy/Resteasy/commit/85c1df23d05a7314ae91c0d42b06406174633d1b).

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +628,42 @@ public void testTerminateEarly() throws Exception {
...
+                    .socketTimeout(10, TimeUnit.SECONDS)
```
yurloc: Deprecated. Please use `readTimeout()` (https://github.com/resteasy/Resteasy/commit/85c1df23d05a7314ae91c0d42b06406174633d1b).

vblagoje: +1 done @yurloc

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
vblagoje: @yurloc +1 agreed. Done

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetricsSolverListener.java
```diff
@@ -42,7 +42,13 @@ public void solvingStarted(DefaultSolverScope solverScope) {
...
     public void solvingEnded(DefaultSolverScope solverScope) {
         metrics.getOptaPlannerSolverCount().dec();
+        metrics.getOptaPlannerSolverScoreCalculationSpeed().labels(solverId).observe(getScoreCalculationSpeed(solverScope));
         timer.observeDuration();
     }
 
+    private double getScoreCalculationSpeed(DefaultSolverScope solverScope) {
+        long timeMillisSpent = solverScope.calculateTimeMillisSpentUpToNow();
+        // Avoid divide by zero exception on a fast CPU
+        return solverScope.getScoreCalculationCount() * 1000L / (timeMillisSpent == 0L ? 1L : timeMillisSpent);
```
yurloc: This formula will give results similar to `1M / 0 = 1M` (for 1 million calculations in "no time"), which is incorrect in my opinion. Such solver would **appear to be slower** than solver with same configuration running on a slower machine, where for example `1M / 0.1 = 10M`.

My proposal is to rewrite the formula so that it gives `<anything> / 0 = Double.MAX_VALUE`. Do you agree?

vblagoje: @yurloc TBH, I have not given this enough thought - I just rewrote what I found in the original method implementation at https://github.com/kiegroup/optaplanner/blob/master/optaplanner-core/src/main/java/org/optaplanner/core/impl/solver/scope/DefaultSolverScope.java#L210

yurloc: Interesting. Maybe @ge0ffrey had good reasons to write it that way. Without explanation it doesn't make sense to me, though.

rsynek: @yurloc may I ask how you got the `1M / 0.1`? The divisor is a `long` type. Not sure what's better - Double.MAX_VALUE is not precise (no computation takes 0 time), nor making the divisor `1L` is. If we don't want to measure nanoseconds.

yurloc: @rsynek `0.1` was meant to represent a very small amount of time larger than zero. I didn't realize we're using milliseconds. In that case, `1` has the same function but I'd still argue that `x / 0 = double.max` is much more precise approximation than `x / 0 = x`.

ge0ffrey: @yurloc That upstream optaplanner-code looks fishy indeed on first sight. This code should 100% match that upstream code, after any change we agree. Some notes:
- Because timeMillisSpent is a long, so it's indeed either 0L or 1L, never 0.1, so that reasoning is void.
- A computer's internal clock has an accuracy, depending on the model and the OS. For example I used to see an accuracy of 15 milliseconds a lot 15 years ago on Windows. That means that something that takes 8 milliseconds will randomly resolve to 0ms or 15ms timeMillisSpent.
- If the timeMillisSpent resolves to 0ms, this doesn't mean that the score calculation was infinitely fast. It's not positive infinity.
- In the same reasoning, Double.MAX_VALUE is going to be far more off than just taking the score calculation count per second (so the score calc multiplied by 1000L and divided by 1L).
-- For example: we have a calc count of 36 over 3 milliseconds. That should end up as **12 per second**.
--- But the clock gave back 0 milliseconds due to the 15ms accuracy. What is the most correct number?
---- 36 * 1000L / 0L = Positive infinity per second
---- Double.MAX_VALUE per second, so somewhere around "2 to the power 1024"  per second
---- 36 * 1000L / 1L = 36 per second
--- The last one is the the closest.

So, given this reasoning, **I believe the upstream optaplanner-core is correct**. Wdyt?

ge0ffrey: Upstream code changes I am lumping on the PR for the new SolverListener:

```

        // Avoid divide by zero exception on a run faster than the clock accuracy (which can be up to 15 milliseconds)
        // In that case, the 0 milliseconds gets rounded up to 1 millisecond,
        // which is closer to the correct result than positive infinity or MAX_VALUE.
        // Ideally, we'd replace that 0 by the half of the clock accuracy instead, but that information is unavailable.
        return getScoreCalculationCount() * 1000L / (timeMillisSpent == 0L ? 1L : timeMillisSpent);
```

The last line is unchanged.

vblagoje: I'll add this comment to the current code on this branch

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +628,42 @@ public void testTerminateEarly() throws Exception {
...
+                    .connectTimeout(10, TimeUnit.SECONDS)
+                    .readTimeout(10, TimeUnit.SECONDS)
+                    .build();
+
+        WebTarget webTarget = httpClient.target(uriString);
+        webTarget.register(new BasicAuthentication(TestConfig.getUsername(), TestConfig.getPassword()));
+        return webTarget;
+    }
+
+    private void checkPrometheusEndpointResponse() {
```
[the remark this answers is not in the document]
vblagoje: Ok @rsynek it makes sense what you suggest. I'll add a test that specifically verifies the metric values taken from Prometheus. I'll try to make it just as a method but in case we need to make a separate class test - I'll do that. And yes, if Optaplanner is disabled, no OptaPlanner metrics are published. Thanks for review. I'll ping you with an update soon

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetricsSolverListener.java
yurloc: @ge0ffrey Thanks for detailed explanation. It makes much more sense now. I don't think the formula is *correct*. It's a good approximation but it is essentially wrong because we *don't know* the actual value. On a second thought, `Double.MAX_VALUE` isn't a good idea because it's indeed a worse approximation.

There's another option that's better than `Double.MAX_VALUE` and it is `Double.NaN`. It's the most *correct* result. I would say it's definitely more correct than `x / 1`. The problem is that it lacks any information about `x`, which makes it a worse approximation than `x / 1`.

| Formula | Correctness score | Approximation score |
| --- | --- | --- |
| x / 1 | Poor | Good |
| NaN | Good | Nil |

So, I can now see both approaches as acceptable. The choice depends on context. I would go for `NaN` since I suppose the result isn't used for further computation, it's just displayed to the user, and I'd like to be made aware of the exceptional (and rare) condition. In practice, since there's no real reason for a change, let's stick with the old formula :slightly_smiling_face:.

vblagoje:
@yurloc @rsynek I squashed the changes and prepared this PR for the final review. @rsynek in integration test I added explicit tests to verify that all metrics are published - along with labels but I did not check their values. I'll see now if I can make some generic values checks that will not be too brittle

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -285,6 +297,42 @@ public void testExecuteSolver() throws Exception {
...
+        solverClient.solvePlanningProblem(CONTAINER_1_ID,
+                SOLVER_PROMETHEUS_ID,
+                planningProblem);
+
+        solverInstance = solverClient.getSolver(CONTAINER_1_ID,
+                SOLVER_PROMETHEUS_ID);
+
+        // solver should finish in 5 seconds, but we wait up to 15s before timing out
+        for (int i = 0; i < 5 && solverInstance.getStatus() == SolverInstance.SolverStatus.SOLVING; i++) {
+            Thread.sleep(3000);
```
rsynek: Thanks, the test should work, but I am a bit concerned about the time the test spends waiting on Thread.sleep(). Also it's not clear at the first glance what the timeout of the test is before someone reads your comment.

If I may offer an alternative solution:
```
final long SOLVER_STATUS_CHECK_PERIOD = 1000L;
while(solverInstance.getStatus() == SolverInstance.SolverStatus.SOLVING) {
 Thread.sleep(SOLVER_STATUS_CHECK_PERIOD);
}
```
and putting a timeout directly on the `@Test` annotation. What do you think about this approach?

vblagoje: @rsynek I copied that hack from another test :-) and focused on this one. Should we correct both tests?

vblagoje: @rsynek I updated my test only and checked for actual metric values published

rsynek: :) I know, I saw the test. If you want to improve the original test too, that would be great. If you find out that these tests differ just by (not)calling the `checkPrometheusMetrics()`, could be worth also reduce the duplication by introducing a new private method containing logic of the test.

If you find it too complicated, feel free to focus just on the tests related to Prometheus and let me know - I will file a Jira about the test refactoring.

Thanks!

vblagoje: Yes, it is a bit different as it uses a different solver. But I can refactor that as well and reuse tests. I'll do that and we can create an additional JIRA for refactoring of these hardcoded 
timeouts

vblagoje: Here it is @rsynek b75fb94e8081eb4e3c1273d2225761ac5be578ae I think my work on this PR is done now :-) LMK

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +644,77 @@ public void testTerminateEarly() throws Exception {
...
+        }
+    }
+
+    private void checkPrometheusMetricsLine(String line) {
+        if (line.startsWith("solver_duration_seconds_count")
+                || line.startsWith("solver_duration_seconds_sum")
+                || line.startsWith("solver_score_calculation_speed_count")
+                || line.startsWith("solver_score_calculation_speed_sum")
+        ) {
+            // all metric values should be a positive double
```
rsynek: The comment is a bit misleading, as you pass a value `1` for comparison. For a positive double I would rather expect `0.0` to be passed to the `checkMetric` method, is that correct?

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +644,77 @@ public void testTerminateEarly() throws Exception {
...
+            checkMetric(line, 1);
+        } else if (line.startsWith("solvers_running")) {
+            //when test is done we should have zero solvers running
```
rsynek: This comment is misleading for me too - the value you pass there is used in `>=` comparison; so if the value of the metric is e.g. 5, it's still >= 0

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -615,6 +644,77 @@ public void testTerminateEarly() throws Exception {
...
+            checkMetric(line, 0);
+        }
+
+    }
+
+    private void checkMetric(String line, double value) {
+        StringTokenizer st = new StringTokenizer(line, " ");
+        Assert.assertEquals(2, st.countTokens());
+        st.nextToken();
+        Assert.assertTrue(Double.parseDouble(st.nextToken()) >= value);
```
rsynek: Generally it's good to avoid assertTrue as much as possible, because it doesn't tell much. 

Assert.equals or Assert.greaterThan serve much better, as they inform you about what value differed from the expected value. With assertTrue, you have only true/false.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -251,38 +266,52 @@ public void testGetSolvers() {
...
     public void testExecuteSolver() throws Exception {
+        testExecuteSolverHelper(SOLVER_1_ID, false);
+    }
+
+    @Test(timeout = 15_000)
+    public void testPrometheusMetrics() throws Exception {
+        testExecuteSolverHelper(SOLVER_PROMETHEUS_ID, true);
+    }
+
+    private void testExecuteSolverHelper(String solverID, boolean checkPrometheusMetrics) throws Exception {
```
rsynek: Just Nitpicking: instead of passing boolean flag, you can call the `checkPrometheusMetrics` method from the end of `testPrometheusMetrics` and skip it in the `testExecuteSolver`.

vblagoje:
@rsynek I agree with your suggestions. Latest changes are at 31699f2c99e99847b4017078bd9bb5f14b04193b

vblagoje:
Thanks @rsynek for your help and guidance!

vblagoje:
@sutaakar if you find some time today would you also please review this PR? Would love to have it merged sometime soon.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
```diff
@@ -156,6 +157,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+
+    <dependency>
```
sutaakar: Is there any reason for declaring this dependency here? I see that it is already used in local-test-run profile.

sutaakar:
@vblagoje I have briefly looked on the changes and I didn't spot any immediate issue apart from one minor, see comment.
However I am neither experienced with Optaplanner nor Prometheus (yet), so I cannot fully evaluate the implementation.

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/pom.xml
vblagoje: +1 good catch, let me check

vblagoje: @sutaakar fixed, thank you

vblagoje:
Thanks @sutaakar and @rsynek . I squashed all the commits into one and rebased to master. Let's see what CI/Jenkins says.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetricsSolverListener.java
cristianonicolai: @vblagoje I created a function to concentrate this conversion, as Prometheus suggests as good practice to store time-related metrics as seconds, see https://prometheus.io/docs/practices/naming/#base-units and https://github.com/kiegroup/droolsjbpm-integration/pull/1799/files#diff-36b4ccc121dfe965799ba3292507b0eaR45 should we make sure we both share the same approach?

vblagoje:
@sutaakar @rsynek Anything else we need to do to get this PR merged?

rsynek:
@vblagoje no more ideas

sutaakar:
Looks ok to me.

sutaakar:
@yurloc @ge0ffrey Is this PR ok to be merged?

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -251,38 +266,49 @@ public void testGetSolvers() {
...
+    public void testPrometheusMetrics() throws Exception {
+        testExecuteSolverHelper(SOLVER_PROMETHEUS_ID);
+        checkPrometheusMetrics();
+    }
+
+    private void testExecuteSolverHelper(String solverID) throws Exception {
         SolverInstance solverInstance = solverClient.createSolver(CONTAINER_1_ID,
-                                                                  SOLVER_1_ID,
-                                                                  SOLVER_1_CONFIG);
+                solverID,
```
yurloc: @vblagoje It would be perfect if you could avoid changing code format when it's not necessary. Moreover this code was formatted properly according to _KIE Java Convetions_. You can import the [rules](https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration) to your IDE if you use IntelliJ IDEA or Eclipse.

Just a minor issue, fix is optional. We can merge without this but since you'll need to update the PR once more to resolve the conflict in `pom.xml` I thought this was worth mentioning.

vblagoje: @yurloc sure I'll do that. I'll update my IntelliJ IDEA so I can have it for the future commits. I'll resolve pom.xml conflict as well and rebase the branch

vblagoje:
@yurloc @sutaakar @rsynek rebased to master, fixed pom.xml conflict and formatting issue. Let's wait for Jenkins

sutaakar:
@ge0ffrey Can you please take a look if this PR can be merged?

kie-server-parent/kie-server-tests/kie-server-integ-tests-optaplanner/src/test/java/org/kie/server/integrationtests/optaplanner/OptaplannerIntegrationTest.java
```diff
@@ -251,10 +266,20 @@ public void testGetSolvers() {
         assertNull(returnedInstance.getScoreWrapper().toScore());
     }
 
-    @Test
+    @Test(timeout = 15_000)
     public void testExecuteSolver() throws Exception {
+        testExecuteSolverHelper(SOLVER_1_ID);
+    }
+
+    @Test(timeout = 15_000)
```
ge0ffrey: Suggestion: Usually we take 60_000 because some jenkins machine are extremely slow

ge0ffrey:
@vblagoje Would you mind handing any suggestions that make sense for you in a separate PR? This PR is a great step forward, no regressions, so I've merged it.

vblagoje:
@ge0ffrey absolutely. I am sure we'll still fine-tune these metrics but even before...anything you notice - do tell me

PR #1780, opened 2019-03-20
[JBPM-8150] [RHPAM-1975] Allow users to change Kie server id in Deployment config and templates

rhtevan:
In response to request from (https://issues.jboss.org/projects/JBPM/issues/JBPM-8150), this PR includes the implementation remove the coupling between KIE server id and name of KIE server Deployment config, allowing users to choose whatever KIE server id they want, during creation of template or later.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/main/java/org/kie/server/controller/openshift/storage/ServerTemplateConverter.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/main/java/org/kie/server/controller/openshift/storage/ServerTemplateConverter.java
@@ -21,17 +21,16 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.kie.server.api.KieServerConstants;
 import org.kie.server.api.model.KieContainerResource;
 import org.kie.server.api.model.KieScannerResource;
 import org.kie.server.api.model.KieServerConfig;
 import org.kie.server.api.model.KieServerConfigItem;
 import org.kie.server.api.model.KieServerMode;
-import org.kie.server.controller.api.KieServerControllerConstants;
 import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
 import org.kie.server.controller.api.model.spec.Capability;
 import org.kie.server.controller.api.model.spec.ContainerConfig;
@@ -44,6 +43,21 @@ import org.kie.server.services.impl.storage.KieServerState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.kie.server.api.KieServerConstants.CAPABILITY_BPM;
+import static org.kie.server.api.KieServerConstants.KIE_DROOLS_SERVER_EXT_DISABLED;
+import static org.kie.server.api.KieServerConstants.KIE_JBPM_SERVER_EXT_DISABLED;
+import static org.kie.server.api.KieServerConstants.KIE_OPTAPLANNER_SERVER_EXT_DISABLED;
+import static org.kie.server.api.KieServerConstants.KIE_SERVER_ID;
+import static org.kie.server.api.KieServerConstants.KIE_SERVER_LOCATION;
+import static org.kie.server.api.KieServerConstants.KIE_SERVER_MODE;
+import static org.kie.server.api.KieServerConstants.PCFG_KIE_BASE;
+import static org.kie.server.api.KieServerConstants.PCFG_KIE_SESSION;
+import static org.kie.server.api.KieServerConstants.PCFG_MERGE_MODE;
+import static org.kie.server.api.KieServerConstants.PCFG_RUNTIME_STRATEGY;
+import static org.kie.server.controller.api.KieServerControllerConstants.KIE_CONTROLLER_OPENSHIFT_PREFER_KIESERVER_SERVICE;
+import static org.kie.server.services.openshift.api.KieServerOpenShiftConstants.KIE_SERVER_SERVICES_OPENSHIFT_SERVICE_NAME;
+import static org.kie.server.services.openshift.api.KieServerOpenShiftConstants.UNKNOWN;
+
 public class ServerTemplateConverter {
 
     private static final Logger logger = LoggerFactory.getLogger(ServerTemplateConverter.class);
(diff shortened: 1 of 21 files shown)


kie-ci6:

sutaakar:
Jenkins retest this

sutaakar:
The failure seems unrelated

rhtevan:
One timeout failure. Shall we retest it.

sutaakar:
Possibly later, first I would like to make a review so we don't trigger build when possible changes could be done.

kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/main/java/org/kie/server/controller/openshift/storage/ServerTemplateConverter.java
```diff
@@ -109,34 +123,37 @@ public static ServerTemplate fromState(KieServerState state) {
...
+        template.setMode(KieServerMode.valueOf(mode.toUpperCase()));
 
         return template;
     }
 
     public static KieServerState toState(ServerTemplate template) {
         String id = template.getId();
-        KieServerMode mode = template.getMode() == null ? KieServerMode.DEVELOPMENT : template.getMode();
+        String name = template.getName();
+        KieServerMode mode = template.getMode() == null ? KieServerMode.PRODUCTION : template.getMode();
```
sutaakar: Any reason why to switch default mode to PRODUCTION?

sutaakar:
Otherwise looks good to me. I would like to check it in real OpenShift environment too.

kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/main/java/org/kie/server/controller/openshift/storage/ServerTemplateConverter.java
rhtevan: Wasn't clear what is the default value. Checked both the Cloud image scripts, which does NOT set the default, and KIE server implementation, and found this:

https://github.com/rhtevan/droolsjbpm-integration/blob/dd66cf9ebd5c975b3a2788819653f00c5ac970ad/kie-server-parent/kie-server-controller/kie-server-controller-openshift/src/main/java/org/kie/server/controller/openshift/storage/ServerTemplateConverter.java#L68

To align with KIE server implementation, changed it back to DEVELOPMENT.

sutaakar:
Jenkins retest this

sutaakar:
@rhtevan When I start just Workbench image I am not able to create a new container. When I add deployment unit and click on finish the container is not registered in Workbench nor added into config map.

sutaakar:
When I create a config map using Workbench and then I create a new Kie server deployment with corresponding Kie server id and spin it up, then org.kie.server.location in config map is not updated, still pointing to "unknown".

sutaakar:
Otherwise it seems to work as expected.

rhtevan:
> @rhtevan When I start just Workbench image (Kie server is scaled to 0) I am not able to create a new container. When I add deployment unit and click on finish the container is not registered in Workbench nor added into config map.
> 
> When I spin up the Kie server pod then it is possible to create a new container in Workbench.

@sutaakar Don't see such issue with the latest image build. ("LastTagTime": "2019-03-22T13:28:06.897841908-04:00")

rhtevan:
> When I create a config map using Workbench and then I create a new Kie server deployment with corresponding Kie server id and spin it up, then org.kie.server.location in config map is not updated, still pointing to "unknown".

At the moment, the required OpenShift objects: DC, Route and Service are not being automatically generated once BC/WB creates a new KIE server CM. However, those objects can be created manually. In terms of the 'unknown' location situation, needs to check if associated Service and Route are being setup properly.

rhtevan:
@sutaakar Lots of culprits should be polished out this time.  More details at the following JIRA.

https://issues.jboss.org/browse/JBPM-8150?focusedCommentId=13714519&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13714519

sutaakar:
Briefly checked and looks good. Will check on OpenShift instance.
Just a  side note, when doing changes to such big PR please do it as a separate commit. That way it is easier to spot and review changes.

sutaakar:
Jenkins retest this

rhtevan:
@sutaakar Do you think it is ready to merge, or still have something to look into?

sutaakar:
I think it is ready for merge.
@mswiderski Can you please take a look?

kie-server-parent/kie-server-services/kie-server-services-openshift/src/main/java/org/kie/server/services/openshift/api/KieServerOpenShift.java
```diff
@@ -15,29 +15,43 @@
...
+
 public interface KieServerOpenShift {
 
     default boolean isKieServerReady() {
         return KieServerLocator.getInstance().isKieServerReady();
     }
 
     default boolean isDCStable(DeploymentConfig dc) {
-        return "True".equals(dc.getStatus().getConditions().get(0).getStatus()) && dc.getStatus().getUnavailableReplicas() == 0;
+        return "True".equals(dc.getStatus().getConditions().get(0).getStatus())
```
mswiderski: any reason you use mixed case and case sensitive equals here?

rhtevan: If we use 'oc' command login to any running OpenShift cluster, then run the following command:
`
$ oc explain DeploymentConfig.status.conditions.status
FIELD: status <string>

DESCRIPTION:
     Status of the condition, one of True, False, Unknown.
`
The possible value of the string literal listed are '**True**', '**False**', '**Unknown**', which is a bit of unusual for Java developers.

mswiderski: I was more thinking why not to use equalsIgnoreCase just to be on safe side so to say

rhtevan: To be cautious is always good. Changed it to use 'equalsIgnoreCase'. Who knows OpenShift may change its spec all of sudden in future to align with what developer get used to, like 'true' and 'false'.

rhtevan: @mswiderski @sutaakar 
Shall we merge it or still have more things to look into?

mswiderski:
Jenkins retest this

mswiderski:
looks like the build was not run on last changes so let's rerun that and then merge

PR #1781, opened 2019-03-21
JBPM-8116: It is not possible to see SLA violations in business centr…

nmirasch:
…al UI

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskInstance.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskInstance.java
@@ -66,6 +66,10 @@ public class TaskInstance {
     private String processId;
     @XmlElement(name="task-container-id")
     private String containerId;
+    @XmlElement(name="sla-compliance")
+    private Integer slaCompliance;
+    @XmlElement(name="sla-due-date")
+    private Date slaDueDate;
 
     @XmlElementWrapper(name="potential-owners")
     @XmlElement(name="task-pot-owners")
@@ -243,6 +247,22 @@ public class TaskInstance {
     public void setContainerId(String containerId) {
         this.containerId = containerId;
     }
+
+    public Integer getSlaCompliance() {
+        return slaCompliance;
+    }
+
+    public void setSlaCompliance(Integer slaCompliance) {
+        this.slaCompliance = slaCompliance;
+    }
+
+    public Date getSlaDueDate() {
+        return slaDueDate;
+    }
+
+    public void setSlaDueDate(Date slaDueDate) {
+        this.slaDueDate = slaDueDate;
+    }
     
     public List<String> getPotentialOwners() {
         return potentialOwners;
@@ -295,6 +315,9 @@ public class TaskInstance {
                 ", processInstanceId=" + processInstanceId +
                 ", processId='" + processId + '\'' +
                 ", containerId='" + containerId + '\'' +
+                ", workItemId=" + workItemId +
+                ", slaCompliance=" + slaCompliance +
+                ", slaDueDate=" + slaDueDate  +
                 '}';
     }
 
@@ -398,6 +421,16 @@ public class TaskInstance {
         
         public Builder containerId(String containerId) {
             taskInstance.setContainerId(containerId);
+            return this;
+        }
+
+        public TaskInstance.Builder slaDueDate(Date slaDueDate) {
+            taskInstance.setSlaDueDate(slaDueDate);
+            return this;
+        }
+
+        public TaskInstance.Builder slaCompliance(Integer slaCompliance) {
+            taskInstance.setSlaCompliance(slaCompliance);
             return this;
         }
 
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/ConvertUtilsTest.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/ConvertUtilsTest.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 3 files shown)


nmirasch:
depends on https://github.com/kiegroup/jbpm/pull/1464
Be able to add the task sla compliance, sla due date and workitemId to TaskInstance
@cristianonicolai @mswiderski @jstastny-cz  Could you take a look? Thanks

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/ConvertUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class ConvertUtilsTest {
+
+    @Test
+    public void testConvertToTask() {
+        UserTaskInstanceDesc userTaskWithSla =
+                new UserTaskInstanceDesc(Long.valueOf(1), "Ready", new Date(), "Task1_name", "Task1_desc", 1,
+                                         "Task1_Owner", "Task1_Creator", "deployment", "Process", 1L, new Date(),
+                                         new Date(), 1L, new Date(), ProcessInstance.SLA_PENDING);
+        UserTaskInstanceDesc userTask =
```
jstastny-cz: There's a minimal chance, that the test will be run around midnight, then the various new Date() calls might fall into different days, effectively leading to failing the test.
Could we transform into dates based on string representation? Also to make sure the Date fields are not swapped, I'd use different dates in the string for each field.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/instance/TaskInstance.java
```diff
@@ -401,6 +424,15 @@ public Builder containerId(String containerId) {
...
 
+        public TaskInstance.Builder slaDueDate(Date slaDueDate) {
+            taskInstance.setSlaDueDate(slaDueDate);
+            return this;
+        }
+
+        public TaskInstance.Builder slaCompliance(Integer slaCompliance) {
+            taskInstance.setSlaCompliance(slaCompliance);
+            return this;
+        }
```
jstastny-cz: we might want to add a newline here.

nmirasch: @jstastny-cz added :)

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/ConvertUtilsTest.java
nmirasch: @jstastny-cz Thanks, but in this case, I don't think that have any issue with the date value regarding the test just checks that the fields have the same date values in both objects

jstastny-cz: Thing is that not all 'new Date()' invocation have to return the same date value (depends on exact time when the constructor is being invoked) .... it is an extreme case, but it's better to prevent it by sticking to fixed values in tests.

nmirasch: @jstastny-cz Yes, I think I understand you but for this test I don't care the date value, The test is just checking they have the same value. Anyway, I'm gone to update this test to use a single new Date()

nmirasch: @jstastny-cz updated

jstastny-cz: @nmirasch thanks for the change, looks good now.
> The test is just checking they have the same value
Still the values could have been different, cause the Date instances in userTaskWithSla and userTask were instantiated in different moment (possibly slipping over midnight) and the test asserts equality over those. I am only trying to illustrate my concern, I've seen tests 'randomly' failing due to this exact issue, thanks again for the change.

nmirasch: @jstastny-cz The test don't compare userTaskWithSla with userTask. The comparison is between userTask and its conversion to TaskInstance ConvertUtils.convertToTask(userTask). But it's ok like this, Thanks for the explanations!

jstastny-cz: :-D I am sorry, I should have read the code more carefully.

nmirasch: @jstastny-cz absolutely, no problem. A lot of things to review!

nmirasch:
rebased

kie-server-parent/kie-server-services/kie-server-services-jbpm/src/test/java/org/kie/server/services/jbpm/ConvertUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jbpm.kie.services.impl.model.UserTaskInstanceDesc;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.server.api.model.instance.TaskInstance;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(MockitoJUnitRunner.class)
```
sutaakar: Just a nitpick, do you need a MockitoJUnitRunner for this test? I don't see any use of mockito here.

nmirasch: agree, I have removed if, thanks!

nmirasch:
@sutaakar rebased and updated the test, could you review this again?

sutaakar:
Test structure looks ok to me now.

PR #1783, opened 2019-03-22
JBPM-8363: EJB Timer error in development mode for existing process during deploy

pefernan:
@jstastny-cz @mswiderski could you please take a look?

This is part of an ensemble please merge with:
https://github.com/kiegroup/jbpm/pull/1466
https://github.com/kiegroup/droolsjbpm-integration/pull/1783

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
@@ -130,7 +130,7 @@ public class JbpmKieServerExtension implements KieServerExtension {
 
     public static final String EXTENSION_NAME = "jBPM";
     private static final String PERSISTENCE_XML_LOCATION = "/jpa/META-INF/persistence.xml";
-    private static final String UPDATING_CONTAINER_PARAM = "jBPMExtensionUpdateContainer";
+    private static final String IS_DISPOSE_CONTAINER_PARAM = "jBPMExtensionIsDisposeContainer";
 
     private static final Logger logger = LoggerFactory.getLogger(JbpmKieServerExtension.class);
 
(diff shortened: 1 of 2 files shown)


kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/JbpmKieServerExtension.java
```diff
@@ -496,6 +497,9 @@ public boolean isUpdateContainerAllowed(String id, KieContainerInstance kieConta
     public void updateContainer(String id, KieContainerInstance kieContainerInstance, Map<String, Object> parameters) {
         // essentially it's a redeploy to make sure all components are up to date,
         // though update of kie base is done only once on kie server level and KieContainer is reused across all extensions
+        // we add this UPDATING_CONTAINER_PARAM
+        parameters.put(UPDATING_CONTAINER_PARAM, Boolean.TRUE);
```
mswiderski: Maybe use such parameter more for the dispose container so the condition would abortInstances or dispose. That way it’s more readable in my opinion

pefernan: @mswiderski you're right, maybe it is not clear enough.. Better now?

MarianMacik:
Just so I understand it:

Before when a pure dispose was made, I had 2 options:
* Processes weren't aborted - default option
* Processes were aborted if `KieServerConstants.KIE_SERVER_PARAM_RESET_BEFORE_UPDATE` was explicitly set to true.

Now, when I purely dispose a container, `IS_DISPOSE_CONTAINER_PARAM` will be always true, so processes will be aborted regardless what is the value of `KieServerConstants.KIE_SERVER_PARAM_RESET_BEFORE_UPDATE`.

Is that right? The behaviour is changing a little bit.

pefernan:
@MarianMacik 
That's right. The issue was that on dev mode when disposing a container (not updating) the active process instances weren't aborted. 
The point is that disposeContainer is called when disposing a container or when updating (dispose + create) so with this param we differenciate if this is a disposal or an update.

PR #1798, opened 2019-04-04
JBPM-8399 - Add new configuration flag that will show a complete erro…

mswiderski:
…r stacktrace of the error returned by kie server in the console

to test it start server with -Dorg.kie.server.stacktrace.included=true

default value is false so it has same behaviour as now.


mswiderski:
@MarianMacik could you please review?

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/UserTaskResource.java
```diff
@@ -298,10 +298,10 @@ public Response exit(@Context HttpHeaders headers,
 
             return createResponse("", v, Response.Status.CREATED, conversationIdHeader);
         } catch (TaskNotFoundException e){
-            return notFound(MessageFormat.format(TASK_INSTANCE_NOT_FOUND, taskId), v, conversationIdHeader);
+            return notFound(errorMessage(e, MessageFormat.format(TASK_INSTANCE_NOT_FOUND, taskId)), v, conversationIdHeader);
```
MarianMacik: Somewhere we also provide a stackTrace for specific exceptions like TaskNotFoundException or PermissionDeniedException, somewhere only for the last, most generic catch blocks. Is that intentional?

mswiderski: yes as the "somewhere" is actually in the lifecycle methods as they can be caused by permission denied underneath

PR #1799, opened 2019-04-05
JBPM-8347 - jBPM Prometheus metrics

cristianonicolai:
Assembly:
- https://github.com/kiegroup/droolsjbpm-integration/pull/1799
- https://github.com/kiegroup/jbpm/pull/1473
- https://github.com/kiegroup/kie-soup/pull/84
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/379

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
@@ -11,13 +11,14 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
-*/
+ */
 
 package org.kie.server.services.casemgmt;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.ServiceLoader;
 
 import org.jbpm.casemgmt.api.CaseRuntimeDataService;
@@ -91,9 +92,8 @@ public class CaseKieServerExtension implements KieServerExtension {
             logger.warn("jBPM extension not found, Case Management cannot work without jBPM extension, disabling itself");
             return;
         }
-        
+
         configureServices(kieServer, registry);
-        
 
         this.services.add(this.caseManagementServiceBase);
         this.services.add(this.caseManagementRuntimeDataService);
@@ -103,9 +103,9 @@ public class CaseKieServerExtension implements KieServerExtension {
 
         initialized = true;
     }
-    
+
     protected void configureServices(KieServerImpl kieServer, KieServerRegistry registry) {
-        
+
         KieServerExtension jbpmExtension = registry.getServerExtension("jBPM");
         List<Object> jbpmServices = jbpmExtension.getServices();
         RuntimeDataService runtimeDataService = null;
@@ -113,22 +113,21 @@ public class CaseKieServerExtension implements KieServerExtension {
         DeploymentService deploymentService = null;
         ProcessInstanceMigrationService processInstanceMigrationService = null;
 
-
-        for( Object object : jbpmServices ) {
+        for (Object object : jbpmServices) {
             // in case given service is null (meaning was not configured) continue with next one
             if (object == null) {
                 continue;
             }
-            if( RuntimeDataService.class.isAssignableFrom(object.getClass()) ) {
+            if (RuntimeDataService.class.isAssignableFrom(object.getClass())) {
                 runtimeDataService = (RuntimeDataService) object;
                 continue;
-            } else if( ProcessService.class.isAssignableFrom(object.getClass()) ) {
+            } else if (ProcessService.class.isAssignableFrom(object.getClass())) {
                 processService = (ProcessService) object;
                 continue;
-            } else if( DeploymentService.class.isAssignableFrom(object.getClass()) ) {
+            } else if (DeploymentService.class.isAssignableFrom(object.getClass())) {
                 deploymentService = (DeploymentService) object;
                 continue;
-            } else if( ProcessInstanceMigrationService.class.isAssignableFrom(object.getClass()) ) {
+            } else if (ProcessInstanceMigrationService.class.isAssignableFrom(object.getClass())) {
                 processInstanceMigrationService = (ProcessInstanceMigrationService) object;
                 continue;
             }
@@ -151,16 +150,16 @@ public class CaseKieServerExtension implements KieServerExtension {
         ((CaseServiceImpl) caseService).setRuntimeDataService(runtimeDataService);
         ((CaseServiceImpl) caseService).setCommandService(new TransactionalCommandService(EntityManagerFactoryManager.get().getOrCreate(persistenceUnitName)));
         ((CaseServiceImpl) caseService).setAuthorizationManager(new AuthorizationManagerImpl(registry.getIdentityProvider(),
-                new TransactionalCommandService(EntityManagerFactoryManager.get().getOrCreate(persistenceUnitName))));
+                                                                                             new TransactionalCommandService(EntityManagerFactoryManager.get().getOrCreate(persistenceUnitName))));
         ((CaseServiceImpl) caseService).setIdentityProvider(registry.getIdentityProvider());
 
         // build case configuration on deployment listener
         CaseConfigurationDeploymentListener configurationListener = new CaseConfigurationDeploymentListener(registry.getIdentityProvider());
 
         // configure case mgmt services as listeners
-        ((KModuleDeploymentService)deploymentService).addListener((CaseRuntimeDataServiceImpl) caseRuntimeDataService);
-        ((KModuleDeploymentService)deploymentService).addListener(configurationListener);
-        
+        ((KModuleDeploymentService) deploymentService).addListener((CaseRuntimeDataServiceImpl) caseRuntimeDataService);
+        ((KModuleDeploymentService) deploymentService).addListener(configurationListener);
+
         CaseInstanceMigrationService caseInstanceMigrationService = new CaseInstanceMigrationServiceImpl();
         ((CaseInstanceMigrationServiceImpl) caseInstanceMigrationService).setCaseRuntimeDataService(caseRuntimeDataService);
         ((CaseInstanceMigrationServiceImpl) caseInstanceMigrationService).setCommandService(new TransactionalCommandService(EntityManagerFactoryManager.get().getOrCreate(persistenceUnitName)));
@@ -182,7 +181,7 @@ public class CaseKieServerExtension implements KieServerExtension {
         ServiceLoader<CaseIdGenerator> generators = ServiceLoader.load(CaseIdGenerator.class);
 
         for (CaseIdGenerator generator : generators) {
-            if (generator.getIdentifier().equals(selectedGenerator))  {
+            if (generator.getIdentifier().equals(selectedGenerator)) {
                 return generator;
             }
         }
@@ -234,13 +233,13 @@ public class CaseKieServerExtension implements KieServerExtension {
         }
         ServiceLoader<KieServerApplicationComponentsService> appComponentsServices = ServiceLoader.load(KieServerApplicationComponentsService.class);
 
-        Object [] services = {
+        Object[] services = {
                 caseManagementServiceBase,
                 caseManagementRuntimeDataService,
                 caseAdminServiceBase,
                 registry
         };
-        for( KieServerApplicationComponentsService appComponentsService : appComponentsServices ) {
+        for (KieServerApplicationComponentsService appComponentsService : appComponentsServices) {
             appComponentsList.addAll(appComponentsService.getAppComponents(EXTENSION_NAME, type, services));
         }
         return appComponentsList;
@@ -254,7 +253,9 @@ public class CaseKieServerExtension implements KieServerExtension {
         if (serviceType.isAssignableFrom(kieContainerCommandService.getClass())) {
             return (T) kieContainerCommandService;
         }
-        return null;
+
+        final Optional<Object> service = services.stream().filter(s -> serviceType.isAssignableFrom(s.getClass())).findFirst();
+        return (T) service.orElse(null);
     }
 
     @Override
@@ -281,15 +282,14 @@ public class CaseKieServerExtension implements KieServerExtension {
     public String toString() {
         return EXTENSION_NAME + " KIE Server extension";
     }
-    
 
     @Override
     public List<Message> healthCheck(boolean report) {
         List<Message> messages = KieServerExtension.super.healthCheck(report);
-        
+
         if (report) {
             messages.add(new Message(Severity.INFO, getExtensionName() + " is alive"));
-        }        
+        }
         return messages;
     }
 }
--- a/kie-server-parent/kie-server-services/kie-server-services-prometheus/README.md
+++ b/kie-server-parent/kie-server-services/kie-server-services-prometheus/README.md
@@ -1,0 +1,62 @@
+Kie Server Prometheus Extension
+===============================
+
+To enable this extension, all that is needed is to set the following system property to false: _org.kie.prometheus.server.ext.disabled=false_.
+By default, this extension is disabled. Once the extension started, you can access the available metrics on _${context}/services/rest/metrics_.  
+For example: 
+
+``` 
+curl -u wbadmin:wbadmin http://localhost:8080/kie-server/services/rest/metrics  
+``` 
+
+# Available metrics for Prometheus
+
+## Kie Server metrics
+
+kie_server_start_time: Kie Server Start Time (Gauge)
+kie_server_deployments_active_total: Kie Server Active Deployments (Gauge)
+kie_server_container_running_total: Kie Server Running Containers (Gauge)
+kie_server_container_started_total: Kie Server Started Containers (Counter)
+
+## Data Sets
+
+kie_server_data_set_registered_total: Kie Server Data Set Registered (Gauge)
+kie_server_data_set_lookups_total: Kie Server Data Set Running Lookups (Gauge)
+kie_server_data_set_execution_time_seconds: Kie Server Data Set Execution Time (Summary)
+kie_server_data_set_execution_total: Kie Server Data Set Execution (Counter)
+
+## Case Management
+
+kie_server_case_duration_seconds: Kie Server Case Duration (Summary)
+kie_server_case_running_total: Kie Server Running Cases (Gauge)
+kie_server_case_started_total: Kie Server Started Cases (Counter)
+
+## Execution Error
+
+kie_server_execution_error_total: Kie Server Execution Errors (Counter)
+
+## Jobs
+
+kie_server_job_running_total: Kie Server Running Jobs (Gauge)
+kie_server_job_scheduled_total: Kie Server Started Jobs (Counter)
+kie_server_job_executed_total: Kie Server Executed Jobs (Counter)
+kie_server_job_cancelled_total: Kie Server Cancelled Jobs (Counter)
+kie_server_job_duration_seconds: Kie Server Job Duration (Summary) 
+
+## Processes
+
+kie_server_process_instance_started_total: Kie Server Started Process Instances (Counter)
+kie_server_process_instance_sla_violated_total: Kie Server Process Instances SLA Violated (Counter)
+kie_server_process_instance_completed_total: Kie Server Completed Process Instances (Counter)
+kie_server_process_instance_running_total: Kie Server Running Process Instances (Gauge)
+kie_server_process_instance_duration_seconds: Kie Server Process Instances Duration (Summary)
+kie_server_work_item_duration_seconds: Kie Server Work Items Duration (Summary)
+
+## Tasks
+
+kie_server_task_added_total: Kie Server Added Tasks (Counter)
+kie_server_task_skipped_total: Kie Server Skipped Tasks (Counter)
+kie_server_task_completed_total: Kie Server Completed Tasks (Counter)
+kie_server_task_failed_total: Kie Server Failed Tasks (Counter)
+kie_server_task_exited_total: Kie Server Exited Tasks (Counter)
+kie_server_task_duration_seconds: Kie Server Task Duration (Summary)
--- a/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusCaseEventListener.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusCaseEventListener.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.services.prometheus;
+
+import java.util.Date;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Gauge;
+import io.prometheus.client.Summary;
+import org.jbpm.casemgmt.api.event.CaseCancelEvent;
+import org.jbpm.casemgmt.api.event.CaseCloseEvent;
+import org.jbpm.casemgmt.api.event.CaseDestroyEvent;
+import org.jbpm.casemgmt.api.event.CaseEvent;
+import org.jbpm.casemgmt.api.event.CaseEventListener;
+import org.jbpm.casemgmt.api.event.CaseStartEvent;
+import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.kie.server.services.prometheus.PrometheusMetrics.millisToSeconds;
+
+public class PrometheusCaseEventListener implements CaseEventListener {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PrometheusCaseEventListener.class);
+
+    protected static final Counter numberOfCasesStarted = Counter.build()
+            .name("kie_server_case_started_total")
+            .help("Kie Server Started Cases")
+            .labelNames("case_definition_id")
+            .register();
+
+    protected static final Gauge runningCases = Gauge.build()
+            .name("kie_server_case_running_total")
+            .help("Kie Server Running Cases")
+            .labelNames("case_definition_id")
+            .register();
+
+    protected static final Summary caseDuration = Summary.build()
+            .name("kie_server_case_duration_seconds")
+            .help("Kie Server Case Duration")
+            .labelNames("case_definition_id")
+            .register();
+
+    protected static void recordRunningCaseInstance(String caseDefinitionId) {
+        runningCases.labels(caseDefinitionId).inc();
+    }
+
+    @Override
+    public void afterCaseStarted(CaseStartEvent event) {
+        LOGGER.debug("After case started: {}", event);
+        numberOfCasesStarted.labels(event.getCaseDefinitionId()).inc();
+        recordRunningCaseInstance(event.getCaseDefinitionId());
+    }
+
+    @Override
+    public void afterCaseClosed(CaseCloseEvent event) {
+        endCase(event);
+    }
+
+    @Override
+    public void afterCaseDestroyed(CaseDestroyEvent event) {
+        endCase(event);
+    }
+
+    @Override
+    public void afterCaseCancelled(CaseCancelEvent event) {
+        endCase(event);
+    }
+
+    protected void endCase(CaseEvent event) {
+        LOGGER.debug("Close Case with Id: {}", event.getCaseId());
+        final CaseFileInstance caseFile = event.getCaseFile();
+        if (caseFile != null) {
+            runningCases.labels(caseFile.getDefinitionId()).dec();
+            final Date caseReopenDate = caseFile.getCaseReopenDate();
+            final Date start = caseReopenDate == null ? caseFile.getCaseStartDate() : caseReopenDate;
+            final double duration = millisToSeconds(System.currentTimeMillis() - start.getTime());
+            caseDuration.labels(caseFile.getDefinitionId()).observe(duration);
+            LOGGER.debug("Case duration: {}s", duration);
+        }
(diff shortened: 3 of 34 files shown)


kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusJobListener.java
```diff
@@ -0,0 +1,103 @@
...
+            .help("Kie Server Cancelled Jobs")
+            .labelNames("container_id", "command_name")
+            .register();
+    protected static final Summary jobDuration = Summary.build()
+            .name("kie_server_job_duration_seconds")
+            .help("Kie Server Job Duration")
+            .labelNames("container_id", "command_name")
+            .register();
+
+    private static ThreadLocal<Summary.Timer> threadLocalStartTime = new ThreadLocal<>();
```
vblagoje: Does this one need to be static?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
```diff
@@ -59,13 +72,62 @@ public boolean isInitialized() {
 
     @Override
     public boolean isActive() {
-        return !disabled;
+        return disabled == false;
```
vblagoje: Nitpicking this change here - isn't this antipattern?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
```diff
@@ -59,13 +72,62 @@ public boolean isInitialized() {
...
+            if (executorService != null) {
+                executorService.addAsyncJobListener(new PrometheusJobListener());
+            }
+
+            final QueryServiceImpl queryService = jBPMExtension.getAppComponents(QueryServiceImpl.class);
+            if (queryService != null) {
+                final DataSetDefRegistry dataSetDefRegistry = queryService.getDataSetDefRegistry();
+                final PrometheusDataSetListener listener = new PrometheusDataSetListener(dataSetDefRegistry);
+                listener.init();
+                dataSetDefRegistry.addListener(listener);
```
vblagoje: Nitpicking: is this bi-directional relationship really needed?

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/prometheus/PrometheusIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+        long abortedInstances = instances.stream().filter(pi -> pi.getState() == STATE_ABORTED).count();
+
+        final List<TaskSummary> tasks = taskClient.findTasks(USER_ID, 0, 1000);
+        final Supplier<Stream<TaskSummary>> taskSummaryStream = () -> tasks.stream().filter(t -> "First task".equals(t.getName()) && CONTAINER_ID.equals(t.getContainerId()) && StringUtils.equalsAny(t.getStatus(), Status.Completed.toString(), Status.Exited.toString()));
+        long totalTasks = taskSummaryStream.get().count();
+        long completedTasks = taskSummaryStream.get().filter(t -> Status.Completed.toString().equals(t.getStatus())).count();
+        long exitedTasks = taskSummaryStream.get().filter(t -> Status.Exited.toString().equals(t.getStatus())).count();
+
+        assertThat(getMetrics()).contains(
+                format("kie_server_process_instance_started_total{container_id=\"prometheus\",process_id=\"per-process-instance-project.usertask\",} %d.0", totalInstances),
```
vblagoje: Great stuff

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/prometheus/PrometheusIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+                format("kie_server_process_instance_completed_total{container_id=\"prometheus\",process_id=\"per-process-instance-project.usertask\",status=\"2\",} %d.0", completedInstances),
```
vblagoje: Excellent

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerEventListener.java
```diff
@@ -0,0 +1,106 @@
...
+            .name("kie_server_container_started_total")
+            .help("Kie Server Started Containers")
+            .labelNames("container_id")
+            .register();
+    protected static final Gauge runningContainers = Gauge.build()
+            .name("kie_server_container_running_total")
+            .help("Kie Server Running Containers")
+            .labelNames("container_id")
+            .register();
+    protected static final Gauge kieServerStartTime = Gauge.build()
```
vblagoje: This is an odd metric. Do we really need it?

MarianMacik:
jenkins execute springboot build

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
```diff
@@ -259,6 +259,7 @@
               <org.jbpm.server.ext.disabled>${org.jbpm.server.ext.disabled}</org.jbpm.server.ext.disabled>
               <org.drools.server.ext.disabled>${org.drools.server.ext.disabled}</org.drools.server.ext.disabled>
               <org.optaplanner.server.ext.disabled>true</org.optaplanner.server.ext.disabled>
+              <org.kie.prometheus.server.ext.disabled>false</org.kie.prometheus.server.ext.disabled>
```
MarianMacik: I think that "convention" here is that what extension is actually tested in a specific module is right at the top defined as false inside `<properties>` tag. And here it is only referenced via `${}`. But probably @sutaakar knows the best how this should be enabled.

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
```diff
@@ -289,6 +290,7 @@
...
```
MarianMacik: Previous comment applicable here too.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
```diff
@@ -59,13 +72,62 @@ public boolean isInitialized() {
...
+                }
+            }
+        }
+
         initialized = true;
+        LOGGER.info("{} started", toString());
+    }
+
+    protected void registerDefaultDescriptor() {
+        System.setProperty("org.kie.deployment.desc.location", "classpath:/META-INF/prometheus-deployment-descriptor-defaults.xml");
```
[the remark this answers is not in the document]
cristianonicolai: @MarianMacik correct, they would need to redefine the Prometheus ones in their custom descriptor.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java
```diff
@@ -0,0 +1,121 @@
...
+            .name("kie_server_process_instance_running_total")
+            .help("Kie Server Running Process Instances")
+            .labelNames("container_id", "process_id")
+            .register();
+    protected static final Summary processInstanceDuration = Summary.build()
+            .name("kie_server_process_instance_duration_seconds")
+            .help("Kie Server Process Instances Duration")
+            .labelNames("container_id", "process_id")
+            .register();
+    protected static final Summary workItemDuration = Summary.build()
```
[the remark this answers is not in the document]
cristianonicolai: fixed

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
```diff
@@ -59,13 +72,62 @@ public boolean isInitialized() {
...
+            }
+        }
+
+        KieServerExtension caseExtension = context.getServerExtension(CaseKieServerExtension.EXTENSION_NAME);
+        if (caseExtension != null) {
+            final CaseManagementRuntimeDataServiceBase caseRuntime = caseExtension.getAppComponents(CaseManagementRuntimeDataServiceBase.class);
+            if (caseRuntime != null) {
+                final CaseInstanceList caseInstances = caseRuntime.getCaseInstances(asList(CaseStatus.OPEN.getName()), 0, Integer.MAX_VALUE, null, false);
+                for (CaseInstance instance : caseInstances.getCaseInstances()) {
+                    recordRunningCaseInstance(instance.getCaseDefinitionId());
```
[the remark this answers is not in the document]
cristianonicolai: correct, to load the number of instances in progress, as no listener would be called for that.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
```diff
@@ -14,37 +14,50 @@
...
     public static final String EXTENSION_NAME = "Prometheus";
+    
+    private static final Logger LOGGER = LoggerFactory.getLogger(PrometheusKieServerExtension.class);
+    private static PrometheusMetrics METRICS = null;
 
-    private static final Boolean disabled = Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_PROMETHEUS_SERVER_EXT_DISABLED, "true"));
-
-    public static final CollectorRegistry registry = CollectorRegistry.defaultRegistry;
-
+    private Boolean disabled = Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_PROMETHEUS_SERVER_EXT_DISABLED, "true"));
```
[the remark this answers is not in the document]
cristianonicolai: the difference is that this one is disabled by default, so in order to enable it for Spring boot ( which uses constructor instead ) I went this way.

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
```diff
@@ -39,12 +40,7 @@
 import org.kie.server.api.KieServerConstants;
 import org.kie.server.api.model.Message;
 import org.kie.server.api.model.Severity;
-import org.kie.server.services.api.KieContainerCommandService;
-import org.kie.server.services.api.KieContainerInstance;
-import org.kie.server.services.api.KieServerApplicationComponentsService;
-import org.kie.server.services.api.KieServerExtension;
-import org.kie.server.services.api.KieServerRegistry;
-import org.kie.server.services.api.SupportedTransports;
+import org.kie.server.services.api.*;
```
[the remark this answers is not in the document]
cristianonicolai: fixed

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
cristianonicolai: happy to change it if needed, let me know.

kie-spring-boot/kie-spring-boot-autoconfiguration/kie-server-spring-boot-autoconfiguration-drools/src/main/java/org/kie/server/springboot/autoconfiguration/drools/PrometheusKieServerAutoConfiguration.java
```diff
@@ -34,10 +40,34 @@
 @EnableConfigurationProperties(KieServerProperties.class)
 public class PrometheusKieServerAutoConfiguration {
 
+    private KieServerProperties properties;
+
+    public PrometheusKieServerAutoConfiguration(KieServerProperties properties) {
+        this.properties = properties;
+    }
+
```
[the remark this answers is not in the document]
cristianonicolai: removed

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerEventListener.java
cristianonicolai: yeah, I research a few examples on how to do this, and will be adding a single metric value to track whenever the server came up. I have one graph in grafana that can then show the uptime for the server.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
cristianonicolai: its because I'm using the _PrometheusDataSetListener_ as a series of different listeners related to data set API: _DataSetDefRegistryListener_, _DataSetPreprocessor_,                                            _DataSetPostProcessor_ so not only it is a listener for the registry but also for every time a dataset is added or modified, we need to register the pre and post processor again.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
cristianonicolai: fixed

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusJobListener.java
cristianonicolai: not necessarily.

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
sutaakar: I think it would be good to keep the same style everywhere. @cristianonicolai Please adjust this.

cristianonicolai: done

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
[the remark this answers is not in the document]
MarianMacik: but SpringBoot extension has `isActive()` method overridden anyway to return `true`. So maybe this is no longer needed?

MarianMacik:
@cristianonicolai
Also please add `<kieserver.prometheus.enabled>true</kieserver.prometheus.enabled>` here:
https://github.com/kiegroup/droolsjbpm-integration/blob/83e57e9ebadc2ab6929977bc3d496411ccf503cf/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml#L419
So we can run `-all` module of KIE Server tests also with Spring Boot.

Moreover, I think this is not needed anymore:
https://github.com/kiegroup/droolsjbpm-integration/blob/83e57e9ebadc2ab6929977bc3d496411ccf503cf/kie-server-parent/kie-server-tests/pom.xml#L1241-L1242
Since we now have a custom `SpringBootPrometheusExtension` which is always active. Can we also rename that extension to `SpringBootPrometheusKieServerExtension` so it is aligned with `SpringBootJBPMKieServerExtension` and `SpringBootCaseKieServerExtension`?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusCaseEventListener.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+package org.kie.server.services.prometheus;
+
+import java.util.Date;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Gauge;
+import io.prometheus.client.Summary;
+import org.jbpm.casemgmt.api.event.*;
```
MarianMacik: One more star import.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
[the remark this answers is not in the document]
MarianMacik: Ignore my comment, I see it is Outdated.

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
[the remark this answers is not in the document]
MarianMacik: There's one more in `PrometheusCaseEventListener`, namely `import org.jbpm.casemgmt.api.event.*;`

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java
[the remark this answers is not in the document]
MarianMacik: There's one more, `runningProcessInstance` -> `runningProcessInstances`.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerEventListener.java
vblagoje: Aha I see, cool! So we'll have one unified metric for this regardless of which KIE service is getting monitored, right?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
vblagoje: +1

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusJobListener.java
vblagoje: If it does not have to be let's make it non-static. Thoughts @MarianMacik ?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerEventListener.java
cristianonicolai: yep

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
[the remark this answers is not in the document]
cristianonicolai: fixed

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java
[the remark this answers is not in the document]
cristianonicolai: fixed

kie-server-parent/kie-server-tests/kie-server-integ-tests-all/pom.xml
cristianonicolai: fixed

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusJobListener.java
cristianonicolai: removed static reference.

cristianonicolai:
@MarianMacik thanks for the feedback, I believe I've addressed all comments but please take another look.

lucamolteni:
Sorry the review wasn't submitted

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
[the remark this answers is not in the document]
MarianMacik: Well, the thing is that with this approach even if they redefine it with a custom descriptor, their descriptor won't be taken into account since it will be overridden by this property setter.

I discussed this with Karel and instead of setting this via the property, we can use a similar approach as for CaseMgmt. There is this piece of code:
https://github.com/kiegroup/droolsjbpm-integration/blob/83e57e9ebadc2ab6929977bc3d496411ccf503cf/kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java#L158-L162

which registers `CaseConfigurationDeploymentListener` only when case mgmt extension is enabled. That listener then reads deployment descriptor and adds CaseEventListeners to CaseEventSupport as you can see here:

https://github.com/kiegroup/jbpm/blob/ec824f1bf1889da57cb4d6368ad6faa73fefc21f/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/event/CaseConfigurationDeploymentListener.java#L69

So we can do something similar. We can just create `PrometheusConfigurationDeploymentListener` which will just do something like:

```
public void onDeploy(DeploymentEvent event) {
        InternalRuntimeManager runtimeManager = (InternalRuntimeManager) event.getDeployedUnit().getRuntimeManager();
        RuntimeEngine runtimeEngine =  runtimeManager.getRuntimeEngine(EmptyContext.get());
        runtimeEngine.getKieSession().addEventListener(new PrometheusProcessEventListener());
        ((EventService<TaskLifeCycleEventListener>)runtimeEngine.getTaskService()).registerTaskEventListener(new PrometheusTaskEventListener());
        if (runtimeManager instanceof PerCaseRuntimeManager) {
            ((PerCaseRuntimeManager) runtimeManager).getCaseEventSupport().addEventListener(new PrometheusCaseEventListener());
        }
```

@mswiderski wdyt? Will it work? I think this is less intrusive (although it is more code) than to setting global deployment descriptor via system property.

I especially don't know if `runtimeManager.getRuntimeEngine(EmptyContext.get());` is enough in case there will be PPI or PerRequest runtime manager.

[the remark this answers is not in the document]
MarianMacik: OK, I had a discussion with @mswiderski and we should use the deployment descriptor way, but we should first check if a user has already defined one with a system property, so this extension doesn't override it.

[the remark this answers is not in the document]
cristianonicolai: ok, I tried the approach you mentioned, it works fine for case management, but I couldn't find a way to register the process and task listeners.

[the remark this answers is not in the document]
cristianonicolai: will add a warning if system property already defined.

kie-server-parent/kie-server-services/kie-server-services-case-mgmt/src/main/java/org/kie/server/services/casemgmt/CaseKieServerExtension.java
```diff
@@ -254,7 +248,9 @@ public void disposeContainer(String id, KieContainerInstance kieContainerInstanc
         if (serviceType.isAssignableFrom(kieContainerCommandService.getClass())) {
             return (T) kieContainerCommandService;
         }
-        return null;
+
+        final Optional<Object> service = services.stream().filter(s -> serviceType.isAssignableFrom(s.getClass())).findFirst();
+        return service.isPresent() ? (T) service.get() : null;
```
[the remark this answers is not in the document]
cristianonicolai: done

kie-server-parent/kie-server-services/kie-server-services-prometheus/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+``` 
+curl -u wbadmin:wbadmin http://localhost:8080/kie-server/services/rest/metrics  
+``` 
+
+# Available metrics for Prometheus
+
+## Kie Server metrics
+
+kie_server_start_time: Kie Server Start Time (Gauge)
```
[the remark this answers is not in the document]
cristianonicolai: correct, for measuring time that is the case, this one I've mixed feeling as they recommend seconds as the unit but here it represents a timestamp. If we want to keep it standard, we should use seconds in here then. wdyt?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusMetrics.java
```diff
@@ -42,6 +42,10 @@ private static long toMicro(long second) {
         return IntStream.range(start, end).mapToDouble(l -> toMicro((long) l)).toArray();
     }
 
+    protected static double millisToSeconds(long millis) {
+        return millis / 1000.0;
+    }
+
```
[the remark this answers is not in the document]
cristianonicolai: I believe so, but from what I researched Prometheus is quite opinionated about using seconds for time-related data.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusJobListener.java
```diff
@@ -0,0 +1,107 @@
...
+            .labelNames("container_id", "command_name")
+            .register();
+
+    protected static final Summary jobDuration = Summary.build()
+            .name("kie_server_job_duration_seconds")
+            .help("Kie Server Job Duration")
+            .labelNames("container_id", "command_name")
+            .register();
+
+    private ThreadLocal<Summary.Timer> threadLocalStartTime = new ThreadLocal<>();
```
cristianonicolai: @lucamolteni regarding threadlocal, this is a listener that is called before and after the job execution, I would say the we can use both approaches, a timestamp in the event or this one. For this one in particular is just that we dont have the time unit for the event, so less intrusive way to collect the data.

lucamolteni: Less intrusive but more fragile, ThreadLocal may give problems when used in ThreadPools

https://techblog.bozho.net/threadlocal-variables-and-thread-pools-it-can-go-wrong/

MarianMacik: I have also noticed that before it was static, now it isn't. Is there any reason for that?

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusKieServerExtension.java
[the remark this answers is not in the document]
MarianMacik: Thanks, looks good now!

MarianMacik:
jenkins execute springboot build

cristianonicolai:
jenkins execute springboot build

cristianonicolai:
@MarianMacik @lucamolteni thanks for the feedback, I managed to completely remove the usage of thread local.

kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    @Override
+    public void beforeNodeLeft(ProcessNodeLeftEvent event) {
+        LOGGER.debug("Before Node left event: {}", event);
+        final NodeInstance nodeInstance = event.getNodeInstance();
+        if (nodeInstance instanceof WorkItemNodeInstance) {
+            WorkItemNodeInstance wi = (WorkItemNodeInstance) nodeInstance;
+            if (wi.getTriggerTime() != null) {
```
MarianMacik: Probably the last thing I noticed. Can this even be null? According to this I don't think so:
https://github.com/kiegroup/jbpm/pull/1473/files#diff-0d58c1b1aa778882879a180d7b1f90bdR386
(if you think it is safer to check it first, ignore this comment)

MarianMacik:
Springboot failed with 3 tests. I think you just need to customize them to Springboot as well.
```
Expecting
...
# HELP kie_server_start_time Kie Server Start Time
# TYPE kie_server_start_time gauge
kie_server_start_time{name="KieServer-SpringBoot",server_id="SpringBoot",location="http://localhost:34232/kie-server-services/services/rest/server",version="7.21.0-SNAPSHOT",} 1.555054849681E12
...
to contain:
 <["kie_server_start_time{name="KieServer@/kie-server-services"",
    "kie_server_deployments_active_total{deployment_id="prometheus",} 1.0",
    "kie_server_container_started_total{container_id="prometheus",} 1.0",
    "kie_server_container_running_total{container_id="prometheus",} 1.0"]>
but could not find:
 <["kie_server_start_time{name="KieServer@/kie-server-services""]>
 
	at org.kie.server.integrationtests.prometheus.PrometheusIntegrationTest.testKieServerStartAndContainersMetrics(PrometheusIntegrationTest.java:91)
```

cristianonicolai:
jenkins execute springboot build

cristianonicolai:
@MarianMacik should be ok now, all tests pass.

sutaakar:
@MarianMacik @lucamolteni Is this PR ok to be merged?

lucamolteni:
yes @sutaakar

PR #1800, opened 2019-04-05
[JBPM-8400] Logs are constantly flooded with 'Unable to load key store. Using password from configuration'

elguardian:
sinking message to debug level

### DIFF
--- a/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
+++ b/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
@@ -49,7 +49,7 @@ public class KeyStoreHelperUtil {
 
             return keyStoreHelper.getPasswordKey(pwdKeyAlias, pwdKeyPassword);
         } catch (RuntimeException re) {
-            logger.warn("Unable to load key store. Using password from configuration");
+            logger.debug("Unable to load key store. Using password from configuration");
             return defaultPassword;
         }
     }


MarianMacik:
@elguardian I have noticed this recently with jbpm-server-distribution 7.19 version. I think this wasn't present in 7.17. What is exactly the issue here?

sutaakar:
Hmm, that may be related to https://github.com/kiegroup/droolsjbpm-integration/pull/1712 by introducing healthcheck

mswiderski:
I don’t think this is the right solution. As the whole keystore was to increase security and the warnings should be annoying if not changed. So this jira is more to make sure we use keystore ourselves too. So I’d say we should use keystore in server zip distribution and maybe installer as well

MarianMacik:
I agree with @mswiderski. In case you show me how to configure it, I might give it a try :)

mswiderski:
@MarianMacik how about docs? https://docs.jboss.org/jbpm/release/7.20.0.Final/jbpm-docs/html_single/#_securing_password_using_key_store

elguardian:
@mswiderski I was investigating yesterday about the keytool maven plugin to generate a standard keystore everytime the jbpm server zip is created. I saw the standalone.xml there to set the properties.

MarianMacik:
@elguardian by quickly looking at it, it doesn't seem to have support of `-importpassword` option of keytool command. Is it true or there is an option to actually create a new keystore? We can also just pre-create a keystore and store it there. We will hardly ever be changing default passwords. Wdyt @mswiderski ?

mswiderski:
my idea was to create the keystore and just add it to server distribution zip so it's used. It should be the same username and password as we use now so I don't think we will be changing that at all

MarianMacik:
ok, I will look at it

MarianMacik:
Should be fixed by https://github.com/kiegroup/kie-wb-distributions/pull/911

PR #1801, opened 2019-04-05
Reduce the Kie server controller ping timeout

sutaakar:
To make KieControllerCrashIntegrationTest more reliable on slower machines.

@elguardian  Can you please take a look?


elguardian:
@sutaakar lgtm. There is no way to fully control the race condition, but somehow that mitigates the problem.

PR #1810, opened 2019-04-16
[JBPM-8177] Security exceptions in localhost log when running kie-server on JWS

elguardian:
bump version to web 3.1 and added the proper constraints to avoid SEVERE log

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/ee7-resources/WEB-INF/web.xml
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-standalone/src/main/ee7-resources/WEB-INF/web.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         version="3.0"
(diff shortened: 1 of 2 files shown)


elguardian:
@MarianMacik do you mind to check the controller in tomcat ?

elguardian:
@MarianMacik that is normal and is by spec IIRC

MarianMacik:
@elguardian Yes, I have nothing against it, I put it here just as a "proof" :)

elguardian:
@mswiderski can you merge if you agree plz ?

MarianMacik:
@mswiderski My only question was about the allowance of OPTION method for the same purpose as in KIE Server. If it is not a big value in controller, we can merge it as it is now.

mswiderski:
@MarianMacik I'd say it would make sense to add the same OPTIONS as people could build JS UI on top of the controller so might be good to align...

MarianMacik:
@elguardian can you add the OPTION method, please? It is the same snippet as here:
https://github.com/kiegroup/droolsjbpm-integration/pull/1810#pullrequestreview-227262625
Just with different URL patterns:
```
<url-pattern>/rest/*</url-pattern>
<url-pattern>/websocket/*</url-pattern>
```

elguardian:
done @MarianMacik let me know if there is anything else, otherwise let's try if someone can merge

MarianMacik:
Tested OPTIONS method as well.

MarianMacik:
@mswiderski please approve so I can merge

PR #1817, opened 2019-05-09
JBPM-8469 - Set default PU name

cristianonicolai:


cristianonicolai:
@mswiderski without setting the PU name in here, it tries to create one with a null name, if not defined in the project level settings.

cristianonicolai:

PR #1819, opened 2019-05-10
[DROOLS-3875] Shade drools-model file based on the GAV of the KJar 

lucamolteni:
Shade drools-model file based on the GAV of the KJar in order to support multiple modules in a KieClassPathContainer

No test attached, but there's a repdroducer attached in  
https://issues.jboss.org/browse/DROOLS-3875


lucamolteni:
@mariofusco @tarilabs @evacchi Please review: this should be also ported to submarine

PR #1824, opened 2019-05-20
[JBPM-8506] Need slf4j adapter for kie server with RHDM 7

elguardian:
added org.slf4j:slf4j-jdk14 to the assembly so weblogic can log


elguardian:
@MarianMacik @mswiderski can you review this ?

sutaakar:
@elguardian Just a quick question, have you tried to check that the change works correctly on WebSphere? I am asking because the ee7 artifact is used in both application servers.

sutaakar:
WebLogic and WebSphere seem to work correctly, just reported DROOLS-4054 .

elguardian:
@mswiderski can you check this one ?

This is a customer case, if goes in plz merge.

PR #1825, opened 2019-05-21
[RHDM-851] Xerces is needed to run camel tests on IBM JDK

lucamolteni:
According to this

https://stackoverflow.com/a/44442682/4206

XercesImpl is needed


lucamolteni:
@jiripetrlik please try this on the IBM JDK CI I've tried on my machine and it works

lucamolteni:
@mariofusco

PR #1829, opened 2019-05-23
[JBPM-8495] Fix integration tests

ruromero:
Fix https://issues.jboss.org/browse/JBPM-8495

Avoid using Arquillian and directly calls the PIM Backend through Apache HTTPClient.
The service is started using antrun plugin


kie-ci6:

process-migration-service/pom.xml
```diff
@@ -399,6 +395,46 @@
...
+                    </arguments>
+                    <async>true</async>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <configuration>
+                    <target>
+                        <sleep seconds="60" />
```
MarianMacik: This is not very useful and quite error-prone. Maybe we can use maven-process-plugin we already use for Spring Boot tests? It has ability to wait for URL to be reachable - it is configured with healthcheckUrl property in configuration.

Here is its configuration:

https://github.com/kiegroup/droolsjbpm-integration/blob/33ab4b666e791d7ea2d0eb901deb69231a81e990/kie-server-parent/kie-server-tests/pom.xml#L1215

Moreover, please create a wildfly profile and put everything related to integration tests under it - process-exec-maven-plugin, cargo configuration etc.

So if you leave executions of cargo and process-exec-maven-plugin under that profile, they won't be run by default, which is what we want. The reason is that PRs run with -Pwildfly enabled, but in case we want to run the integration tests from community with productized builds, we have to turn off automation in community (by not activating wildfly profile).

Thanks.

(Since this is not related to implementation, we still have time to merge this I think.)

process-migration-service/pom.xml
```diff
@@ -514,6 +359,185 @@
         </plugins>
     </build>
     <profiles>
+        <profile>
+            <id>wildfly</id>
```
MarianMacik: Sorry to change this once again, but I realized that for example the failsafe configuration should be outside this profile since we want to run integration tests but with internal cargo configuration. Right now we either run integration tests with community wildfly + process-exec-plugin or we don't run them at all.

I think the best thing would be to stick with the configuration from kie-server-tests-parent here:
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/pom.xml

So I think the best approach would be to define cargo-general config outside of profile. Then in wildfly profile there should be just wildfly-specific configuration. We can even make process-exec-maven-plugin outside of that profile.

Then as with kie-server we can put 2 profiles here. One which will turn off the container (cargo plugin) + microservice (process-exec-maven-plugin), active by default:

https://github.com/kiegroup/droolsjbpm-integration/blob/a852ff6ff6b94f5763045286af4dcbb0009f08ea/kie-server-parent/kie-server-tests/pom.xml#L1262-L1294

And the second profile to turn off integration tests by default:
https://github.com/kiegroup/droolsjbpm-integration/blob/a852ff6ff6b94f5763045286af4dcbb0009f08ea/kie-server-parent/kie-server-tests/pom.xml#L1331-L1350

So when we run tests internally, it will be possible to just do `-Pcustom-container` and just cargo + microservice setup will be turned off, but failsafe will be active. Moreover, when we do just `clean install`, no tests will be run. And when we do `-Pwildfly`, community KIE Server + microservice will be run and tested.

process-migration-service/pom.xml
```diff
@@ -48,6 +48,9 @@
         <kie.server.base.http.url>
             http://${container.hostname}:${container.port}/${kie.server.context}/services/rest/server
         </kie.server.base.http.url>
+        <pim.server.http.url>http://localhost:8180</pim.server.http.url>
+        <pim.server.http.rest>${pim.server.http.url}/rest</pim.server.http.rest>
+        <pim.server.http.health>${pim.server.http.rest}/health/readiness</pim.server.http.health>
```
MarianMacik: There are some properties which are not used I think:
```
<kie.server.context.factory>org.wildfly.naming.client.WildFlyInitialContextFactory</kie.server.context.factory>
<kie.server.controller.context>kie-server-controller-services</kie.server.controller.context>
<org.kie.server.controller.templatefile>${project.build.directory}/server-template-storage.xml</org.kie.server.controller.templatefile>
<webservice.http.url>http://${container.hostname}:${container.port}/kie-server-test-web-service/AcmeDemoInterface?wsdl</webservice.http.url>
```

ruromero:
All changes should be done and I also rebased the branch with the latest changes from master

ruromero:
@MarianMacik I have removed a commit belonging to #1842 added by mistake

MarianMacik:
@sutaakar Can you please briefly check the pom.xml - mainly Cargo configuration. Thanks!

MarianMacik:
ok to test

process-migration-service/pom.xml
```diff
@@ -2,549 +2,630 @@
...
-            <artifactId>kie-server</artifactId>
-            <classifier>${kie.server.classifier}</classifier>
-            <type>war</type>
-            <scope>test</scope>
-        </dependency>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.thorntail</groupId>
+        <artifactId>bom-all</artifactId>
```
mareknovotny: Why we need here to import `io.thorntail:bom-all` instead of just `io.thorntail:bom` ?

MarianMacik: To be honest, I don't know. Maybe it would work also without it.

PR #1832, opened 2019-05-27
[JBPM-8491] Do not allow filter by container alias

ruromero:
Fix https://issues.jboss.org/browse/JBPM-8491

Propagate 404 when necessary instead of throwing 500 error.
The KieServer API filters by container alias and container id. In order to prevent inconsistencies when migrating, an additional validation has been added to enforce containerID matches the request.

Besides, in the frontend side the errors are captured to prevent user from moving forward with invalid data

### DIFF
--- a/process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
+++ b/process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
@@ -21,7 +21,9 @@ import org.kie.processmigration.model.ProcessInfo;
 import org.kie.processmigration.model.ProcessInfos;
 import org.kie.processmigration.model.RunningInstance;
 import org.kie.processmigration.model.exceptions.InvalidKieServerException;
+import org.kie.processmigration.model.exceptions.ProcessDefinitionNotFoundException;
 import org.kie.processmigration.service.KieService;
+import org.kie.server.api.exception.KieServicesHttpException;
 import org.kie.server.api.marshalling.MarshallingFormat;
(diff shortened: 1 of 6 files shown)


kie-ci10:

process-migration-service/src/main/java/org/kie/processmigration/service/impl/KieServiceImpl.java
```diff
@@ -139,17 +143,29 @@ private ProcessServicesClient getProcessServicesClient(String kieServerId) throw
...
+        }
 
         //Add this replacement here because in react-svgmt, ? and = are not allowed.
         svgFile = svgFile.replaceAll("\\?shapeType=BACKGROUND", "_shapeType_BACKGROUND");
         processInfo.setSvgFile(svgFile);
 
         ProcessDefinition pd = getProcessServicesClient(kieServerId).getProcessDefinition(containerId, processId);
+        if(!pd.getContainerId().equals(containerId)) {
+            throw new ProcessDefinitionNotFoundException(kieServerId, containerId, processId);
+        }
```
MarianMacik: Can't we just list containers available in KIE server and check if the container id from the input is among them?

ruromero: This is already reported as an enhancement https://issues.jboss.org/browse/JBPM-8473

MarianMacik: ok, thanks

MarianMacik:
ok to test

PR #1833, opened 2019-05-27
Test for complete error stacktrace of the error returned by kie server

gmunozfe:
@MarianMacik please review it.

New module for testing the feature _"JBPM-8399 - Add new configuration flag that will show a complete error stacktrace.."_ setting the system property at cargo plugin:

`<org.kie.server.stacktrace.included>true</org.kie.server.stacktrace.included>`

### DIFF
--- a/kie-server-parent/kie-server-wars/kie-server/src/main/assembly/assembly-ee7-container.xml
+++ b/kie-server-parent/kie-server-wars/kie-server/src/main/assembly/assembly-ee7-container.xml
@@ -68,6 +68,8 @@
         <include>org.kie.server:kie-server-controller-websocket-common</include>
 
         <include>org.jbpm:jbpm-services-ejb-timer</include>
+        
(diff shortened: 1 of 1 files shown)


MarianMacik:
Hi, we have discussed this with @sutaakar and probably the best way how to test this would be to reuse current tests in the `-all` test module. However, this is not possible right now as this flag will not include stack trace, rather it will overwrite the default message with the stack trace itself. Since the name of the property is `org.kie.server.stacktrace.included`, you can improve its behaviour here:

https://github.com/kiegroup/droolsjbpm-integration/blob/715ceefc40ed5784ea77273bd07a88fe52e9a427/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java#L224-L236

So it will print the default message and then on a new line stackTrace. By doing this it will be more accurate (include stack trace) + we will be able to reuse our current tests. Wdyt?

PR #1837, opened 2019-05-28
[JBPM-8399] Added test for complete error stacktrace returned by kie-server

gmunozfe:
@MarianMacik added test into kie-server-integ-tests-all and change _errorMessage_ method to print always default message (and then, complete stack trace if property is enabled)

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
@@ -222,17 +222,17 @@ public class RestUtils {
     }
     
     public static String errorMessage(Throwable e, String defaultMessage) {
+        StringWriter sw = new StringWriter();
+        PrintWriter pw = new PrintWriter(sw);
+
+        pw.println(defaultMessage);
+
         if (INCLUDE_STACKTRACE) {
-            StringWriter sw = new StringWriter();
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/jbpm/rest/JbpmRestCustomIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Test;
+import org.kie.server.api.marshalling.MarshallingFormat;
+import org.kie.server.api.model.ReleaseId;
+import org.kie.server.integrationtests.config.TestConfig;
+import org.kie.server.integrationtests.shared.KieServerDeployer;
+import org.kie.server.integrationtests.shared.KieServerUtil;
+import org.kie.server.integrationtests.shared.basetests.RestOnlyBaseIntegrationTest;
+
+
+public class JbpmRestCustomIntegrationTest extends RestOnlyBaseIntegrationTest {
```
MarianMacik: Please change the name of the class to something more descriptive, e.g. `StackTraceIncludedIntegrationTest` or something which says that we test stack trace. Also this class doesn't have to test only jbpm endpoints, so jbpm reference is not needed.

gmunozfe: Thanks for the feedback! Class name changed

PR #1843, opened 2019-06-07
JBPM-8399: Fix EAP test for stack trace included

gmunozfe:
@sutaakar please review


sutaakar:
I would say there is no need to test "at ", otherwise looks ok

gmunozfe:
Ok, removed "at ", thanks!

sutaakar:
@mswiderski Can you please take a look? It is a simple test adjustments addressing stacktrace changes in EAP for Java11.

PR #1842, opened 2019-06-07
[JBPM-8501] Allow reloading KIE Server clients

ruromero:
Fix https://issues.jboss.org/browse/JBPM-8501

UI changes:
* Allow refresh the KieServer lists and disable those whose state is not "SUCCESS"
* Run migrations radio button is by default "Now"
* Gradually moving from axios to fetch and restful calls
* Notify the users when the process definitions cannot be retrieved

Backend changes:
* Async retry to create the KieServer clients and reload status each time the kieServers status is requested
* Pending reformat in some files
* ProcessDefinitions fetching are now per process

Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

PR #1844, opened 2019-06-07
[JBPM-8492] Allow any task to be mapped

ruromero:
Fix https://issues.jboss.org/browse/JBPM-8492
Merge after https://github.com/kiegroup/droolsjbpm-integration/pull/1842

* Allow users to map any type of task
  * They must be of the same type.
  * This validation only happens in the UI.
* Simplified process definition json.
* Dropdown tasks only show `taskName (taskType)` instead of `taskName: taskID`


ruromero:
@jstastny-cz it should be fixed now

ruromero:
Closing for #1869

PR #1861, opened 2019-07-09
added missing versions

mbiarnes:


mbiarnes:
> Should theses versions not come from a `<dependencyManagement>` declaration in a parent?
AFA I can see there is no springframework in a <dependencyManagament> in a parent.
Si the version is coming from kie-parent as 4.3.3.RELEASE
There is an override of version.org.springframework https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-spring/pom.xml#L23 - but if you want to use this version you have to specify this or at least create a dependecyManagement section.

manstis:
@mbiarnes Thank-you for the explanation... I'd like to think Maven would complain if you try to use a version not defined in a `dependencyManagement` section (isn't that the whole point of the dependency management?).. So, IMO, overriding the version in a _local_ (to this `pom`) `dependencyManagement` section would be more _consistent_ with our _existing_ configuration..

mbiarnes:
@manstis OK - let me do this

mbiarnes:
> @mbiarnes Thank-you for the explanation... I'd like to think Maven would complain if you try to use a version not defined in a `dependencyManagement` section (isn't that the whole point of the dependency management?).. So, IMO, overriding the version in a _local_ (to this `pom`) `dependencyManagement` section would be more _consistent_ with our _existing_ configuration..

@manstis DONE - better now? Tested with a mvn dependency:tree and all artifacts are using the 5.1.3.RELEASE version

mariofusco:
The necessity of this PR is only caused by the fact that the maven properties override mechanism is not working anymore because all projects should extend kie-platform-bom instead of kie-parent as -integration is doing now. We will send another PR to fix this accordingly.

PR #1862, opened 2019-07-10
[PLANNER-1553] Able to configure thread pool queue size in execution …

tkobayas:
…server

It doesn't contain a unit test because I'm not sure about the way/place for such a test (But I tested manually on my RHPAM 7.3.1 and confirmed it works expectedly). If there is a suitable way/place, please kindly guide me. Thanks!

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
@@ -112,6 +112,8 @@ public class KieServerConstants {
     public static final String KIE_SERVER_ROUTER = "org.kie.server.router";
     public static final String KIE_SERVER_ROUTER_ATTEMPT_INTERVAL = "org.kie.server.router.connect";
 
+    public static final String KIE_OPTAPLANNER_THREAD_POOL_QUEUE_SIZE = "org.optaplanner.server.ext.thread.pool.queue.size";
+
     // ProcessConfig configuration item constants
     public static final String PCFG_RUNTIME_STRATEGY = "RuntimeStrategy";
     public static final String PCFG_KIE_BASE = "KBase";
--- a/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
@@ -88,12 +88,16 @@ public class OptaplannerKieServerExtension
         // If new jobs are submitted and all threads are busy, the default reject policy will kick in.
         int availableProcessorCount = Runtime.getRuntime().availableProcessors();
         int resolvedActiveThreadCount = Math.max(1, availableProcessorCount - 2);
+        int queueSize = Integer.parseInt(System.getProperty(
+                KieServerConstants.KIE_OPTAPLANNER_THREAD_POOL_QUEUE_SIZE, String.valueOf(resolvedActiveThreadCount)));
+        logger.info("Creating a ThreadPoolExecutor with corePoolSize = " + resolvedActiveThreadCount + ","
+                + " maximumPoolSize = " + resolvedActiveThreadCount + ", queueSize = " + queueSize);
         this.threadPool = new ThreadPoolExecutor(
                 resolvedActiveThreadCount,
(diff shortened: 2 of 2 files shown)


kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
```diff
@@ -88,12 +88,16 @@ public void init(KieServerImpl kieServer, KieServerRegistry registry) {
         // If new jobs are submitted and all threads are busy, the default reject policy will kick in.
         int availableProcessorCount = Runtime.getRuntime().availableProcessors();
         int resolvedActiveThreadCount = Math.max(1, availableProcessorCount - 2);
+        int queueSize = Integer.parseInt(System.getProperty(
+                KieServerConstants.KIE_OPTAPLANNER_THREAD_POOL_QUEUE_SIZE, String.valueOf(resolvedActiveThreadCount)));
+        logger.info("Creating a ThreadPoolExecutor with corePoolSize = " + resolvedActiveThreadCount + ","
```
sutaakar: Is this information important enough to deserve info log level? Wouldn't be enough to use debug level?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
```diff
@@ -112,6 +112,8 @@
     public static final String KIE_SERVER_ROUTER = "org.kie.server.router";
     public static final String KIE_SERVER_ROUTER_ATTEMPT_INTERVAL = "org.kie.server.router.connect";
 
+    public static final String KIE_OPTAPLANNER_THREAD_POOL_QUEUE_SIZE = "org.optaplanner.server.ext.thread.pool.queue.size";
```
ge0ffrey: Why the ".ext" in the name? Which other properties do that convention too?

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
tkobayas: I'm under impression that component's configuration values deserve INFO level. And also helpful from support point of view.

For example)
~~~
2019-07-10 16:26:36,400 INFO  [org.jbpm.executor.impl.ExecutorImpl] (EJB default - 1) Starting jBPM Executor Component ...
         - Thread Pool Size: 1
         - Retries per Request: 3
         - Load from storage interval: 0 SECONDS (if less or equal 0 only initial sync with storage)
~~~

But if you push again for DEBUG level, it's fine for me to change. Please let me know your thought.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
tkobayas: This property is used only for OptaplannerKieServerExtension and it's identified by "org.optaplanner.server.ext.XXX".

https://github.com/kiegroup/droolsjbpm-integration/blob/7.23.x/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java#L31

However, if you suggest any property name, I will follow :)

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
sutaakar: Sounds reasonable, lets keep there INFO.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
ge0ffrey: Agreed, this name makes sense. Thanks!

kie-server-parent/kie-server-services/kie-server-services-optaplanner/src/main/java/org/kie/server/services/optaplanner/OptaplannerKieServerExtension.java
ge0ffrey: @sutaakar ok, I 'll presume you're ok with merging this now?

PR #1860, opened 2019-07-08
DROOLS-4278 Applying PMML model using kie-server fails

lanceleverich:
Updated the ApplyScorecardModelIntegrationTest to include a test where
the command is included in a BatchExecutionCommand.
Added ApplyScorecardMultiModuleIntegrationTest to test that the
ApplyPmmlModelCommand works when there are multiple KieBase entries in a
KieContainer.

PR #1866, opened 2019-07-15
Fix for DROOLS-4278 integration test

sutaakar:
@lanceleverich Can you please take a look? This is the fix for https://github.com/kiegroup/droolsjbpm-integration/pull/1860 .
However it is not fully working yet. The PMML call fails on "Cannot find a default KieSession". The problem is that Kie server expect container to have stateful Kie session by default unless specified otherwise. I am not sure about the PMML backgroud, should stateful or stateless session be used?

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/pmml/ApplyScorecardMultiModuleIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/pmml/ApplyScorecardMultiModuleIntegrationTest.java
@@ -32,6 +32,7 @@ public class ApplyScorecardMultiModuleIntegrationTest extends PMMLApplyModelBase
     private static ReleaseId releaseId = new ReleaseId("org.kie.server.testing", "scorecard_mm", "1.0.0.Final");
 
     private static final String CONTAINER_ID = "scorecard_mm";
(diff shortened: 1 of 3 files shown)


lanceleverich:
The session should be able to be stateful or stateless.

sutaakar:
@lanceleverich Please review now.

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/src/test/java/org/kie/server/integrationtests/drools/pmml/ApplyScorecardMultiModuleIntegrationTest.java
```diff
@@ -39,7 +39,7 @@
     @BeforeClass
     public static void buildAndDeployArtifacts() {
         KieServerDeployer.buildAndDeployCommonMavenParent();
-        KieServerDeployer.buildAndDeployMavenProjectFromResource("/kjars-sources/scorecard");
```
lanceleverich: I know that it is nitpicking and that I originally coded it this way, but we might want to make the path for the resource into a `private static final String`

lanceleverich: Just realized that I put the above comment on the line before the one I intended request for change.

sutaakar: Adjusted

PR #1870, opened 2019-07-18
JBPM-8655: Task operations are slower after a join was introduced in …

nmirasch:
…"getTaskInstanceById" query

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
@@ -110,6 +110,8 @@ public interface UserTaskServicesClient {
 
     TaskInstance findTaskById(Long taskId);
 
+    TaskInstance findTaskById(Long taskId, boolean withSLA);
+
     List<TaskSummary> findTasksAssignedAsBusinessAdministrator(String userId, Integer page, Integer pageSize);
 
     List<TaskSummary> findTasksAssignedAsBusinessAdministrator(String userId, List<String> status, Integer page, Integer pageSize);
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
@@ -828,17 +828,21 @@ public class UserTaskServicesClientImpl extends AbstractKieServicesClientImpl im
     }
 
     @Override
-    public TaskInstance findTaskById(Long taskId) {
+    public TaskInstance findTaskById(Long taskId){
+        return findTaskById(taskId, false);
+    }
+
+    @Override
+    public TaskInstance findTaskById(Long taskId, boolean withSLA) {
         if( config.isRest() ) {
             Map<String, Object> valuesMap = new HashMap<String, Object>();
             valuesMap.put(TASK_INSTANCE_ID, taskId);
 
             return makeHttpGetRequestAndCreateCustomResponse(
-                    build(loadBalancer.getUrl(), QUERY_URI + "/" + TASK_GET_URI, valuesMap), TaskInstance.class);
-
+                    build(loadBalancer.getUrl(), QUERY_URI + "/" + TASK_GET_URI + "?withSLA=" + withSLA, valuesMap), TaskInstance.class);
 
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new DescriptorCommand( "QueryService", "getTaskById", new Object[]{taskId}) ) );
+            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new DescriptorCommand( "QueryService", "getTaskById", new Object[]{taskId, withSLA}) ) );
             ServiceResponse<TaskInstance> response = (ServiceResponse<TaskInstance>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM" ).getResponses().get(0);
 
             throwExceptionOnFailure(response);
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/RuntimeDataResource.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/RuntimeDataResource.java
@@ -528,16 +528,15 @@ public class RuntimeDataResource {
     @Path(TASK_GET_URI)
     @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML})
     public Response getTaskById(@Context HttpHeaders headers, 
-            @ApiParam(value = "task id to load task instance", required = true) @PathParam(TASK_INSTANCE_ID) Long taskId) {
+            @ApiParam(value = "task id to load task instance", required = true) @PathParam(TASK_INSTANCE_ID) Long taskId,
+            @ApiParam(value = "optional include SLA data - defaults to false", required = false) @QueryParam("withSLA") @DefaultValue("false") boolean withSLA) {
         Variant v = getVariant(headers);
         // no container id available so only used to transfer conversation id if given by client
         Header conversationIdHeader = buildConversationIdHeader("", context, headers);
 
-        TaskInstance userTaskDesc = null;
         try {
-            userTaskDesc = runtimeDataServiceBase.getTaskById(taskId);
+            TaskInstance userTaskDesc = runtimeDataServiceBase.getTaskById(taskId, withSLA);
             return createCorrectVariant(userTaskDesc, headers, Response.Status.OK, conversationIdHeader);
-
         } catch (TaskNotFoundException e) {
             return notFound(MessageFormat.format(TASK_INSTANCE_NOT_FOUND, taskId), v, conversationIdHeader);
         } catch (Exception e) {
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
@@ -1313,7 +1313,6 @@ public class RuntimeDataServiceIntegrationTest extends JbpmKieServerBaseIntegrat
         taskClient.findTaskById(-9999l);
     }
 
-
     @Test
     public void testFindTasks() throws Exception {
         Map<String, Object> parameters = new HashMap<String, Object>();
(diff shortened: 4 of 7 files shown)


nmirasch:
depends on https://github.com/kiegroup/jbpm/pull/1516

cristianonicolai:
jenkins please retest this

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
```diff
@@ -828,17 +828,23 @@ public TaskInstance findTaskByWorkItemId(Long workItemId) {
     }
 
     @Override
-    public TaskInstance findTaskById(Long taskId) {
+    public TaskInstance findTaskById(Long taskId){
+        return findTaskById(taskId, false);
+    }
+
+    @Override
+    public TaskInstance findTaskById(Long taskId, boolean includeSLA) {
```
MarianMacik: Not sure if we need 2 ways of getting a  `TaskInstance` without SLA (either by calling `findTaskById(taskId)` or `findTaskById(taskId, false)`). I understand that this way we can reuse the implementation, but we can also do `findTaskById` and `findTaskByIdWithSLA` and provide one private method as implementation. Just a suggestion, I don't want to be strict too much :)

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
```diff
@@ -1863,6 +1913,7 @@ private void assertTaskInstace(TaskInstance expected, TaskInstance actual) {
         assertEquals(expected.getProcessId(), actual.getProcessId());
         assertEquals(expected.getContainerId(), actual.getContainerId());
         assertEquals(expected.getProcessInstanceId(), actual.getProcessInstanceId());
+        assertEquals(expected.getSlaCompliance(), actual.getSlaCompliance());
```
MarianMacik: Just noticed that here we don't compare workitemId, which is quite a crucial part when we want to compare 2 TaskInstances. Do you mind adding it here?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
MarianMacik: Now I get it, that by doing that you can use just `client.findTaskById(taskId, includeSLA);` in `RemoteTaskServiceImpl#getTask`. So I guess it is just a matter of whether we want to make it more pleasant in our UI component or in the KIE Server client API.

nmirasch: @MarianMacik  there's no problem on changing our UI services, the only thing that stopped me was preserve the API in order to maintain the backwards compatibility. As you prefer, we can keep just the findTaskById(taskId,boolean) o keep both methods

MarianMacik: Yes, there is no doubt we have to preserve the backwards compatibility, but it will be preserved even with 2 separate methods since the first one (`findTaskById(taskId)`) is already there. We would just add the second one, i.e. `findTaskByIdWithSLA(taskId)`. We can also let other reviewers to express their opinions. Wdyt @mswiderski @cristianonicolai ?

nmirasch: @MarianMacik sorry I misunderstood The point is where to do the check. If you think is more clear create the two methods instead pass the boolean, is good to me. ( Believe me during the development I pass throw having both, but I decided pass the boolean). Let me change it

MarianMacik: It was just an idea :) I find more safe to have an API with two methods like `findTaskById(taskId)` and `findTaskByIdWithSLA(taskId)` - I think it is more readable than to have something like `findTaskById(taskId)` and `findTaskById(taskId, true/false)`.

But as I said, we don't need to implement. I would wait for other guys to say their opinion. I don't want to be nitpicking.

nmirasch: updated

sutaakar:
@nmirasch Please address test failures. The changes seem to affect both REST and JMS.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-jbpm/src/main/java/org/kie/server/remote/rest/jbpm/RuntimeDataResource.java
```diff
@@ -528,14 +528,19 @@ public Response getTaskByWorkItemId(@Context HttpHeaders headers,
...
         Header conversationIdHeader = buildConversationIdHeader("", context, headers);
 
         TaskInstance userTaskDesc = null;
         try {
-            userTaskDesc = runtimeDataServiceBase.getTaskById(taskId);
+            if(includeSLA){
+                userTaskDesc = runtimeDataServiceBase.getTaskById(taskId);
+            } else {
+               userTaskDesc = runtimeDataServiceBase.getTaskByIdWithSLA(taskId);
+            }
```
MarianMacik: I think this should be the other way around.

nmirasch: yes, updated.

nmirasch:
jenkins please retest this

cristianonicolai:
jenkins please retest this

sutaakar:
@nmirasch Can you please analyse test failures and apply corrections?

cristianonicolai:
jenkins please retest this

nmirasch:
jenkins please retest this

kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/RuntimeDataServiceIntegrationTest.java
```diff
@@ -1352,19 +1356,130 @@ public void testFindTasks() throws Exception {
...
+
+            taskById = taskClient.findTaskByWorkItemId(workItems.get(0).getId());
+            assertTaskInstance(expectedTaskInstace, taskById);
+        } finally {
+            processClient.abortProcessInstance(CONTAINER_ID, processInstanceId);
+        }
+    }
+
+ /*   @Test
+    public void testFindTasksWithSLA() throws Exception {
```
sutaakar: Can you please remove these commented tests?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java
```diff
@@ -110,6 +110,8 @@
 
     TaskInstance findTaskById(Long taskId);
 
+    TaskInstance findTaskByIdWithSLA(Long taskId);
```
cristianonicolai: I still think that in terms of keeping a standard compare to the same API we should have two methods: findTaskById(Long taskId) and findTaskById(Long taskId, Boolean withSla), that would mean just that the default would be false for reusing the new method, keeping the expected previous behaviour. See an example: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UserTaskServicesClient.java#L102
Usually we also only include what is been filtered on on the method name. My 2cents

nmirasch: @cristianonicolai for me this makes sense. That was something that we talk throw the PR comments. @MarianMacik are we good with this change?

MarianMacik: @cristianonicolai @nmirasch OK, you are right. I thought that we should avoid doing that, but on the other hand we already have this standard. The improvement I suggested would be just a drop in the ocean :) I agree it's better to be consistent.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
```diff
@@ -828,17 +828,25 @@ public TaskInstance findTaskByWorkItemId(Long workItemId) {
...
+        return findTaskById(taskId, false);
+    }
+
+    @Override
+    public TaskInstance findTaskById(Long taskId, boolean withSLA) {
         if( config.isRest() ) {
             Map<String, Object> valuesMap = new HashMap<String, Object>();
             valuesMap.put(TASK_INSTANCE_ID, taskId);
 
+            String queryString = "";
```
cristianonicolai: you probably dont even need this, just send the value of _withSLA_ always?

nmirasch: @cristianonicolai sure, updated, Thanks!

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UserTaskServicesClientImpl.java
```diff
@@ -828,17 +828,21 @@ public TaskInstance findTaskByWorkItemId(Long workItemId) {
...
             valuesMap.put(TASK_INSTANCE_ID, taskId);
 
             return makeHttpGetRequestAndCreateCustomResponse(
-                    build(loadBalancer.getUrl(), QUERY_URI + "/" + TASK_GET_URI, valuesMap), TaskInstance.class);
-
+                    build(loadBalancer.getUrl(), QUERY_URI + "/" + TASK_GET_URI + "?withSLA=" + withSLA, valuesMap), TaskInstance.class);
 
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList( (KieServerCommand) new DescriptorCommand( "QueryService", "getTaskById", new Object[]{taskId}) ) );
+            CommandScript script =new CommandScript( Collections.singletonList( (KieServerCommand) new DescriptorCommand( "QueryService", "getTaskById", new Object[]{taskId, withSLA}) ) );
```
MarianMacik: Formatting a little bit off.

nmirasch: @MarianMacik let's do the things right :S.. Retriggering. Thanks Marian!

cristianonicolai:
jenkins please retest this

PR #1871, opened 2019-07-19
[JBPM-8497] Document integration with production databases

ruromero:

Fix https://issues.jboss.org/browse/JBPM-8497

### DIFF
--- a/process-migration-service/README.md
+++ b/process-migration-service/README.md
@@ -33,7 +33,7 @@ Is the execution of a defined plan, applied to a set of process instances. These
 
 ## Build
 
-```{bash}
+```bash
 mvn clean package
 ```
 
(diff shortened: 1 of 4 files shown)


process-migration-service/src/main/resources/META-INF/persistence.xml
```diff
@@ -1,18 +1,17 @@
...
-	xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd">
-	<persistence-unit name="migration-unit"
-		transaction-type="JTA">
-		<provider>org.hibernate.jpa.HibernatePersistenceProvider</provider>
-		<class>org.kie.processmigration.model.Migration</class>
-		<class>org.kie.processmigration.model.Plan</class>
-		<class>org.kie.processmigration.model.MigrationReport</class>
-		<properties>
-			<property name="hibernate.max_fetch_depth" value="3" />
-			<property name="hibernate.hbm2ddl.auto" value="create" />
```
MarianMacik: This part has been omitted. So now Hibernate will not create database tables for PIM schema. I guess we should leave create here for now unless we want to create DDL scripts for each database.

process-migration-service/README.md
```diff
@@ -22,46 +22,46 @@ Is the execution of a defined plan, applied to a set of process instances. These
...
 
 * JRE 1.8 or greater
 * Running KIE Process server
 
-# Build
+## Build
 
-```
-$ mvn clean package
+```{bash}
```
MarianMacik: One note. According to GitHub Markdown, this should be like ` ```bash` for syntax highlighting.
https://help.github.com/en/articles/creating-and-highlighting-code-blocks#syntax-highlighting

process-migration-service/src/main/resources/META-INF/persistence.xml
ruromero: If we leave `create` each time the application starts, the database schema and data will be dropped. What do you suggest?

MarianMacik:
Well, according to the [Hibernate documentation](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#configurations-hbmddl), `update` could be the way to go. This won't drop the schema.

PR #1889, opened 2019-08-15
BAQE-982: Remove cargo-core-uberjar from Kie server test dependencies

sutaakar:
To fix build issues.


sutaakar:
Jenkins retest this

PR #1890, opened 2019-08-16
DROOLS-4440 Include whole project in kie-maven-plugin build

baldimir:
Related Drools PR: https://github.com/kiegroup/drools/pull/2507

Fixes that now all project files are included in the build. Originally only resources were included, however when having a project pom.xml included in the build, the projects dependencies are automatically added to the build. This fixes a scenario when one KieBase includes another KieBase from a dependency KJar. 

JIRA: https://issues.jboss.org/browse/DROOLS-4440

@etirelli could you please review as Mario and Luca are on PTO?

### DIFF
--- a/kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
+++ b/kie-maven-plugin/src/main/java/org/kie/maven/plugin/BuildMojo.java
@@ -24,7 +24,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


baldimir:

baldimir:

baldimir:

baldimir:

baldimir:

baldimir:

baldimir:
I took a little more conservative approach with the build files filter. I mean it is more close to what it was before + pom.xml file, because including also the POJO files from the kjar project in the build failed some downstream tests. However I will still need to talk about this with @mariofusco. The behaviour is strange to me, because when a KJar is built now, the build doesn't fail when there is a compilation problem in a POJO class from the KJar project. The error is just logged in the build output. I would expect the build to fail when any problem in any KJar file occurs. No matter if DRL or POJO.

baldimir:

PR #1892, opened 2019-08-20
RHPAM-2328: removed unsupported 'io.thorntail:bom-all' and 'swagger'

mareknovotny:
https://issues.jboss.org/browse/RHPAM-2328

removed unsupported thorntail deps from PIM

### DIFF
--- a/process-migration-service/pom.xml
+++ b/process-migration-service/pom.xml
@@ -66,13 +66,6 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>io.thorntail</groupId>
-        <artifactId>bom-all</artifactId>
-        <version>${version.thorntail}</version>
-        <scope>import</scope>
-        <type>pom</type>
-      </dependency>
-      <dependency>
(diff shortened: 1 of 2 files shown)


ruromero:
So I understand that what is unsupported is the Swagger fraction, right?

MarianMacik:
That's a good question. I forgot that we use it in one class. And it compiles beause swagger-annotations are on the classpath. So I think we should ask if there is a supported swagger fraction, otherwise we cannot have Swagger here. @mareknovotny what do you think?

mareknovotny:
We can remove it or keep just the swagger annotation

ruromero:
> We can remove it or keep just the swagger annotation

Then let's remove also the Swagger annotations (and dependency) until this fraction is stable. If that ever happens :)

mareknovotny:
done @ruromero, @MarianMacik

mareknovotny:
Jenkins please retest this

mareknovotny:
Jenkins please retest this

PR #1895, opened 2019-08-22
Add test coverage measurement for droolsjbpm-integration repo

sutaakar:
@rsynek Can you please take a look?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -166,5 +166,47 @@
         <module>drools-examples-android</module>
       </modules>
     </profile>
+    <profile>
+      <id>run-code-coverage</id>
+      <properties>
+        <jacoco.excludes></jacoco.excludes>
+        <jacoco.agent.line>-javaagent:${settings.localRepository}/org/jacoco/org.jacoco.agent/${version.jacoco.plugin}/org.jacoco.agent-${version.jacoco.plugin}-runtime.jar=destfile=${jacoco.exec.file},append=true,excludes=${jacoco.excludes}</jacoco.agent.line>
+        <surefire.argLine>
+          -Dfile.encoding=${project.build.sourceEncoding}
+          ${jacoco.agent.line}
+        </surefire.argLine>
+      </properties>
+      <build>
+        <pluginManagement>
+          <plugins>
+            <plugin>
+              <groupId>org.codehaus.cargo</groupId>
+              <artifactId>cargo-maven2-plugin</artifactId>
+              <configuration>
+                <configuration>
+                  <properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -166,5 +166,71 @@
         <module>drools-examples-android</module>
       </modules>
     </profile>
+    <profile>
+      <id>run-code-coverage</id>
+      <properties>
+        <jacoco.excludes></jacoco.excludes>
+        <jacoco.exec.file>${project.build.directory}/jacoco.exec</jacoco.exec.file>
```
rsynek: Let's take the jacoco.exec.file from kie-parent, which resides in the project's root directory/target, to fully benefit from the append mode and get cross-module coverage.

pom.xml
```diff
@@ -166,5 +166,71 @@
...
+          <groupId>org.jacoco</groupId>
+          <artifactId>org.jacoco.agent</artifactId>
+          <classifier>runtime</classifier>
+          <scope>test</scope>
+        </dependency>
+      </dependencies>
+      <build>
+        <pluginManagement>
+          <plugins>
+            <plugin>
```
rsynek: Eventually this entire plugin declaration can be removed if we use JaCoCo in the java agent mode including unit tests. That means you have to pass the agent line property to surefire plugin instead of the reference to the jacoco.exec file and also declare dependency from surefire on the agent. See https://github.com/kiegroup/optaplanner/blob/master/pom.xml#L147

pom.xml
```diff
@@ -166,5 +166,71 @@
...
+                <systemPropertyVariables>
+                  <jacoco-agent.destfile>${jacoco.exec.file}</jacoco-agent.destfile>
+                </systemPropertyVariables>
+              </configuration>
+            </plugin>
+          </plugins>
+        </pluginManagement>
+        <plugins>
+          <plugin>
+            <groupId>org.jacoco</groupId>
```
rsynek: can be removed - see above

pom.xml
```diff
@@ -166,5 +166,71 @@
...
+                  <artifactId>org.jacoco.agent</artifactId>
+                  <version>${version.jacoco.plugin}</version>
+                </dependency>
+              </dependencies>
+            </plugin>
+            <plugin>
+              <artifactId>maven-surefire-plugin</artifactId>
+              <configuration>
+                <systemPropertyVariables>
+                  <jacoco-agent.destfile>${jacoco.exec.file}</jacoco-agent.destfile>
```
rsynek: Let's pass here the agent line instead of jacoco.exec file and declare dependency on the agent jar.

sutaakar:
@rsynek Comments addressed, please have another look.

PR #1901, opened 2019-08-27
[JBPM-8725] Invalid REST API return type - quated string instead of json/xml (DO NOT MERGE/DISCUSSION)

elguardian:
adding the posibility to return strict json object when only id are returned

### DIFF
--- a/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/rest/RestEasy960Util.java
+++ b/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/rest/RestEasy960Util.java
@@ -16,6 +16,7 @@
 package org.kie.server.common.rest;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import javax.ws.rs.core.HttpHeaders;
@@ -37,12 +38,14 @@ public class RestEasy960Util {
     private static final String ACCEPT_LANGUAGE = "Accept-Language";
    
     public static List<Variant> variants 
-        = Variant.mediaTypes(MediaType.APPLICATION_XML_TYPE, MediaType.APPLICATION_JSON_TYPE).add().build();
+            = Variant.mediaTypes(MediaType.APPLICATION_XML_TYPE, MediaType.APPLICATION_JSON_TYPE,
+                                 new MediaType("application", "json", Collections.singletonMap("strict", "true")),
+                                 new MediaType("application", "xml", Collections.singletonMap("stritct", "true"))).add().build();
     public static Variant defaultVariant 
         = Variant.mediaTypes(MediaType.APPLICATION_XML_TYPE).add().build().get(0);
     public static final Variant jsonVariant 
         = Variant.mediaTypes(MediaType.APPLICATION_JSON_TYPE).add().build().get(0);
-    
+
     public static Variant getVariant(HttpHeaders headers) { 
         // copied (except for the acceptHeaders fix) from RestEasy's RequestImpl class
         ServerDrivenNegotiation negotiation = new ServerDrivenNegotiation();
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/WrapperMessageBodyWriter.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/WrapperMessageBodyWriter.java
@@ -1,0 +1,55 @@
+package org.kie.server.remote.rest.common.util;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Map;
+
+import javax.ws.rs.WebApplicationException;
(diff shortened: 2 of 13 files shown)


elguardian:
@MarianMacik This is PR talked this morning.
I just did it for json. for xml we need a wrapped object or something. Probably something like
<identifier>
 value
</identifier>
for json could be something like
{
   id : "value"
}
TBD. I added a parameter in the media type for controlling whether to return a strict object or not.
Let me know your feedback

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/WrapperMessageBodyWriter.java
```diff
@@ -0,0 +1,59 @@
...
+
+        Map<String, String> parameters = mediaType.getParameters();
+        Boolean isStrict = Boolean.parseBoolean(parameters.get("strict"));
+
+        MarshallingFormat format = MarshallingFormat.fromType(mediaType.toString());
+        Marshaller marshaller = MarshallerFactory.getMarshaller(format, Thread.currentThread().getContextClassLoader());
+        if ((isStrict || STRICT_ID_JSON) && MediaType.APPLICATION_JSON_TYPE.isCompatible(mediaType)) {
+            entityStream.write(marshaller.marshall(identifier).getBytes());
+        } else {
+            entityStream.write(identifier.getId().toString().getBytes());
```
MarianMacik: This is one way how to do it.
The other thing I was thinking about is to just simply create a wrapper JAXB object, something like
https://github.com/kiegroup/droolsjbpm-integration/blob/7d93d03f9de8a68d78d04ba25d71bdc80eaa3112/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/model/definition/ProcessDefinitionList.java#L30-L63

And then we can decide for example directly in a JAX-RS endpoint method whether we want to wrap a returned number inside that wrapper or not based on whether the `strict` is set either in a header or globally.

If it is not set at all, we won't wrap e.g. the process instance id and will return just text.
If it is set, then we will wrap it inside the annotated wrapper and either let JAX-RS take care of it automatically or perhaps use our own marshaller and marshall it.

I think this approach requires less code, works for both XML and JSON and doesn't require a custom BodyWriter. What do you think?

elguardian: Hi Marian, Regarding the spec I argue that your approach is not align.

4.2.2 Message Body Writer
The MessageBodyWriter interface defines the contract between the JAX-RS runtime and components that
provide mapping services from a Java type to a representation. A class wishing to provide such a service
implements the MessageBodyWriter interface and may be annotated with @Provider for automatic
discovery.

So the responsible for providing the representation of a java object is that interface. 

I did wrap already the object in an Wrapper and it is up to the MessageBodyWriter to decided how it is represented. Add that to the endpoint mixes responsabilities. this impl makes more sense to me as it is more align to that definition. In the end when you produce a mime type you are using default providers.

MarianMacik: Well, it was just an idea :) But then we mix responsibilities already with this ProcessDefinitionList object. We don't register any provider for that, we just marshal it ourselves. Also I don't think we use any default providers because before returning the response object we already marshal it using our internal marshaller into String. Which (probably) means that JAX-RS will leave it as is.
As I said, it was just an idea with less code. It would work the same also for XML whereas your approach needs to modify the Provider for XML as well. So it's up to you.

elguardian:
this is ok to test

elguardian:
with this strategy I cannot discrimate among xml/json as marshalling logic is spread all across resources. At the point I am marshalling here is already too late.

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/rest/RestEasy960Util.java
```diff
@@ -37,12 +38,14 @@
     private static final String ACCEPT_LANGUAGE = "Accept-Language";
    
     public static List<Variant> variants 
-        = Variant.mediaTypes(MediaType.APPLICATION_XML_TYPE, MediaType.APPLICATION_JSON_TYPE).add().build();
+            = Variant.mediaTypes(MediaType.APPLICATION_XML_TYPE, MediaType.APPLICATION_JSON_TYPE,
+                                 new MediaType("application", "json", Collections.singletonMap("strict", "true")),
+                                 new MediaType("application", "xml", Collections.singletonMap("stritct", "true"))).add().build();
```
MarianMacik: Just one note that there was a typo in `stritct`. But I guess you tried it locally without it. It's a pity that it didn't work. What about my idea to do it directly in our marshaller?

elguardian: @MarianMacik #1907 the is done in the JSONMarshaller, it is the only way I found to keep backward compatibility.

PR #1903, opened 2019-08-28
[JBPM-8400] Logs are constantly flooded with 'Unable to load key store. Using password from configuration'

elguardian:
added some logic to print the message just the first time

### DIFF
--- a/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
+++ b/kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
@@ -1,5 +1,8 @@
 package org.kie.server.common;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import org.drools.core.util.KeyStoreHelper;
 import org.kie.server.api.KieServerConstants;
 import org.kie.server.api.model.KieServerConfig;
@@ -19,21 +22,11 @@ public class KeyStoreHelperUtil {
 
     private static final Logger logger = LoggerFactory.getLogger(KeyStoreHelperUtil.class);
 
+    private static Set<String> invocationCache = new HashSet<>();
+
     public static String loadServerPassword() {
-        String passwordKey;
-        KeyStoreHelper keyStoreHelper = new KeyStoreHelper();
-
-        try {
-            String pwdKeyAlias = System.getProperty(PROP_PWD_SERVER_ALIAS, "");
-            char[] pwdKeyPassword = System.getProperty(PROP_PWD_SERVER_PWD, "").toCharArray();
-
-            passwordKey = keyStoreHelper.getPasswordKey(pwdKeyAlias, pwdKeyPassword);
-        } catch (RuntimeException re) {
-            logger.warn("Unable to load key store. Using password from configuration");
-            passwordKey = System.getProperty(KieServerConstants.CFG_KIE_PASSWORD, "kieserver1!");
-        }
-
-        return passwordKey;
+        String defaultPassword = System.getProperty(KieServerConstants.CFG_KIE_PASSWORD, "kieserver1!");
+        return loadPasswordKey(PROP_PWD_SERVER_ALIAS, PROP_PWD_SERVER_PWD, defaultPassword);
     }
 
     public static String loadControllerPassword(final KieServerConfig config) {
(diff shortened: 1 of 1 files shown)


kie-server-parent/kie-server-common/src/test/java/org/kie/server/common/KeyStoreHelperUtilTest.java
```diff
@@ -26,6 +26,16 @@
...
 
+    @Test
+    public void testMultipleLoadsPassword() {
+        final String defaultPassword = "default";
+        final String otherDefaultPassword = "otherDefault";
+        String password = loadControllerPassword(defaultPassword);
+        assertEquals(defaultPassword, password);
+        password = loadControllerPassword(otherDefaultPassword);
+        assertEquals(otherDefaultPassword, password);
+    }
```
MarianMacik: Not sure this test is even needed. Is there any special case why we need to test it twice in a row?

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
```diff
@@ -19,38 +22,42 @@
...
+            }
         }
+        return passwordKey;
+    }
+
+    private static boolean hasNotBeenInvoked(String pwdKeyAliasProperty, String pwdKeyPasswordProperty, String passwordKey) {
+        final String key = pwdKeyAliasProperty + "-" + pwdKeyPasswordProperty;
+        String oldPwd = invocationCache.get(key);
+        invocationCache.put(key, passwordKey);
+        return oldPwd == null;
```
MarianMacik: Should we always overwrite the HashMap? I mean we could put there a new value just in case `oldPwd == null` and return true, otherwise return false. Not sure what has a bigger performance impact. We also store password here, maybe just an empty String would be sufficient as a value?

kie-server-parent/kie-server-common/src/test/java/org/kie/server/common/KeyStoreHelperUtilTest.java
elguardian: I did the test to check on local the message was not don twice. I will get rid of it.

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
elguardian: then just maybe a ordered set with the key and check whether exists or not. that is fine for me.

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
```diff
@@ -19,38 +22,44 @@
...
+            if (hasNotBeenInvoked(pwdKeyAliasProperty, pwdKeyPasswordProperty, passwordKey)) {
+                // only print when it has not been invoked
+                logger.warn("Unable to load key store. Using password from configuration");
+            }
+        }
+        return passwordKey;
+    }
+
+    private static boolean hasNotBeenInvoked(String pwdKeyAliasProperty, String pwdKeyPasswordProperty, String passwordKey) {
+        final String key = pwdKeyAliasProperty + "-" + pwdKeyPasswordProperty;
```
MarianMacik: I just noticed that we concatenate 2 properties, but do we need them both? Maybe `pwdKeyAliasProperty` would be enough to distinguish between server and controller keys?

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
```diff
@@ -19,38 +22,44 @@
...
     private static final Logger logger = LoggerFactory.getLogger(KeyStoreHelperUtil.class);
 
-    public static String loadServerPassword() {
-        String passwordKey;
-        KeyStoreHelper keyStoreHelper = new KeyStoreHelper();
-
-        try {
-            String pwdKeyAlias = System.getProperty(PROP_PWD_SERVER_ALIAS, "");
-            char[] pwdKeyPassword = System.getProperty(PROP_PWD_SERVER_PWD, "").toCharArray();
+    private static Set<String> invocationCache = new TreeSet<>();
```
MarianMacik: Actually for `contains` operations the `HashSet` should be even quicker.

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
elguardian: In theory yes, but I don't presume anything.

elguardian:
done @MarianMacik

kie-server-parent/kie-server-common/src/main/java/org/kie/server/common/KeyStoreHelperUtil.java
```diff
@@ -19,38 +22,43 @@
...
     private static final Logger logger = LoggerFactory.getLogger(KeyStoreHelperUtil.class);
 
-    public static String loadServerPassword() {
-        String passwordKey;
-        KeyStoreHelper keyStoreHelper = new KeyStoreHelper();
-
-        try {
-            String pwdKeyAlias = System.getProperty(PROP_PWD_SERVER_ALIAS, "");
-            char[] pwdKeyPassword = System.getProperty(PROP_PWD_SERVER_PWD, "").toCharArray();
+    private static Map<String, String> invocationCache = new HashMap<>();
```
MarianMacik: As we already discussed, `HashSet` is the way to go.

elguardian:
@MarianMacik done with the Hashset

mareknovotny:
Known test failure on master

```
java.lang.ClassNotFoundException: org.drools.compiler.kproject.models.KieBaseModelImpl not found by org.kie.dmn.core.osgi [122]
	at org.kie.karaf.itest.KieDMNKarafIntegrationTest.createKieDMNRuntime(KieDMNKarafIntegrationTest.java:103)
	at org.kie.karaf.itest.KieDMNKarafIntegrationTest.testDMNbasic(KieDMNKarafIntegrationTest.java:71)
```

elguardian:
@MarianMacik can we merge this one ?... Also I did the backport in 7.26.x #1906

PR #1913, opened 2019-10-04
[JBPM-8801] Add explicit createdBy assignment to a task for tests

sutaakar:
@mareknovotny This should fix the FDB issues.


elguardian:
@sutaakar  @mareknovotny  @MarianMacik 
it changes the bpm2 setting the createdBy in stone causing  to fit and pass all the test using the project-defintion kjar in integration test.

let's wait for the outcome but seems ok to me.

PR #1924, opened 2019-10-17
[JBPM-8855] Smart Router management endpoints authentication

elguardian:
adding functionality for kie service instance auth in smart router.

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
@@ -110,6 +110,9 @@ public class KieServerConstants {
     public static final String KIE_SERVER_PARAM_RESET_BEFORE_UPDATE = "KieServerResetBeforeUpdate";
 
     public static final String KIE_SERVER_ROUTER = "org.kie.server.router";
+    public static final String KIE_ROUTER_MANAGEMENT_PASSWORD = "org.kie.server.router.management.password";
+    public static final String KIE_ROUTER_MANAGEMENT_USERNAME = "org.kie.server.router.management.username";
+
     public static final String KIE_SERVER_ROUTER_ATTEMPT_INTERVAL = "org.kie.server.router.connect";
 
     public static final String KIE_OPTAPLANNER_THREAD_POOL_QUEUE_SIZE = "org.optaplanner.server.ext.thread.pool.queue.size";
--- a/kie-server-parent/kie-server-router/kie-server-router-client/src/main/java/org/kie/server/router/client/KieServerRouterEventListener.java
+++ b/kie-server-parent/kie-server-router/kie-server-router-client/src/main/java/org/kie/server/router/client/KieServerRouterEventListener.java
@@ -46,8 +46,12 @@ public class KieServerRouterEventListener implements KieServerEventListener {
     private static final String ROUTER_REMOVE_URL = "/mgmt/remove";
     
     private String serverId = System.getProperty(KieServerConstants.KIE_SERVER_ID);
+
     private String serverURL = System.getProperty(KieServerConstants.KIE_SERVER_LOCATION);
     private String routerURL = System.getProperty(KieServerConstants.KIE_SERVER_ROUTER);
+    private String username = System.getProperty(KieServerConstants.KIE_ROUTER_MANAGEMENT_USERNAME, serverId);
+    private String password = System.getProperty(KieServerConstants.KIE_ROUTER_MANAGEMENT_PASSWORD);
+
     private int failedAttemptsInterval = Integer.parseInt(System.getProperty(KieServerConstants.KIE_SERVER_ROUTER_ATTEMPT_INTERVAL, "10"));
     
     private KieContainerResourceFilter activeOnly = new KieContainerResourceFilter(ReleaseIdFilter.ACCEPT_ALL, KieContainerStatusFilter.parseFromNullableString("STARTED"));
@@ -168,8 +172,15 @@ public class KieServerRouterEventListener implements KieServerEventListener {
     protected boolean send(String url, String containerId, String payload, boolean add, boolean retry) {
 
         try {
-            KieServerHttpRequest httpRequest = KieServerHttpRequest.newRequest(url)
-                    .followRedirects(true)
+            KieServerHttpRequest httpRequest = null;
+            // auth will be used if password is provided. user will be the server id
+            if (password != null && !password.isEmpty()) {
+                httpRequest = KieServerHttpRequest.newRequest(url, username, password);
+            } else {
+                httpRequest = KieServerHttpRequest.newRequest(url);
+            }
+
+            httpRequest.followRedirects(true)
                     .contentType("application/json")
                     .accept("application/json")
                     .timeout(5000)
@@ -221,6 +232,9 @@ public class KieServerRouterEventListener implements KieServerEventListener {
         return alias;
     }
 
+    public String getUsername() {
+        return username;
+    }
     protected List<String> routers() {
         ArrayList<String> list = new ArrayList<>();
         String[] routerUrls = routerURL.split(",");
--- a/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
+++ b/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
@@ -16,19 +16,12 @@
 
 package org.kie.server.router;
 
-import static org.kie.server.router.KieServerRouterConstants.KIE_CONTROLLER;
-import static org.kie.server.router.KieServerRouterConstants.KIE_SERVER_CONTROLLER_ATTEMPT_INTERVAL;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_HOST;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE_KEYALIAS;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE_PASSWORD;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_PORT;
-import static org.kie.server.router.KieServerRouterConstants.ROUTER_PORT_TLS;
-
 import java.net.URI;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -41,6 +34,26 @@ import java.util.concurrent.TimeUnit;
 
 import javax.net.ssl.SSLContext;
 
+import io.undertow.Handlers;
+import io.undertow.Undertow;
+import io.undertow.security.api.AuthenticationMechanism;
+import io.undertow.security.api.AuthenticationMode;
+import io.undertow.security.handlers.AuthenticationCallHandler;
+import io.undertow.security.handlers.AuthenticationConstraintHandler;
+import io.undertow.security.handlers.AuthenticationMechanismsHandler;
+import io.undertow.security.handlers.SecurityInitialHandler;
+import io.undertow.security.idm.IdentityManager;
+import io.undertow.security.impl.BasicAuthenticationMechanism;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.handlers.BlockingHandler;
+import io.undertow.server.handlers.PathHandler;
+import io.undertow.server.handlers.ResponseCodeHandler;
+import io.undertow.server.handlers.proxy.ProxyHandler;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
 import org.jboss.logging.Logger;
 import org.json.JSONArray;
 import org.json.JSONException;
@@ -53,19 +66,22 @@ import org.kie.server.router.handlers.KieServerInfoHandler;
 import org.kie.server.router.handlers.OptionsHttpHandler;
 import org.kie.server.router.handlers.QueriesDataHttpHandler;
 import org.kie.server.router.handlers.QueriesHttpHandler;
+import org.kie.server.router.identity.IdentityService;
 import org.kie.server.router.proxy.KieServerProxyClient;
 import org.kie.server.router.repository.FileRepository;
 import org.kie.server.router.spi.ConfigRepository;
 import org.kie.server.router.utils.HttpUtils;
 import org.kie.server.router.utils.SSLContextBuilder;
 
-import io.undertow.Handlers;
-import io.undertow.Undertow;
-import io.undertow.server.HttpHandler;
-import io.undertow.server.handlers.BlockingHandler;
-import io.undertow.server.handlers.PathHandler;
-import io.undertow.server.handlers.ResponseCodeHandler;
-import io.undertow.server.handlers.proxy.ProxyHandler;
+import static org.kie.server.router.KieServerRouterConstants.KIE_CONTROLLER;
+import static org.kie.server.router.KieServerRouterConstants.KIE_ROUTER_MANAGEMENT_SECURED;
+import static org.kie.server.router.KieServerRouterConstants.KIE_SERVER_CONTROLLER_ATTEMPT_INTERVAL;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_HOST;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE_KEYALIAS;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_KEYSTORE_PASSWORD;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_PORT;
+import static org.kie.server.router.KieServerRouterConstants.ROUTER_PORT_TLS;
 
 public class KieServerRouter {
 
@@ -81,8 +97,13 @@ public class KieServerRouter {
     private static final String KEYSTORE_PASSWORD = System.getProperty(ROUTER_KEYSTORE_PASSWORD);
     private static final String KEYSTORE_KEYALIAS = System.getProperty(ROUTER_KEYSTORE_KEYALIAS);
     private static final boolean TLS_ENABLED = KEYSTORE_PATH != null && !KEYSTORE_PATH.isEmpty();
-    private int failedAttemptsInterval = Integer.parseInt(System.getProperty(KIE_SERVER_CONTROLLER_ATTEMPT_INTERVAL,
-                                                                             "10"));
+    private int failedAttemptsInterval = Integer.parseInt(System.getProperty(KIE_SERVER_CONTROLLER_ATTEMPT_INTERVAL, "10"));
+
+    private static final boolean MANAGEMENT_SECURED = Boolean.parseBoolean(System.getProperty(KIE_ROUTER_MANAGEMENT_SECURED, "false"));
+    private static final String IDENTITY_PROVIDER = System.getProperty(KieServerRouterConstants.KIE_ROUTER_IDENTITY_PROVIDER, "default");
+
+    public static final String CMD_ADD_USER = "addUser";
+    public static final String CMD_REMOVE_USER = "removeUser";
 
     private String CONTROLLER = System.getProperty(KIE_CONTROLLER);
 
@@ -96,6 +117,7 @@ public class KieServerRouter {
             "      \"id\" : \"" + KieServerInfoHandler.getRouterId() + "\"\n" +
             "}";
 
+
     private ServiceLoader<ConfigRepository> configRepositoryServiceLoader = ServiceLoader.load(ConfigRepository.class);
 
     private Undertow server;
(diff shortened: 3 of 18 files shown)


kie-server-parent/kie-server-router/kie-server-router-client/src/main/java/org/kie/server/router/client/KieServerRouterEventListener.java
```diff
@@ -193,7 +202,7 @@ protected boolean send(String url, String containerId, String payload, boolean a
                                 if (add) {
                                     logger.info("Added '{}' as server location for container '{}'", serverURL, containerId);
                                 } else {
-                                    logger.info("Removed '{}' as server location for container '{}'", serverURL, containerId);
```
sutaakar: Please revert the formatting change here.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -189,8 +239,18 @@ public void start(String host,
         pathHandler.addExactPath("/containers",
                                  new ContainersHttpHandler(notFoundHandler,
                                                            adminHandler));
-        pathHandler.addPrefixPath("/mgmt",
-                                  adminHandler);
+        HttpHandler mgmtHandler = adminHandler;
+        if (MANAGEMENT_SECURED) {
+            IdentityManager idm = getIdentityService();
+            mgmtHandler = new AuthenticationCallHandler(mgmtHandler);
```
sutaakar: I think it may be better not to reuse mgmtHandler reference, but rather assign objects to different variables. It will increase code readability.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,94 @@
+package org.kie.server.router.identity;
```
sutaakar: Please add licence header.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/IdentityService.java
```diff
@@ -0,0 +1,18 @@
+package org.kie.server.router.identity;
```
sutaakar: Same here

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/KieServerInstanceAccount.java
```diff
@@ -0,0 +1,41 @@
+package org.kie.server.router.identity;
```
sutaakar: And here

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -109,7 +128,38 @@ public KieServerRouter() {
         log.info("KIE Server router repository implementation is " + repository);
     }
 
-    public static void main(String[] args) {
+    public static void main(String[] args) throws Exception {
+        Options options = new Options();
```
sutaakar: What is the purpose of this CLI adjustment?

elguardian: you need a way to add / remove instances from the identity manager...

sutaakar: ok, thanks

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,94 @@
...
+        data.remove(kieServerInstanceId);
+        saveKieServerInstanceData(data);
+    }
+
+    private Properties loadKieServerInstanceData() {
+        Properties properties = new Properties();
+        try (InputStream is = new FileInputStream(userFile)) {
+            properties.load(is);
+        } catch (Exception e) {
+            log.errorf("file not found %1$s", userFile);
```
sutaakar: Please separate the FileNotFoundException exception handling from other IO exceptions.
FileNotFoundException can be expected in case user creates first entry, in this case warning could be enough, saying that file was not found, empty instance data are returned.
For other IO exception I would leave there error severity.

elguardian:
This is ok to test.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,115 @@
...
+            log.errorf(e, "somo error ocurred during access to %1$s", userFile);
+        }
+        return properties;
+    }
+
+    private void saveKieServerInstanceData(Properties data) {
+        try (OutputStream out = new FileOutputStream(userFile)) {
+            data.store(out, "Kie Server Router");
+        } catch (FileNotFoundException e) {
+            log.infof("file not found %1$s", userFile);
```
sutaakar: FileOutputStream has a different usage of FileNotFoundException than FileInputStream. Here I would propose deleting specific handling of FileNotFoundException and left just IOException, stating that file cannot be created and logging stacktrace.
```
Throws:
FileNotFoundException - if the file exists but is a directory rather than a regular file, does not exist but cannot be created, or cannot be opened for any other reason
```

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -107,10 +126,42 @@
...
         log.info("KIE Server router repository implementation is " + repository);
+        MANAGEMENT_SECURED = Boolean.parseBoolean(System.getProperty(KIE_ROUTER_MANAGEMENT_SECURED, "false"));
+        IDENTITY_PROVIDER = System.getProperty(KieServerRouterConstants.KIE_ROUTER_IDENTITY_PROVIDER, "default");
     }
 
-    public static void main(String[] args) {
+    public static void main(String[] args) throws Exception {
+        Options options = new Options();
+
+        Option addInstanceOption = Option.builder("addInstance")
```
sutaakar: Would it be possible to separate these user adding/deleting into separate class with main method?
That would help functionality separation as this main method already takes care of starting Smart router.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -107,10 +126,42 @@
     public KieServerRouter() {
         configRepositoryServiceLoader.forEach(repo -> repository = repo);
         log.info("KIE Server router repository implementation is " + repository);
+        MANAGEMENT_SECURED = Boolean.parseBoolean(System.getProperty(KIE_ROUTER_MANAGEMENT_SECURED, "false"));
```
sutaakar: Is there a specific reason to keep the initialization of attributes in constructor and not among other constants?

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -331,4 +395,16 @@ protected void boostrapFromControllerResponse(String jsonResponse,
 
         adminHandler.addControllerContainers(containers);
     }
+
+    private IdentityService getIdentityService() {
```
sutaakar: Would it have more sense to provide FileStoredIdentityManager as default IdentityService and use service loader to discover other implementations?
That would be in line with what we already do in https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-standalone-impl/src/main/java/org/kie/server/controller/service/StandaloneKieServerControllerImpl.java
Also that way we can get rid of IdentityService.id as whatever user provide will be used, otherwise default will be used.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
elguardian: yes... base test class provided by QE does not allow to process thos evariables during class loading initialization.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
elguardian: it works like that. you cannot get rid of id, you need a way to differenciate different providers.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
elguardian: sure

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
elguardian: ok

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/IdentityService.java
sutaakar: Please add licence header here.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
sutaakar: ok

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
sutaakar: I see, thanks.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/IdentityService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import io.undertow.security.idm.IdentityManager;
+
+public interface IdentityService extends IdentityManager {
+
+    default String id() {
+        return "default";
```
sutaakar: I think that the default value shouldn't be defined as user is required to specify his service id, otherwise the default value would be used and it wouldn't be clear what implementation is applied (customer one or default one).

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
```diff
@@ -110,6 +110,8 @@
     public static final String KIE_SERVER_PARAM_RESET_BEFORE_UPDATE = "KieServerResetBeforeUpdate";
 
     public static final String KIE_SERVER_ROUTER = "org.kie.server.router";
+    public static final String KIE_ROUTER_MANAGEMENT_PASSWORD = "org.kie.router.management.password";
```
sutaakar: I think it may be better to use org.kie.server.router.management.password to keep the same prefix as for other router related options.

kie-server-parent/kie-server-tests/kie-server-integ-tests-router/src/test/java/org/kie/server/integrationtests/router/rest/KieServerRouterAuthTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.kie.server.controller.client.exception.KieServerControllerHTTPClientException;
+import org.kie.server.integrationtests.router.client.KieServerRouterClient;
+import org.kie.server.integrationtests.shared.basetests.RestOnlyBaseIntegrationTest;
+import org.kie.server.router.Configuration;
+import org.kie.server.router.KieServerRouter;
+import org.kie.server.router.KieServerRouterConstants;
+
+public class KieServerRouterAuthTest extends RestOnlyBaseIntegrationTest {
```
sutaakar: AFAIK the "extends RestOnlyBaseIntegrationTest" is not needed here.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public Account verify(String id, Credential credential) {
+        if (!(credential instanceof PasswordCredential)) {
+            return null;
+        }
+        PasswordCredential passwordCredential = (PasswordCredential) credential;
+        KieServerInstanceAccount account = getAccount(id);
+        String challenge = HashUtil.hash(id, new String(passwordCredential.getPassword()));
+        if (account != null && challenge.equals(account.getHash())) {
+            log.infof("Authentication succesfully for %1$2", id);
```
sutaakar:
```suggestion
            log.infof("Authentication succesfully for %1$s", id);
```

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,115 @@
...
+            return account;
+        }
+        log.infof("Authentication failed for %1$2", id);
```
sutaakar:
```suggestion
        log.infof("Authentication failed for %1$s", id);
```

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,115 @@
...
+import org.jboss.logging.Logger;
+
+public class FileStoredIdentityManager implements IdentityService {
+
+    protected static final Logger log = Logger.getLogger(FileStoredIdentityManager.class);
+
+    private File userFile;
+
+    public FileStoredIdentityManager() {
+        userFile = new File("users.properties");
```
sutaakar: Would it be possible to make location of users.properties configurable?

sutaakar: Something similar to https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/repository/FileRepository.java#L39

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/IdentityService.java
elguardian: ok.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
elguardian: @sutaakar  I don't see this clear now... it is not the same to do java -jar <jar name> options than java -jar <jar name> <class> <options> ... I think it is more cumbersome

sutaakar: @elguardian Do you plan some changes related to options?

elguardian: @sutaakar as I stated before. No, it does not make sense; it ise less user friendly.

elguardian:
Jenkins restest this please.

sutaakar:
@elguardian Could you please increase test coverage, for example by adding https://github.com/elguardian/droolsjbpm-integration/pull/1/files#diff-4828b795501ad6c187d13a873a11a5cf ?

elguardian:
@sutaakar done, resolved code smells too

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/IdentityService.java
```diff
@@ -0,0 +1,16 @@
+package org.kie.server.router.identity;
```
sutaakar: Please add licence header.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,118 @@
...
+            return account;
+        }
+        log.infof("Authentication failed for %1$s", id);
+        return null;
+    }
+
+    private KieServerInstanceAccount getAccount(String id) {
+        Map<String, KieServerInstanceAccount> users = new HashMap<>();
+        Properties file = loadKieServerInstanceData();
+        file.stringPropertyNames().forEach(name -> users.put(name, new KieServerInstanceAccount(name, file.getProperty(name))));
```
MarianMacik: What about just doing `String pwd = file.getProperty(id)` here and then create an instance of KieServerInstanceAccount from it using `return new KieServerInstanceAccount(id, pwd)`? We don't need to iterate over the whole file all the time, do we?

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,118 @@
...
+            log.errorf(e, "An error ocurred. File <%1$s> is a directory, it does not exist or it cannot be opened", userFile);
+        }
+        return properties;
+    }
+
+    private void saveKieServerInstanceData(Properties data) {
+        try (OutputStream out = new FileOutputStream(userFile)) {
+            data.store(out, "Kie Server Router");
+        } catch (IOException e) {
+            log.errorf(e, "An error ocurred. File <%1$s> is a directory, it does not exist or it cannot be opened", userFile);
```
MarianMacik: As @sutaakar pointed out, when opening a FileOutputStream the possible reasons for FileNotFoundException are:
```
Throws:
FileNotFoundException - if the file exists but is a directory rather than a regular file, does not exist but cannot be created, or cannot be opened for any other reason
```

So the line should be

`"An error ocurred. File <%1$s> is a directory, it cannot be created or it cannot be opened"`

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
```diff
@@ -0,0 +1,118 @@
...
+        data.remove(kieServerInstanceId);
+        saveKieServerInstanceData(data);
+    }
+
+    private Properties loadKieServerInstanceData() {
+        Properties properties = new Properties();
+        try (InputStream is = new FileInputStream(userFile)) {
+            properties.load(is);
+        } catch (IOException e) {
+            log.errorf(e, "An error ocurred. File <%1$s> is a directory, it does not exist or it cannot be opened", userFile);
```
MarianMacik: What happens when we create the first entry? Then I guess a warning should be enough. I think the separation of FileNotFoundException from IOException might be a good idea here - just to distinguish WARN from ERROR.

kie-server-parent/kie-server-tests/kie-server-integ-tests-router/src/test/java/org/kie/server/integrationtests/router/rest/KieServerRouterAuthTest.java
```diff
@@ -0,0 +1,104 @@
...
+        repository = new File("target/standalone-router-repo");
+        repository.mkdirs();
+        System.setProperty(KieServerRouterConstants.ROUTER_REPOSITORY_DIR, repository.getAbsolutePath());
+        System.setProperty(KieServerRouterConstants.KIE_ROUTER_MANAGEMENT_SECURED, "true");
+        System.setProperty(KieServerRouterConstants.KIE_ROUTER_IDENTITY_PROVIDER, "mock");
+
+
+        // setup and start router
+        Integer port = allocatePort();
+        router = new KieServerRouter(true, "mock");
```
MarianMacik: Is this needed if we are setting this using system properties? Would the `new KieServerRouter()` do the same thing?

sutaakar: I think that the system properties should be dropped, as they are not used.

kie-server-parent/kie-server-router/kie-server-router-client/src/main/java/org/kie/server/router/client/KieServerRouterEventListener.java
```diff
@@ -168,8 +170,15 @@ public void afterContainerDeactivated(KieServer kieServer, KieContainerInstance
     protected boolean send(String url, String containerId, String payload, boolean add, boolean retry) {
 
         try {
-            KieServerHttpRequest httpRequest = KieServerHttpRequest.newRequest(url)
-                    .followRedirects(true)
+            KieServerHttpRequest httpRequest = null;
+            // auth will be used if password is provided. user will be the server id
```
sutaakar: @elguardian Would it be possible to add optional system property specifying username? And if username is not specified then Kie server id is used?
That would allow customers to specify credentials according their needs.

elguardian: no.

kie-server-parent/kie-server-tests/kie-server-integ-tests-router/src/test/java/org/kie/server/integrationtests/router/rest/KieServerRouterAuthTest.java
elguardian: It is convenience for unit testing. so it is not going to changed.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
elguardian: It is an error the file should be there. The system is not recovering from it (the operation is not being performed). So it cannot be a warning

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
elguardian: fine

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/identity/FileStoredIdentityManager.java
elguardian: fine

elguardian:
@sutaakar @gmunozfe @MarianMacik I think I didn't leave anything behind.

kie-server-parent/kie-server-router/kie-server-router-client/src/main/java/org/kie/server/router/client/KieServerRouterEventListener.java
MarianMacik: @elguardian To be honest, this seems to me like a big overkill. Imagine users having tens of KIE Servers connected to one Smart Router. Then they will have to create a separate account for each kie server just to have the connection secured. Moreover, if they want to change the password, they will have to change all lines in the property file. The idea of @sutaakar is quite good actually. Of course, if they want to have it per kie server, it will be possible. But I imagine that vast majority will use just one or two accounts for many kie servers. It is the same as with kie server connection to controller.

elguardian: public static final String KIE_ROUTER_MANAGEMENT_USERNAME = "org.kie.server.router.management.username";

Overkiliing is having this behavior in here for several reviews and reviewers and change it now

kie-server-parent/kie-server-tests/kie-server-integ-tests-router/src/test/java/org/kie/server/integrationtests/router/rest/KieServerRouterAuthNoIdPTest.java
```diff
@@ -0,0 +1,73 @@
...
+        router.stop(true);
+        router = null;
+
+        Stream.of(repository.listFiles()).forEach(f-> f.delete());
+        repository.delete();
+    }
+
+    @Test
+    public void testValidAuthMgmtKieServerRouter() throws Exception {
+        router = new KieServerRouter();
```
gmunozfe: @elguardian this test is failing because this constructor needs to be updated to received the identity provider.
```suggestion
        router = new KieServerRouter(true, "fake");
```

elguardian:
@MarianMacik change of the name of the command

elguardian:
this is ok to test.

PR #1928, opened 2019-10-23
[OLD] PLANNER-1622: [Tasks Assignment] Planning model iteration #1

wmedvede:


wmedvede:
Hi @ge0ffrey  @rsynek would you mind review this PR?

NOTE: Iv'e created this PR Yesterday but had to be killed since the CI was dead and even after it was supposed to work again there was no way to make the build for the PR work :). So this last is the correct one.

Finally this PR includes the initial model + constraints, etc., for the task assigning integration.
As it was mentioned the location of the sources might change when this stuff is finally integrated with kie-server, etc. But by now let's start moving this under the jbpm-task-assigning module, etc.

Thanks!

rsynek:
@wmedvede started looking into the PR; will finish my review either later today or tomorrow.

wmedvede:
> @wmedvede started looking into the PR; will finish my review either later today or tomorrow.

Thanks, I'll take a look at the sonar alerts after the review to take a look at any potential changes all together.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TaskHelperTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static final int SIZE = 10;
+    private static final Random RANDOM = new Random();
+
+    private List<User> availableUsers;
+    private List<Group> availableGroups;
+    private Task task;
+
+    @Before
+    public void setUp() {
```
rsynek: After reading through this test case I don't think a list of 10 user and 10 groups is needed - neither is using the `Random` to pick one of them.
Having just 2 users and 2 groups would be sufficient - without keeping them in collections and picking them randomly.
Please see first my comment about eventually getting rid of the `TaskHelper` class.

rsynek:
> > @wmedvede started looking into the PR; will finish my review either later today or tomorrow.
> 
> Thanks, I'll take a look at the sonar alerts after the review to take a look at any potential changes all together.

Walter, SonarCloud check has failed due to 0% test coverage - the reason is trivial. This module does not have any coverage measurement configured yet. We can discuss setting it up in the next meeting.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/resources/org/jbpm/task/assigning/solver/taskAssigningScoreRules.drl
```diff
@@ -0,0 +1,88 @@
...
+    then
+        scoreHolder.addHardConstraintMatch(kcontext, 0, $acceptsAssignedUser);
+end
+
+// ############################################################################
+// Soft constraints
+// ############################################################################
+
+// First level soft constraint for penalizing the assignment of the PLANNING_USER. The PLANNING_USER is by definition
+// a potential owner for all the tasks. But we want to avoid the assignming of this user "as soon as possible".
```
rsynek: Typo in "assignming"

jbpm-task-assigning/jbpm-task-assigning-core/src/main/resources/org/jbpm/task/assigning/solver/taskAssigningScoreRules.drl
```diff
@@ -0,0 +1,88 @@
...
+// The PLANNING_USER is e.g. a jBPM user with administration grants for enabling the management of Tasks that couldn't
+// be assigned to anybody. e.g. if a Task needs to be assigned to user that belongs to group HR, but no users with that
+// group exists the PLANNING_USER is assigned instead.
+// This tasks can e.g. be latter manually completed by the PLANNING_USER or manually reassigned
+// to a different jBPM user by using the jBPM tasks lists adminitration, etc.
```
rsynek: Typo "adminitration"

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/BaseTaskAssigningTest.java
```diff
@@ -0,0 +1,155 @@
...
+        });
+    }
+
+    public static String printSolution(TaskAssigningSolution solution) {
+        StringBuilder builder = new StringBuilder();
+        printSolution(solution, builder);
+        return builder.toString();
+    }
+
+    public static String printUser(User user) {
```
rsynek: Suggestion: these printing methods could be actually implemented as `toString()` of the particular classes they print information about.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/BaseTaskAssigningTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public static final String _50TASKS_5USERS_SOLUTION = "/data/unsolved/50tasks-5users.xml";
+    public static final String _100TASKS_5USERS_SOLUTION = "/data/unsolved/100tasks-5users.xml";
+    public static final String _500TASKS_20USERS_SOLUTION = "/data/unsolved/500tasks-20users.xml";
+
+    protected boolean writeTestFiles() {
+        return Boolean.parseBoolean(System.getProperty("org.jbpm.task.assigning.test.writeFiles", "false"));
+    }
+
+    protected SolverFactory<TaskAssigningSolution> createSolverFactory() {
+        SolverFactory<TaskAssigningSolution> solverFactory = SolverFactory.createEmpty();
```
rsynek: The `createEmpty()` method has been deprecated recently, please see https://github.com/kiegroup/optaplanner/pull/577

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/BaseTaskAssigningTest.java
```diff
@@ -0,0 +1,155 @@
...
+        SolverConfig config = solverFactory.getSolverConfig();
```
rsynek: ditto

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Test
+    public void isLowLevelTest() {
+        Stream.of(7, 8, 9, 10).forEach(level -> assertTrue(PriorityHelper.isLowLevel(level)));
+    }
+
+    @Test
+    public void calculateWeightedPenaltySuccessfulTest() {
+        int priority = new Random().nextInt(11);
```
rsynek: There is hardly ever a reason to use `Random` in tests as we want to have tests 100% stable.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class PriorityHelperTest {
+
+    @Rule
```
rsynek: nitpicking: `Rule` annotation is no longer supported in jUnit 5 (thinking about future migration).

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Given a TaskAssigningSolution asserts the following constraints.
+     * <p>
+     * 1) All tasks are assigned to a user
+     * 2) The assigned user for a task is a potentialOwner for the task or the PLANNING_USER
+     * 3) All tasks are assigned.
+     * @param solution a solution.
+     */
+    private void assertConstraints(TaskAssigningSolution solution) {
+        int[] totalTasks = {0};
```
rsynek: If we always access just the 0 index of the array, do we need an array at all?

wmedvede:
Hi @rsynek , so according to last meeting I'll move this stuff to a separate feature branch "jbpm-task-assigning". But let's complete the review on this one so just let me know when you finish the review and we continue with the normal procedure. As soon this PR is ready I'll create the new branch and push this stuff in.

Thanks

rsynek:
@wmedvede +1 I was thinking about the same approach. I realise I put there rather a big number of comments (and I still need to cover the tests in this PR), as I might be biased towards a code style used in OptaPlanner. If there is anything that needs deeper discussion, feel free to bring it up on the chat.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/User.java
```diff
@@ -0,0 +1,138 @@
...
+        return this;
+    }
+
+    @Override
+    public boolean isUser() {
+        return true;
+    }
+
+    @Override
+    public Integer getEndTime() {
```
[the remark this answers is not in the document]
wmedvede: It makes sense due to the nature of how the chained graph is built. It's a bit tricky but it's required for the way optaplanner builds the graph with the solutions etc.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/solver/PriorityHelper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public static boolean isMediumLevel(int priority) {
+        return priority == 3 || priority == 4 || priority == 5 || priority == 6;
+    }
+
+    public static boolean isLowLevel(int priority) {
+        return priority == 7 || priority == 8 || priority == 9 || priority == 10;
+    }
+
+    public static int calculateWeightedPenalty(int priority, int endTime) {
```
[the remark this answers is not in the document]
wmedvede: At this method the unit of time is not relevant since penalization proportional to the value of the endTime no matter the unit.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/solver/PriorityHelper.java
```diff
@@ -0,0 +1,51 @@
...
+import org.jbpm.task.assigning.TaskAssigningRuntimeException;
+
+/**
+ * Helper class for manging priority calculations for tasks coming from the jBPM runtime.
+ * By convention jBPM tasks priorities goes from  [0 (high),... 5 (medium),... 10 (low)]
+ */
+public class PriorityHelper {
+
+    public static boolean isHighLevel(int priority) {
+        return priority == 0 || priority == 1 || priority == 2;
```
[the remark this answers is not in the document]
wmedvede: here we really only wants to have  0, 1 and 2 under the same category. Not all integers <=2

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/solver/PriorityHelper.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.jbpm.task.assigning.model.solver;
+
+import org.jbpm.task.assigning.TaskAssigningRuntimeException;
+
+/**
+ * Helper class for manging priority calculations for tasks coming from the jBPM runtime.
+ * By convention jBPM tasks priorities goes from  [0 (high),... 5 (medium),... 10 (low)]
+ */
+public class PriorityHelper {
```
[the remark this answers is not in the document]
wmedvede: A see this more like a helper class right now.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/solver/TaskHelper.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.jbpm.task.assigning.model.OrganizationalEntity;
+import org.jbpm.task.assigning.model.Task;
+import org.jbpm.task.assigning.model.TaskOrUser;
+import org.jbpm.task.assigning.model.User;
+
+public class TaskHelper {
```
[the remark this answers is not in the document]
wmedvede: I prefer to keep the model classes as plain as possible manage more "complex" methods in helper classes.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/BaseTaskAssigningTest.java
wmedvede: I prefer to not by now. I really came up with this "printXXX" methods for being able to see the test results in a tabular manner, etc. And I can change them free if I have particular needs related with the test.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
wmedvede: only option I see here is to add a try {} catch {} (or variants) block to emulate what the @Rule already does. But I'm not sure that this is the best strategy. By following this sort of handmade strategy we'll end-up in a situation were different programmers implement different hooks, and this will be even harder to track when we move to jUnit5. I think the best strategy is to apply the same change in all places at the time we get there.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
wmedvede: This random test were useful to me during development, and I don't believe they might cause issues. But what I can to is to keep them here but  @Ignored so I can go back to them when I need or want to run them locally.

wmedvede: Ah, sorry, this particular random test can be made non random. I was refering to some other tests that do random operations on the solver.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
wmedvede: this variable is used to collect information inside a solution.getUserList().forEach at the time we do other processing.  So I think it's ok.

jbpm-task-assigning/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+  <packaging>pom</packaging>
+
+  <name>jBPM Task Assigning</name>
+  <description>jBPM Task Assigning</description>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.kie.soup</groupId>
+        <artifactId>kie-soup-bom</artifactId>
```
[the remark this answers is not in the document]
wmedvede: Correct, right now we can build without it. But in next PR it's needed. So I think we can keep it with no problem.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
```diff
@@ -0,0 +1,333 @@
...
+    /**
+     * When the previousTask changes we need to update the startTime for current task and also the startTime
+     * for all the tasks that comes after.  previousTask -> currentTask -> C -> D -> E since each task can only start
+     * after his previous one has finished. As part of the update the endTime for the modified tasks will also be updated.
+     */
+    @CustomShadowVariable(variableListenerClass = StartAndEndTimeUpdatingVariableListener.class,
+            // Arguable, to adhere to API specs (although this works), nextTask and user should also be a source,
+            // because this shadow must be triggered after nextTask and user (but there is no need to be triggered by those)
+            sources = {@PlanningVariableReference(variableName = "previousTaskOrUser")})
+    private Integer startTime; // In minutes
```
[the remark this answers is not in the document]
wmedvede: hmmm, I don't like to bind the "unit" to the variable name

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TaskHelperTest.java
wmedvede: was changed to 2

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
```diff
@@ -0,0 +1,333 @@
...
+ * <p>
+ * Shadow variable:
+ * Let all Tasks have a reference to the anchor of the chain, the assigned user.
+ * @AnchorShadowVariable(sourceVariableName = "previousTaskOrUser")
+ * private User user;
+ * <p>
+ * CustomShadowVariable startTime:
+ * Additionally is a convenient shadow variable is declared for having the startTime of a task already calculated.
+ * @CustomShadowVariable(variableListenerClass = StartTimeUpdatingVariableListener.class,
+ * // Arguable, to adhere to API specs (although this works), nextTask and user should also be a source,
```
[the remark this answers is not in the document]
wmedvede: Haven't seen this comment when reviewing. 
Well, this comment present in the other places in optaplanner when this chained graph structure is being used. So thought it could make sense to keep it. If not it can be removed.

wmedvede:
Hi @rsynek I've just pushed the review changes plus some comments, etc. This is ready from my side.
Thanks

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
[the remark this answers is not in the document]
rsynek: The point is that if the variable name is descriptive enough, there is no need for the comment, which brings following benefits:
- meaning of the variable is clear not only at the declaration, but anywhere it's used
- comments often become obsolete as the code evolves - because with each relevant change, someone has to think about updating the comment too

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/BaseProblemFactChangeTest.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    protected TaskAssigningSolution executeSequentialChanges(TaskAssigningSolution solution, List<? extends ProgrammedProblemFactChange> changes) throws Exception {
+        Solver<TaskAssigningSolution> solver = createDaemonSolver();
+
+        //store the first solution that was produced by the solver for knowing how things looked like at the very
+        //beginning before any change was produced.
+        final TaskAssigningSolution[] initialSolution = {null};
+        final boolean[] changesInProgress = {false};
```
rsynek: Potential race condition.

Although this might be a viable solution for changing the flag inside the lambda, but the variable is accessed by two threads simultaneously. One way of dealing with this could be using `AtomicReference`:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicReference.html

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AssignTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,320 @@
...
+     * asserts that the assignment defined by the change is not violated (exists in) by the solution.
+     * The assignment defined in the change must also be pinned in the produced solution as well as any other
+     * previous assignment for the given user.
+     * @param change The change that was executed for producing the solution.
+     * @param solution The produced solution.
+     */
+    private void assertAssignTaskProblemFactChangeWasProduced(AssignTaskProblemFactChange change, TaskAssigningSolution solution) {
+        User internalUser = solution.getUserList().stream()
+                .filter(user -> Objects.equals(user.getId(), change.getUser().getId()))
+                .findFirst().orElse(null);
```
rsynek: Instead of asserting nullity of the returned value one line below, we can throw exception straight away.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/RemoveTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.jbpm.task.assigning.model.Task;
+import org.jbpm.task.assigning.model.TaskAssigningSolution;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+
+public class RemoveTaskProblemFactChangeTest extends BaseProblemFactChangeTest {
```
rsynek: The same concerns as for the other tests in the `realtime` package.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AssignTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,320 @@
...
+        }
+        assignTaskProblemFactChangeTest(solution, solutionResource, RANDOM_TEST, programmedChanges);
+    }
+
+    private void assignTaskProblemFactChangeTest(TaskAssigningSolution solution,
+                                                 String solutionResource,
+                                                 String testType,
+                                                 List<ProgrammedAssignTaskProblemFactChange> programmedChanges) throws Exception {
+        TaskAssigningSolution initialSolution = executeSequentialChanges(solution, programmedChanges);
+        if (writeTestFiles()) {
```
rsynek: soft suggestion: let's move writing the files into a separate method in order not to pollute this method with unnecessary details.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
```diff
@@ -0,0 +1,107 @@
...
+import static org.jbpm.task.assigning.model.solver.TaskHelper.isPotentialOwner;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class TestSolver extends BaseTaskAssigningTest {
+
+    private static final long MILLISECONDS_TIME_SPENT_LIMIT = 20000;
```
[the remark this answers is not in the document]
wmedvede: correct, it's a "good enough" value to get a feasible solution. (On the basis that It's know by construction of the model and the input data set that a feasible solution can be produced).

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test
+    public void startSolverAndSolution24Tasks8UsersTest() throws Exception {
+        testSolverStartAndSolution(MILLISECONDS_TIME_SPENT_LIMIT, _24TASKS_8USERS_SOLUTION);
+    }
+
+    @Test
+    public void startSolverAndSolution50Tasks5UsersTest() throws Exception {
```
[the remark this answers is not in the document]
wmedvede: It was useful to me when I started working with optaplanner for the first time to consider different data sets and be sure my modelling/programming worked well, etc. But if it implies a huge penalization some of this methods can be tagged as @Ignored to not affect the normal tests execution and still have them available if I need to run them at a later point

jbpm-task-assigning/jbpm-task-assigning-core/src/main/resources/org/jbpm/task/assigning/solver/taskAssigningScoreRules.drl
```diff
@@ -0,0 +1,88 @@
...
+    when
+        Task(user != null && User.PLANNING_USER.getEntityId().equals(user.getEntityId()))
+    then
+        // a penalization is added each time the PLANNING_USER is assigned.
+        scoreHolder.addSoftConstraintMatch(kcontext, 0, -1);
+end
+
+rule "High level priority"
+    when
+        $task : Task(user != null, PriorityHelper.isHighLevel(priority))
```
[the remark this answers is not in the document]
wmedvede: I see your suggestion, but I don't like this approach for this use case.
See that at the constraints level we want to minimize constraints in this order

1) higher priority tasks
2) minimize makespan (not related with the priority)
3) medium priority tasks
4) low priority tasks

I believe this split is more readable and also more flexible in case e.g. we want to change the constraints order, or even add a new constraint in the middle, etc.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/resources/org/jbpm/task/assigning/solver/taskAssigningScoreRules.drl
```diff
@@ -0,0 +1,88 @@
...
+    then
+        scoreHolder.addSoftConstraintMatch(kcontext, 1, PriorityHelper.calculateWeightedPenalty($task.getPriority(), $task.getEndTime()));
+end
+
+rule "Minimze makespan (starting with the latest ending user first)"
+    when
+        Task(user != null, nextTask == null, $endTime : endTime)
```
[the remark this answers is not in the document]
wmedvede: yes, here we have a "trick" again derived from the way optaplanner builds the chained graph, etc.
Basically the Task that has  nextTask == null is the last task of the chain.

User -> Task1.nextTask -> Task2.nextTask ->Task3.nextTask == null.

And the last task the one that has to "totalTime" (endTime) that takes to execute all the chain. So to minimize the make span, we need to consider the "totalTime" of all the chains in the solution.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/resources/org/jbpm/task/assigning/solver/taskAssigningScoreRules.drl
```diff
@@ -0,0 +1,88 @@
...
+    then
+        // Fairness and load balancing trick (see docs): squared to assure correctness in corner cases
+        scoreHolder.addSoftConstraintMatch(kcontext, 2, - ($endTime * $endTime));
```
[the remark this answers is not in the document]
wmedvede: Sure I will, but I have to say that this formula is being using under @ge0ffrey 's benediction. I'm not yet expert at this part, so I'll take a look and see if there's suitable improvement for this use case. Thanks for pointing this out.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AddTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Test
+    public void addTaskProblemFactChange500Tasks5UsersRandomTest() throws Exception {
+        addTaskProblemFactChangeRandomSetTest(_500TASKS_20USERS_SOLUTION);
+    }
+
+    @Test
+    public void addTaskProblemFactChangeTaskAlreadyExistsTest() throws Exception {
+        TaskAssigningSolution solution = readTaskAssigningSolution(_24TASKS_8USERS_SOLUTION);
+        long taskId = new Random().nextInt(solution.getTaskList().size());
```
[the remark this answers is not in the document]
wmedvede: It's a matter of choice, it's not strictly necessary. What the test tries check is, "no matter which task you pick" if you try to add it again to the solution you'll get an error.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AddTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.stream.Collectors;
+
+import org.jbpm.task.assigning.model.Task;
+import org.jbpm.task.assigning.model.TaskAssigningSolution;
+import org.jbpm.task.assigning.model.User;
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class AddTaskProblemFactChangeTest extends BaseProblemFactChangeTest {
```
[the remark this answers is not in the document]
wmedvede: Oh, yes, it was useful to me during development to challenge my implementation against different  datasets with different sizes etc. If it represents a problem probably because it might slowdown the tests processing time, I can @Ignore some of them and keep them there in case I want to go back into them for running my stuff locally

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AddTaskProblemFactChangeTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Test
+    public void addTaskProblemFactChange24Tasks5UsersTest() throws Exception {
+        addTaskProblemFactChangeTest(_24TASKS_8USERS_SOLUTION, Arrays.asList(24L, 25L, 30L, 40L));
+    }
+
+    @Test
+    public void addTaskProblemFactChange24Tasks5UsersRandomTest() throws Exception {
+        addTaskProblemFactChangeRandomSetTest(_24TASKS_8USERS_SOLUTION);
```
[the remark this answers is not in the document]
wmedvede: same as before, it was useful to me under development. I have also tests that covers a deterministic set of changes.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/BaseProblemFactChangeTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        final Semaphore programNextChange = new Semaphore(0);
+        final Semaphore allChangesWereProduced = new Semaphore(0);
+        final Semaphore executorThreadStarted = new Semaphore(0);
```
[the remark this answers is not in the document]
wmedvede: good catch, it's not needed or it's redundant, since nothing will be done until a change is produced.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/BaseProblemFactChangeTest.java
```diff
@@ -0,0 +1,149 @@
...
+                    programNextChange.release();
+                } else {
+                    solver.terminateEarly();
+                    allChangesWereProduced.release();
+                }
+            }
+        });
+
+        //Programmed changes producer Thread.
+        ExecutorService executor = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
wmedvede: I think that at the end of the day when we reach let's say the last line of the test, the only "Runnable" command that the Executor is running has finished, since the Test can only finish when all the changes were produced.
So the only active thread that the executor had, was already finished his work. So at this point when the test finish and the JVM finishes the executor will have a graceful die.

But your suggestion is good. +1

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/BaseProblemFactChangeTest.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+        });
+
+        executorThreadStarted.acquire();
+        solver.solve(solution);
+
+        assertTrue(programmedChanges.isEmpty());
+        assertEquals(totalProgrammedChanges, scheduledChanges.size());
+        assertEquals(0, pendingChanges[0]);
+        return initialSolution[0];
```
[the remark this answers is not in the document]
wmedvede: I want to keep the initial solution because in scenarios (runs) when I print the produced solutions during the test execution in files, I can see "how the solution looked like" before applying any of the changes. And then I print how the solution looks like after applying change#1, change#2 and so on. I was very useful to me during development to have a visual tracking that things worked well. By default the files printing is disabled so it's no a problem to keep this.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/BaseProblemFactChangeTest.java
wmedvede: Agreed, atomic reference is being used in other places where it's really needed. (will came up in next PR). But in this particular case by construction this potential concurrency scenario will never happen thanks to the use of programNextChange.acquire();

But no problem, it can be changed.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AssignTaskProblemFactChangeTest.java
wmedvede: I agree, another valid option could be to throw an Exception. But for this case I like more the current one.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/RemoveTaskProblemFactChangeTest.java
wmedvede: more or less same explanation if you refer to the random tests

wmedvede:
Hi @rsynek I've just pushed a new commit with some more changes regarding your last review.
Let's try to move this forward, so I can remove this branch and create the feature branch we talked about.

Thanks!

wmedvede:
Hi  @triceo I'll remove @ge0ffrey from this thread since it won't be feasible for him to follow all the PRs.
Would you mind take a look?
Thanks.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/AbstractPersistable.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.jbpm.task.assigning.model;
+
+import java.io.Serializable;
+
+import org.apache.commons.lang3.builder.CompareToBuilder;
+import org.optaplanner.core.api.domain.lookup.PlanningId;
+
+public abstract class AbstractPersistable implements Serializable,
+                                                     Comparable<AbstractPersistable> {
```
triceo: A couple weeks ago, we've refactored `AbstractPersistable` to no longer be `Comparable`.
Just FYI, in case you want to keep the code in sync with `optaplanner-examples`.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
```diff
@@ -0,0 +1,339 @@
...
+    @Override
+    public Integer getEndTime() {
+        return endTime;
+    }
+
+    public void setEndTime(Integer endTime) {
+        this.endTime = endTime;
+    }
+
+    public int getDuration() {
```
triceo: Could we use `Duration` here? We will then avoid the problem of what unit this is in.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
```diff
@@ -0,0 +1,333 @@
...
+    // ************************************************************************
+    // Complex methods
+    // ************************************************************************
+
+    /**
+     * Indicates if the currently assigned user can execute this task. If this is the case, the user is a potential
+     * owner of the task.
+     * @return 0 if the assigned user can execute this task, -1 in any other case.
+     */
+    public int acceptsAssignedUser() {
```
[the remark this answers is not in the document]
triceo: I agree.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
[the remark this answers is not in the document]
triceo: Also, there are several uses of time in various classes in this PR. Several methods in several classes. Adding the unit to the name of these methods will make it clear to the user that, if I have `Task.getEndTime()` and `TaskOrUser.getEndTime()`, that it is the same unit and that I can perform the same operations on them.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Test
+    public void calculateWeightedPenaltyFailureTest() {
+        String expectedMessage = "Task priority %s is out of range. " +
+                "A valid priority value must be between 0 (inclusive) " +
+                " and 10 (inclusive)";
+        Stream.of(-2, -1, 11, 12).forEach(priority -> {
+            expectedException.expectMessage(String.format(expectedMessage, priority));
+            PriorityHelper.calculateWeightedPenalty(priority, 1234);
```
triceo: You can replace the `expectedException` by `Assertions.assertThatThrownBy(...)` in AssertJ. 
This is something we're starting to do in Optaplanner core as well.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/StartAndEndTimeUpdatingVariableListenerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        assertEquals((long) (task1.getDuration() + task2.getDuration() + task3.getDuration()), (long) task4.getStartTime());
+        assertEquals((long) (task1.getDuration() + task2.getDuration() + task3.getDuration() + task4.getDuration()), (long) task4.getEndTime());
+        assertEquals((long) (task1.getDuration() + task2.getDuration()) + task3.getDuration() + task4.getDuration(), (long) task5.getStartTime());
+        assertEquals((long) (task1.getDuration() + task2.getDuration()) + task3.getDuration() + task4.getDuration() + task5.getDuration(), (long) task5.getEndTime());
+
+        Stream.of(task3, task4, task5).forEach(task -> {
+            verify(scoreDirector).beforeVariableChanged(task, "startTime");
+            verify(scoreDirector).afterVariableChanged(task, "startTime");
+            verify(scoreDirector).beforeVariableChanged(task, "endTime");
+            verify(scoreDirector).afterVariableChanged(task, "endTime");
```
triceo: If this fails, I think it's going to be difficult to figure out which task (`task3`, `task4`, `task5`) actually failed verification. (Line number won't tell you that, maybe the exception message will have something useful.)

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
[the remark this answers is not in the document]
triceo: I agree we shouldn't have long-running tests in the core test suite. Optaplanner generally uses the turtle test pattern for that.

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/solver/TaskDifficultyComparator.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Comparator;
+
+import org.apache.commons.lang3.builder.CompareToBuilder;
+import org.jbpm.task.assigning.model.Task;
+
+public class TaskDifficultyComparator implements Comparator<Task>,
+                                                 Serializable {
+
+    @Override
+    public int compare(Task a, Task b) {
```
triceo: One other change we've also done over the last several weeks in the `optaplanner-examples` codebase was to convert all `Comparator`s from `CompareToBuilder` to static `Comparator.comparing(...)`.

It performs better, which will be important in the context of a server app, and also doesn't require a dependency on an external library.

wmedvede:
@rsynek @triceo 

Last comments in and turtle tests applied. Similar pattern, but restricted only for developers, aplied for some tests that does random operations. This last tests can be removed when they are not useful anymore.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AssignTaskProblemFactChangeTest.java
rsynek: Well, as on the line below, there is junit's `assertNotNull`, the only thing we learn when the `internalUser` is null will be information that the line 292 failed as it assumed the object be not null.
On the other hand, if we throw an exception from the stream directly, we can provide additional information that would help with investigating the failure.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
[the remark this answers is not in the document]
rsynek: The reason I asked is that test using this type of termination might be unstable: once it runs in slower environment, 20 seconds might not be sufficient (and also the other way - on a faster computer it may do unnecessary computation).

If we use OptaPlanner in a reproducible mode, each time it will take the same number of steps to reach the same solution. Thus I would rather use step count as a termination condition.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
rsynek: The solution is what @triceo proposed below - using `Assertions.assertThatThrownBy` from the AssertJ library:
https://joel-costigliola.github.io/assertj/

jbpm-task-assigning/jbpm-task-assigning-core/src/main/java/org/jbpm/task/assigning/model/Task.java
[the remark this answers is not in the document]
rsynek: I don't see this comment being addressed. I think the consistency across multiple classes is important.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AbstractProblemFactChangeTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public abstract class AbstractProblemFactChangeTest extends AbstractTaskAssigningCoreTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    final Random random = new Random();
```
rsynek: This `Random` instance seems to be unused.

rsynek:
@wmedvede thanks for incorporating our comments. I put there some new - rather development of the topics we already touched upon in the first round of review.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/PriorityHelperTest.java
wmedvede: suggestion from @triceo  was applied

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/realtime/AbstractProblemFactChangeTest.java
wmedvede: as discussed it's used in subclasses

wmedvede:
Hi @rsynek last modifications we discussed Yesterday are in.

jbpm-task-assigning/jbpm-task-assigning-core/src/test/java/org/jbpm/task/assigning/model/solver/TestSolver.java
```diff
@@ -0,0 +1,97 @@
...
+    public void startSolverAndSolution24Tasks8Users() throws Exception {
+        testSolverStartAndSolution(MILLISECONDS_TIME_SPENT_LIMIT, SET_OF_24TASKS_8USERS_SOLUTION.resource(), 20);
+    }
+
+    /**
+     * Tests that solver for the tasks assigning problem definition can be properly started, a solution can be produced,
+     * and that some minimal constrains are met by de solution.
+     */
+    private void testSolverStartAndSolution(long millisecondsSpentLimit, String solutionResource, int expectedSteps) throws Exception {
+        Solver<TaskAssigningSolution> solver = createNonDaemonSolver(millisecondsSpentLimit);
```
rsynek: Not sure if we understood each other correctly. I meant using step count termination [1] instead of `millisecondsSpentLimit` and adding a timeout just to be sure the test never hangs.

The timeout is there, but the solver is still configured to terminate after a certain period of time.

[1] https://docs.optaplanner.org/latest/optaplanner-docs/html_single/#stepCountTermination

wmedvede:
Hi @rsynek stepCountLimit added.

wmedvede:
Hi @rsynek as commented this PR will be closed in favor of https://github.com/kiegroup/droolsjbpm-integration/pull/1940

PR #1930, opened 2019-10-29
BXMSPROD-560 lock-treatment-tool added to productized profile

Ginxo:
BXMSPROD-560

### DIFF
--- a/process-migration-service/pom.xml
+++ b/process-migration-service/pom.xml
@@ -292,6 +292,7 @@
             </goals>
             <configuration>
               <nodeVersion>${node.version}</nodeVersion>
+              <npmVersion>${npm.version}</npmVersion>
             </configuration>
           </execution>
           <execution>
--- a/process-migration-service/pom.xml
+++ b/process-migration-service/pom.xml
@@ -23,6 +23,7 @@
     <frontend-maven-plugin.version>1.7.6</frontend-maven-plugin.version>
     <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
     <node.version>v11.6.0</node.version>
+    <npm.version>6.9.0</npm.version>
     <!-- Following properties are used only for testing -->
     <!-- This property needs to be overridden when running tests against different version of the WAR (e.g. running
          SNAPSHOT tests against Beta3 WAR) so the tests actually know what exact version of server is being tested -->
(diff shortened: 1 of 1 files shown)


Ginxo:
Same here https://github.com/kiegroup/appformer/pull/828#issuecomment-547385306

MarianMacik:
@Ginxo just a brief explanation why this is done would be helpful :) I guess it is because we want to change the URL of the registry we download dependencies from?

Just FYI this can by overridden by setting `npmRegistryURL` as a Maven property either in a POM or on the command line. It should override the repository used.

Ginxo:
@MarianMacik in one hand we need to assure the library versions consistency, which is done thanks to the lock files (package-lock.json, yarn.lock...) in the other hand we want to use a different registry than the one which is defined in those lock files which is not possible in case the registry is not accessible even specifying the `registry` flag.
Since npm 5 it is possible to remove the `resolved` field from package-lock.json which is exactly what the lock-treatment-tool does.
Let me know if you need more details, I'm glad to clarify whatever doubt you have.

MarianMacik:
@Ginxo ok, so it is just a different approach than to use a specified registry using `npmRegistryURL`, isn't it? Then the registry will be chosen from the `.npmrc` file of the npm installation?

Ginxo:
@MarianMacik yes you are right, but just to be very very precise I would say once the registry is removed from the lock file, the registry is taken from the `--registry` flag or from the local configuration as you pointed out.

MarianMacik:
@Ginxo actually it works for us also in our internal network where we cannot use the public npm - it just overrides even the `package-lock.json` for us, no need to use another tool.

Ginxo:
@MarianMacik good to hear that, anyway it is under the `productized` profile which could safe our life (productization team life) in the future builds in case you change something or even you decide to use yarn instead of npm.

MarianMacik:
@mareknovotny @Ginxo Fine then. My last concern is about consistency. In other repositories like optaweb-employee-rostering the lock-treatment-tool is run with every build, there is no separate profile for that. Is that OK for you?

Ginxo:
@MarianMacik we initially added to the productized profile but one of the reviewers "invited" me to take it out https://github.com/kiegroup/optaweb-employee-rostering/pull/313#issuecomment-548519167

mareknovotny:
@Ginxo so I think I am for having it in "default" build too, are you changing it here?

process-migration-service/pom.xml
```diff
@@ -23,6 +23,7 @@
     <frontend-maven-plugin.version>1.7.6</frontend-maven-plugin.version>
     <maven-resources-plugin.version>3.1.0</maven-resources-plugin.version>
     <node.version>v11.6.0</node.version>
+    <npm.version>6.9.0</npm.version>
```
MarianMacik: For this to have an effect, you have to put `npmVersion` property with the value of this property in the frontend-maven-plugin configuration.

process-migration-service/pom.xml
```diff
@@ -285,21 +286,44 @@
         <executions>
           <execution>
             <id>install node and npm</id>
+            <phase>initialize</phase>
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
             <configuration>
               <nodeVersion>${node.version}</nodeVersion>
```
MarianMacik: More precisely, here.

Ginxo: done, thanks

Ginxo:
@ruromero please review it again, thanks
@mareknovotny ^

PR #1931, opened 2019-10-29
Remove Spring Boot dependency management

cuijulian:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1086

### DIFF
--- a/kie-camel/pom.xml
+++ b/kie-camel/pom.xml
@@ -163,6 +163,10 @@
           <groupId>javax.ws.rs</groupId>
           <artifactId>javax.ws.rs-api</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-jcl</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
 
--- a/kie-camel/pom.xml
+++ b/kie-camel/pom.xml
@@ -115,6 +115,12 @@
       <groupId>org.kie</groupId>
       <artifactId>kie-spring</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-jcl</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
--- a/kie-spring-boot/pom.xml
+++ b/kie-spring-boot/pom.xml
@@ -48,6 +48,12 @@
         <version>${version.org.springframework.boot}</version>
         <type>pom</type>
         <scope>import</scope>
+        <exclusions>
+          <exclusion>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.springframework.boot</groupId>
@@ -57,6 +63,10 @@
           <exclusion>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-api</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
           </exclusion>
         </exclusions>
       </dependency>
--- a/kie-spring-boot/pom.xml
+++ b/kie-spring-boot/pom.xml
@@ -9,8 +9,8 @@
   language governing permissions and ~ limitations under the License. -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
@@ -33,8 +33,6 @@
   </modules>
 
   <properties>
-    <version.org.springframework>5.1.3.RELEASE</version.org.springframework>
-    <version.org.spring-boot>2.1.1.RELEASE</version.org.spring-boot>
     <version.org.apache.cxf>3.2.6</version.org.apache.cxf>
     <version.me.snowdrop.narayana>2.1.0</version.me.snowdrop.narayana>
     <version.org.apache.commons-dbcp2>2.4.0</version.org.apache.commons-dbcp2>
@@ -47,20 +45,20 @@
       <dependency>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-dependencies</artifactId>
-        <version>${version.org.spring-boot}</version>
+        <version>${version.org.springframework.boot}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
       <dependency>
-	<groupId>org.springframework.boot</groupId>
-	<artifactId>spring-boot-starter</artifactId>
-	<version>${version.org.spring-boot}</version>
-	<exclusions>
-	    <exclusion>
-	      <groupId>org.apache.logging.log4j</groupId>
-              <artifactId>log4j-api</artifactId>
-            </exclusion>
-	  </exclusions>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter</artifactId>
+        <version>${version.org.springframework.boot}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
 
       <!-- spring boot auto configuration modules -->
--- a/camel-container-tests/camel-container-tests-module/pom.xml
+++ b/camel-container-tests/camel-container-tests-module/pom.xml
@@ -127,6 +127,12 @@
     <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-spring</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-jcl</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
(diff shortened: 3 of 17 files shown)


MarianMacik:
jenkins retest this

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins retest this

MarianMacik:
jenkins execute springboot build

kie-spring-boot/pom.xml
```diff
@@ -44,24 +42,6 @@
...
-	<groupId>org.springframework.boot</groupId>
-	<artifactId>spring-boot-starter</artifactId>
-	<version>${version.org.spring-boot}</version>
-	<exclusions>
-	    <exclusion>
-	      <groupId>org.apache.logging.log4j</groupId>
-              <artifactId>log4j-api</artifactId>
-            </exclusion>
-	  </exclusions>
-      </dependency>
```
MarianMacik: I am a little bit worried about this exclusion which is not materialized here currently. Maybe we should leave it here as it was.

cuijulian: Good point.

cuijulian:
jenkins retest this

MarianMacik:
@cuijulian So I checked why the build fails on Appformer repository but I haven't found anything. For some reason we have a compilation error in the uberfire-structure-backend module:

`[ERROR] /home/jenkins/workspace/KIE/master/pullrequest/droolsjbpm-integration-springboot-pullrequests/upstream-repos/appformer/uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomJsonReaderDefault.java:[38,13] cannot find symbol`

Appformer is built after `droolsjbpm-build-bootstrap`, so there is some influence from the `droolsjbpm-build-bootstrap` on the Appformer repo.

cuijulian:
> @cuijulian So I checked why the build fails on Appformer repository but I haven't found anything. For some reason we have a compilation error in the uberfire-structure-backend module:
> 
> `[ERROR] /home/jenkins/workspace/KIE/master/pullrequest/droolsjbpm-integration-springboot-pullrequests/upstream-repos/appformer/uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomJsonReaderDefault.java:[38,13] cannot find symbol`
> 
> Appformer is built after `droolsjbpm-build-bootstrap`, so there is some influence from the `droolsjbpm-build-bootstrap` on the Appformer repo.

@MarianMacik I can take a look at the Appformer repo. This is strange; even though Appformer inherits from `kie-parent` it doesn't use any Spring dependencies. I suspect the newer version of Spring Boot overrides other dependency versions in Appformer.

cuijulian:
jenkins retest this

cuijulian:

cuijulian:
@MarianMacik I resolved the Appformer compilation issue (it was what I suspected, adding Spring Boot 2.2 to `kie-parent` introduced transitive dependencies that overrode required dependencies in other repos, such as Appformer). But the downstream build exposed yet another problem: a duplicate dependency class in `kie-camel`. I will push a fix tomorrow.

MarianMacik:
@cuijulian Thank you. I thought that `javax.json.JsonObject` is a part of JDK, but it is from Java EE. Now it makes sense why it failed. In case you have any issues excluding `kie-camel` duplicate classes, ping me and I will look at it as well.

cuijulian:
@MarianMacik Sorry, one correction. The duplicate classes were found in `droolsjbpm-bpms-distribution` and `droolsjbpm-brms-distribution` according to the Jenkins build:

`06:05:09 Duplicate classes found:
06:05:09 
06:05:09   Found in:
06:05:09     org.slf4j:jcl-over-slf4j:jar:1.7.26:compile
06:05:09     org.springframework:spring-jcl:jar:5.2.0.RELEASE:compile
06:05:09   Duplicate classes:
06:05:09     org/apache/commons/logging/Log.class
06:05:09     org/apache/commons/logging/impl/NoOpLog.class
06:05:09     org/apache/commons/logging/impl/SimpleLog.class
06:05:09     org/apache/commons/logging/LogFactory.class`

However, the Jenkins build result is inconsistent with my local build, which works fine. The [POM](https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-spring/pom.xml#L144) in `kie-spring` clearly excludes the commons-logging dependency which should solve this issue, but for some reason the build still fails. Has this happened before, and what should I do?

MarianMacik:
@cuijulian In fact, duplicated classes were found in these modules:

* kie-camel
* kie-spring-test
* jbpm-spring-boot-starter-basic
* kie-server-spring-boot-autoconfiguration
* droolsjbpm-bpms-distribution
* droolsjbpm-brms-distribution

So you were 50% right :) When you look at the console output, at the end there are all failures listed after the `BUILD FAILURE`. The reason why it's better to look at all the failures at the bottom of the build log is that the Maven build is run using the `-fae (fail at end)` option which will continue building modules even after the failure. That's the reason why not only `BPMS distribution` and `BRMS distribution` have `FAILURE` listed but others as well.

With kie-camel I had to exclude the dependency also from camel-cxf https://github.com/kiegroup/droolsjbpm-integration/pull/1931/files#diff-d6743a971a2d28ef7caa84bf53ee4139R166 even though from the Maven perspective it is not needed - dependency tree is without the spring-jcl artifact even without that exclusion. But the enforcer plugin was still showing issues - probably there is either a bug in the enforcer plugin or OSGI bundles cause it troubles for some reason. I have also included some JakartaEE artifacts even though the enforcer didn't find it but when I opened them they contained classes already present on our classpath in JavaEE packages.

And regarding different results on a local machine and on Jenkins, did you try to run your local build with the `-Dfull` option? Maybe that caused the discrepancy.

MarianMacik:
jenkins retest this

MarianMacik:
jenkins retest this

cuijulian:
jenkins retest this

cuijulian:
jenkins execute springboot build

MarianMacik:
@mareknovotny 

```
The following artifacts could not be resolved:
io.netty:netty-buffer:jar:3.10.6.Final,
io.netty:netty-transport:jar:3.10.6.Final,
io.netty:netty-handler:jar:3.10.6.Final,
io.netty:netty-transport-native-epoll:jar:linux-x86_64:3.10.6.Final,
io.netty:netty-transport-native-kqueue:jar:osx-x86_64:3.10.6.Final,
io.netty:netty-codec-http:jar:3.10.6.Final: Could not find artifact io.netty:netty-buffer:jar:3.10.6.Final in mirror-central
```
I guess it is connected with the change in the droolsjbpm-build-bootstrap PR which changed the version of netty?

mareknovotny:
@MarianMacik yes, this is not ok. I already added to your comment there that this is not ok

cuijulian:

cuijulian:
jenkins retest this

cuijulian:
jenkins execute springboot build

MarianMacik:
Spring Boot job failed to compile because of kiegroup/kie-jenkins-scripts#638. It should be fixed now.

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins execute springboot build

MarianMacik:
jenkins retest this

MarianMacik:
jenkins retest this

cuijulian:
jenkins retest this

cuijulian:
jenkins execute springboot build

cuijulian:
jenkins execute springboot build

MarianMacik:
@cuijulian Woohoo! Everything is green!

kie-camel/pom.xml
```diff
@@ -114,7 +114,6 @@
     <dependency>
       <groupId>org.kie</groupId>
       <artifactId>kie-spring</artifactId>
-      <scope>test</scope>
```
MarianMacik: I tried `mvn clean install -DskipTests -Dfull` with the test scope and it works. Can you please put it back to test? Otherwise we would have many artifacts on compile path which are not necessary.

cuijulian: @MarianMacik It doesn't work for me. When I run `mvn clean install -DskipTests -Dfull` with the test scope the `maven-enforcer-plugin` complains about duplicate classes caused by `spring-jcl`. This happens even though I clearly exclude that dependency from `kie-spring`. Changing the scope back to compile fixes this. Could this be a bug in the enforcer in test scope?

Also, does it work when you run `mvn clean install -Dfull`? Don't we need to run tests to verify before merging?

cuijulian: I'll try adding the test scope back and see what Jenkins says. Perhaps it's an issue with my local machine.

cuijulian:
jenkins retest this

cuijulian:
jenkins execute springboot build

MarianMacik:
@cuijulian As I said here: https://github.com/kiegroup/droolsjbpm-integration/pull/1931#issuecomment-552117910 

> With kie-camel I had to exclude the dependency also from camel-cxf https://github.com/kiegroup/droolsjbpm-integration/pull/1931/files#diff-d6743a971a2d28ef7caa84bf53ee4139R166 even though from the Maven perspective it is not needed - dependency tree is without the spring-jcl artifact even without that exclusion. But the enforcer plugin was still showing issues - probably there is either a bug in the enforcer plugin or OSGI bundles cause it troubles for some reason.

So I just did it again. I think the issue is that because kie-spring is test dependency, enforcer somehow excludes just the test scope of jcl? And then will find again another jcl dependency coming from camel-cxf, this time with a normal compile scope. But dependency tree of course doesn't show any jcl dependency coming from camel-cxf because it has already been pulled from kie-spring.

The other option which worked was to declare kie-spring test dependency after camel-cxf so camel-cxf pulls jcl and excludes it and enforcer was ok with that.

So apparently enforcer works differently than Maven resolution does and it seems to only respect exclusions from non-test dependencies? But at the same time this works with other projects without an issue, so maybe this is a corner case. I even tried latest Maven Enforcer plugin (3.0.0-M3 version) and it didn't help. Now it should be OK.

@mareknovotny just FYI ^^^ that Enforcer plugin may be not that reliable as we thought.

MarianMacik:
jenkins execute springboot build

MarianMacik:
test failure not related - failing also for other PRs

cuijulian:
@MarianMacik could you take a look at this PR when you get a chance: https://github.com/kiegroup/jbpm-work-items/pull/114#issuecomment-558209008

MarianMacik:
jenkins retest this

MarianMacik:
@cuijulian Green!

cuijulian:
> @cuijulian Green!

Thanks a lot for your help! Fingers crossed that the full downstream build passes.

PR #1934, opened 2019-11-04
[JBPM-8885] Nodes after multi-Instance subprocess are not blurred eventhough process instance is completed.

elguardian:
Added a property to increase the number of nodes retrieved from database

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
@@ -41,10 +41,18 @@ import org.slf4j.LoggerFactory;
 import static org.jbpm.process.svg.processor.SVGProcessor.ACTIVE_BORDER_COLOR;
 import static org.jbpm.process.svg.processor.SVGProcessor.COMPLETED_BORDER_COLOR;
 import static org.jbpm.process.svg.processor.SVGProcessor.COMPLETED_COLOR;
+import static org.kie.server.api.KieServerConstants.KIE_SERVER_IMAGESERVICE_MAX_NODES;
 
 public class ImageServiceBase {
 
     private static final Logger logger = LoggerFactory.getLogger(ImageServiceBase.class);
+
+    /**
+     * This causes the image service to limit the number of nodes (performance). 
+     * Due to this limitation it could cause a known issue not blurring all the nodes active or completed depending on
+     * the process size.
+     */
+    private static final int MAX_NODES = Integer.parseInt(System.getProperty(KIE_SERVER_IMAGESERVICE_MAX_NODES, "1000"));
 
     private RuntimeDataService dataService;
     private Map<String, ImageReference> imageReferenceMap;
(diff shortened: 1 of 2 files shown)


elguardian:
Reuse of queries cannot happens as it will have side effects in other places of the code
Cannot create custom queries for retrieve just the ids set of the nodes (the image process also depends on the identity column)
just left with a custom property and increase the number so I don't hurt too much the performance.

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
```diff
@@ -41,11 +41,19 @@
...
 public class ImageServiceBase {
 
     private static final Logger logger = LoggerFactory.getLogger(ImageServiceBase.class);
 
+    /**
+     * This causes the image service to limit the number of nodes (performance). 
+     * Due to this limitation it could cause a known issue not blurring all the nodes active or completed depending on
+     * the process size.
+     */
+    private static final int MAX_NODES = Integer.parseInt(System.getProperty(KIE_SERVER_IMAGESERVICE_MAX_NODES, "10000"));
```
MarianMacik: I would make the default `1000` as it was before. Just for the sake of backward compatibility.

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
```diff
@@ -121,8 +129,8 @@ public String getActiveProcessImage(String containerId, long procInstId, String
         if (imageSVG != null) {
             // find active nodes and modify image
             Map<String, String> subProcessLinks = new HashMap<>();
-            Collection<NodeInstanceDesc> activeLogs = dataService.getProcessInstanceHistoryActive(procInstId, new QueryContext(0, 1000));
-            Collection<NodeInstanceDesc> completedLogs = dataService.getProcessInstanceHistoryCompleted(procInstId, new QueryContext(0, 1000));
+            Collection<NodeInstanceDesc> activeLogs = dataService.getProcessInstanceHistoryActive(procInstId, new QueryContext(0, MAX_NODES));
+            Collection<NodeInstanceDesc> completedLogs = dataService.getProcessInstanceHistoryCompleted(procInstId, new QueryContext(0, MAX_NODES));
```
MarianMacik: Did you consider using paging for that?

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
MarianMacik: OK, if we take it as a fix, I see no problem. But the behaviour changes slightly now :)

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
MarianMacik: Can you please elaborate on this?
> yep but if the limit (X) there is for performance reason it does not make much sense to retrieve in chunks. you do it all in one go (just need to pass query context as null) .

Not sure I got it.

But yes, we have to fetch all data at once, I was mistaken by the fact that we are doing some checks for it or we are showing progress, but neither of that is true. :)

elguardian: I don't want to remove the limit (there must be a reason), but I will allow to retrieve everything if the parameter is 0 (infinity).

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
elguardian: I meant I will change it.

MarianMacik: OK then. Thanks.

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/ImageServiceBase.java
MarianMacik: Thanks, that makes sense.

MarianMacik:
@mswiderski any comments?

PR #1945, opened 2019-11-21
JBPM-8781 - API to view all of sub-cases belong to a parent case - fix ambiguous column

MarianMacik:
Just a test commit to execute just KIE Server tests.


MarianMacik:
jenkins retest this

PR #1950, opened 2019-11-27
[JBPM-8864] not able to delete kieContainer after deploying container several times with same containerID due to an error

elguardian:
the system registers the container even before the container was created causing problem during unregistering (cannot dispose because the container is null in the registered instance so it cannot 
dispose in kieservice container map. The solution is to delay till we have the container active and then
overwrite the container.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/KieServerImpl.java
@@ -31,6 +31,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
+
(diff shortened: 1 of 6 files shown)


sutaakar:
@elguardian Can you please add a test reproducing the issue?

elguardian:
done @sutaakar

sutaakar:
@elguardian Please remove the unrelated package-lock.json changes.
Also the test failures seem to be related to these changes.

elguardian:
Saw it... some weak tests (race condition) and a change of behaviour in the list of containers when the container is failing during deployment

elguardian:
fixing test case as missing modules in maven repository was considered  equal to deployment failure
Jenkins retest this please

elguardian:
Jenkins retest this please

elguardian:
@sutaakar thx for the hint, I was wasting a lot of time trying to make everything work in the current project. can you approve ?

elguardian:
@kiegroup/gatekeepers 
can anyone merge this ?

sutaakar:
SonarCloud failures are acceptable, coverage is calculated also from changes in test classes related just for local testing.

PR #1951, opened 2019-12-02
[JBPM-8864] Move KieServerContainerListFilteringIntegrationTest to the jbpm test …

sutaakar:
…module

KieServerContainerListFilteringIntegrationTest needs to be adjusted to fail while initializing extension.
Previous approach with providing non existing kjar version cannot be used any more as latest fixes will differentiate between deployment failure caused by missing artifacts in Maven repository (container won't be deployed) and extension failures (container will be deployed with failed state).

@elguardian This PR should fix test issues related to https://github.com/kiegroup/droolsjbpm-integration/pull/1950

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/KieServerContainerListFilteringIntegrationTest.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/KieServerContainerListFilteringIntegrationTest.java
@@ -13,12 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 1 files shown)


elguardian:
that is already in the PR 
https://github.com/kiegroup/droolsjbpm-integration/pull/1950/files#diff-0e79059cbd4485c2b2f0c0d373c3fac6R81
It was working within eclipse but for some reason failing outside. :S I will check this

sutaakar:
It doesn't work in common test module as the jBPM extension is disabled by default there, see https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/pom.xml#L314

elguardian:
I will moved it ir requires less changes in the code to make it work, 
I need to change some of the tests anyway.

elguardian:
did you approach.

PR #1952, opened 2019-12-03
[JBPM-8922] RemoteBusinessRuleTaskHandler does not work with https url

elguardian:
force https detection during load balancing.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/util/SSLUtilities.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/util/SSLUtilities.java
@@ -1,0 +1,59 @@
+package org.kie.server.client.util;
+
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSession;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+
+public final class SSLUtilities {
+
+    private static HostnameVerifier verifier;
+
+    private static TrustManager[] trustManagers;
+
+    public static void trustAllHostnames() {
+        if (verifier == null) {
+            verifier = new AllValidHostnameVerifier();
+        }
+        HttpsURLConnection.setDefaultHostnameVerifier(verifier);
+    }
+
+    public static void trustAllHttpsCertificates() {
+        SSLContext context;
+
+        if (trustManagers == null) {
+            trustManagers = new TrustManager[]{new AllValidTrustManager()};
+        }
+        try {
+            context = SSLContext.getInstance("SSL");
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/ForceDetectUseSSLTest.java
```diff
@@ -0,0 +1,26 @@
+package org.kie.server.client;
```
mswiderski: missing license header

kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/util/SSLUtilities.java
```diff
@@ -0,0 +1,59 @@
+package org.kie.server.client.util;
```
mswiderski: missing license header

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesConfigurationImpl.java
```diff
@@ -121,10 +121,24 @@ public KieServicesConfigurationImpl(String url, CredentialsProvider credentialsP
...
     public KieServicesConfigurationImpl(String url, CredentialsProvider credentialsProvider, long timeout) {
         this.transport = Transport.REST;
-
         this.serverUrl = url;
         this.timeoutInMillisecs = timeout;
         this.credentialsProvider = credentialsProvider;
+        //        this.useSsl = autoDetectUseSSL(url);
+    }
+
+    public boolean autoDetectUseSSL(String url) {
```
mswiderski: not sure I follow this method... as the useSsl seems to be used by JMS layer and now we just check the http endpoints... 

https://github.com/kiegroup/droolsjbpm-integration/blob/bc0380c7785501b53943a81f5421d10a4ecc1f3d/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/KieServicesConfigurationImpl.java#L63

could you please explain a bit what are we trying to do here?

elguardian:
sorry about the confusion. The problem is more how the system should behave (i messed up with the PR). The test case relevant should be https://github.com/kiegroup/droolsjbpm-integration/pull/1952/files#diff-3794651273729e747889499f86dea895R18

This test fails during kie services client constructor because the init of the server client requires at least one endpoint to be available. 

In the jira reported the https is failing because the truststore is not configured (I created a test case regarding load balancing that prooves that it works with http and https) but what about if you try to deploy while there is no endpoint available (real endpoint down) ? Not really sure if this needs to fail or not). The jira of this is not really a bug but a missconfiguration so not really sure if this was on purpose or not. @mswiderski  wdyt ?

mswiderski:
@elguardian in general the endpoint should be available, at least one at the time of creation of kie server client. So if that is not available I'd say it's really an edge case and the kie server client itself can't do much about it.
Personally I prefer to fail fast when there are no endpoints available as it could be temporarily or it can be as simple as typo and thus way more severe.

PR #1954, opened 2019-12-04
[JBPM-8931] Form Retrieval not honouring bypass auth user setting

fsou:
When using bypass auth user, FormService now can use the "user" query parameter to check permissions to see task instead of using the authenticated user.

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
@@ -124,6 +124,41 @@ public interface UIServicesClient {
     String getTaskFormByType(String containerId, Long taskId, String formType);
 
     /**
+     * Returns task form for given task id that belongs to given container as specified user. It returns default form type
+     * which is (FORM - build with form modeler). If there is a need to select the type use #getProcessFormForUserByType
+     * Introduced to allow client to take advantage of bypass auth user connecting to server.
+     * @param userId userId making the request to bypass auth user
+     * @param containerId container identifier where task resides
+     * @param taskId unique task id
+     * @param language language that form should be filtered for
+     * @return  string representation (json or xml depending on client marshaling selection) of the task form
+     */
+    String getTaskFormAsUser(String userId, String containerId, Long taskId, String language);
+
+    /**
+     * Returns task form for given task id that belongs to given container as specified user.
+     * Introduced to allow client to take advantage of bypass auth user connecting to server.
+     * @param userId userId making the request to bypass auth user
+     * @param containerId container identifier where task resides
+     * @param taskId unique task id
+     * @param language language that form should be filtered for
+     * @param formType type of form to be returned (FORM - default (form modeler), FRM - v7 forms, FTL - freemarker template)
+     * @return  string representation (json or xml depending on client marshaling selection) of the task form
+     */
+    String getTaskFormByTypeAsUser(String userId, String containerId, Long taskId, String language, String formType);
+
+    /**
+     * Returns task form for given task id that belongs to given container as specified user as raw content - without filtering values by language.
+     * It returns default form type which is (FORM - build with form modeler). If there is a need to select the type use #getProcessFormByType
+     * Introduced to allow client to take advantage of bypass auth user connecting to server.
+     * @param userId userId making the request to bypass auth user
+     * @param containerId container identifier where task resides
+     * @param taskId unique task id
+     * @return  string representation of the task form without any marshalling
+     */
+    String getTaskRawFormAsUser(String userId, String containerId, Long taskId);
+
+    /**
      * Returns process image (svg) of the given process id that belongs to given container
      * @param containerId container identifier where process resides
      * @param processId  unique process id
--- a/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
+++ b/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
@@ -127,36 +127,36 @@ public interface UIServicesClient {
      * Returns task form for given task id that belongs to given container as specified user. It returns default form type
      * which is (FORM - build with form modeler). If there is a need to select the type use #getProcessFormForUserByType
      * Introduced to allow client to take advantage of bypass auth user connecting to server.
-     * @param userId userId making the request to bypass auth user
      * @param containerId container identifier where task resides
      * @param taskId unique task id
      * @param language language that form should be filtered for
+     * @param userId userId making the request to bypass auth user
      * @return  string representation (json or xml depending on client marshaling selection) of the task form
      */
-    String getTaskFormAsUser(String userId, String containerId, Long taskId, String language);
+    String getTaskFormAsUser(String containerId, Long taskId, String language, String userId);
 
     /**
      * Returns task form for given task id that belongs to given container as specified user.
      * Introduced to allow client to take advantage of bypass auth user connecting to server.
-     * @param userId userId making the request to bypass auth user
      * @param containerId container identifier where task resides
      * @param taskId unique task id
      * @param language language that form should be filtered for
      * @param formType type of form to be returned (FORM - default (form modeler), FRM - v7 forms, FTL - freemarker template)
+     * @param userId userId making the request to bypass auth user
      * @return  string representation (json or xml depending on client marshaling selection) of the task form
      */
-    String getTaskFormByTypeAsUser(String userId, String containerId, Long taskId, String language, String formType);
+    String getTaskFormByTypeAsUser(String containerId, Long taskId, String language, String formType, String userId);
 
     /**
      * Returns task form for given task id that belongs to given container as specified user as raw content - without filtering values by language.
      * It returns default form type which is (FORM - build with form modeler). If there is a need to select the type use #getProcessFormByType
      * Introduced to allow client to take advantage of bypass auth user connecting to server.
-     * @param userId userId making the request to bypass auth user
      * @param containerId container identifier where task resides
      * @param taskId unique task id
+     * @param userId userId making the request to bypass auth user
      * @return  string representation of the task form without any marshalling
      */
-    String getTaskRawFormAsUser(String userId, String containerId, Long taskId);
+    String getTaskRawFormAsUser(String containerId, Long taskId, String userId);
 
     /**
      * Returns process image (svg) of the given process id that belongs to given container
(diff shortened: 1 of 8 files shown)


mswiderski:
ok to test

fsou:
I'll try to do that.

sutaakar:
Please add the user parameter also to Kie server client for JMS - https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java#L137

That should fix the failing tests.

fsou:
Thank you @sutaakar .

Will do that right now.

I am still struggling to make the new tests, I am taking this as starting point
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-all/src/test/java/org/kie/server/integrationtests/jbpm/rest/FormServiceRestOnlyIntegrationTest.java

Please let me know if I am in the wrong path.

FS

sutaakar:
Reusing these integration tests for checking auth bypass could be difficult.
The problem is that bypass needs to configure system property for Kie server, as currently it is not set in integration tests.

Seems like the only option to test it would be by unit test for FormServiceBase

fsou:
@sutaakar  thank you for your input.

I am making the change you've recommended previously.
It's making me change the whole UIServicesClient interface to add the new parameter userId.
I am taking a look at UserTaskServicesClient to replicate the behaviour.

Again, I would appreciate your oppinion if I am on the right track.
Thank you,
FS

sutaakar:
It would be better to create a new method in client  with this userId rather than modifying old one. This way there will be no backward incompatible change.

fsou:
Ok, I'll add three new methods that will add the extra parameter:
* getTaskForm
* getTaskFormByType
* getTaskRawForm

I'll work on this and add unit testing.

Thank you again.

fsou:
Added tests to FormServiceBase and added changes suggested by @sutaakar

fsou:
Well, how to you use to handle this kind of error, when the ammount of parameters gets greater than 7?
I don't have access to the build details to see what else is going on with testing.
I am finishing tests for UIServicesClient changes.

sutaakar:
About amount of parameters - it is a general rule in SonarCloud, in this case there is no other way. You can ignore that SonarCloud failure for now.

sutaakar:
About those test failures, they are caused by current differences between number of arguments in Kie server JMS client and getFormDisplayTask method in FormServiceBase.
As you can see from https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java#L137 the Kie server JMS client references getFormDisplayTask method in FormServiceBase. It is followed by list of arguments supplied to that method. Since you added userId parameter in that method then you also need to adjust the JMS client and add there the parameter too.

You can check that your implementation works by running `mvn clean install -Pwildfly -Dit.test=FormServiceIntegrationTest` from kie-server-integ-tests-all Maven module (don't forget to build the kie-server first as the test use built Kie server war file).

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/test/java/org/kie/server/services/jbpm/ui/FormServiceBaseTest.java
```diff
@@ -0,0 +1,140 @@
+package org.kie.server.services.jbpm.ui;
+
+/*
```
sutaakar: Please move the licence header declaration before package definition.

fsou: Done!

fsou:
Added tests, fixed the JMS Command construction to pass the userId.

kie-server-parent/kie-server-remote/kie-server-client/src/test/java/org/kie/server/client/UIServicesClientTest.java
```diff
@@ -0,0 +1,126 @@
...
+public class UIServicesClientTest extends BaseKieServicesClientTest {
+
+    private final String dummyFormContent = "form content";
+
+    private UIServicesClient uiClient;
+
+    @Before
+    public void createClient() {
+
+        config.setCapabilities(List.of(KieServerConstants.CAPABILITY_BPM_UI));
```
sutaakar: Our code needs to be compatible with Java 8. List.of() was brought in Java 9. Can you please refactor the code to be Java 8 compatible?

fsou: Of course! I wasn't aware of the Java 8 compatibility, sorry.
Will to that right away.

fsou: Converted to Arrays.asList on last commit.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
```diff
@@ -123,6 +123,41 @@
...
+     * Returns task form for given task id that belongs to given container as specified user. It returns default form type
+     * which is (FORM - build with form modeler). If there is a need to select the type use #getProcessFormForUserByType
+     * Introduced to allow client to take advantage of bypass auth user connecting to server.
+     * @param userId userId making the request to bypass auth user
+     * @param containerId container identifier where task resides
+     * @param taskId unique task id
+     * @param language language that form should be filtered for
+     * @return  string representation (json or xml depending on client marshaling selection) of the task form
+     */
+    String getTaskFormAsUser(String userId, String containerId, Long taskId, String language);
```
sutaakar: Can you please move the userId parameter, to make him for example last one (similar as is done for example in UserTaskServicesClient)? Our convention here is to keep containerId as first parameter.

fsou: Ok, I'll do that change right away.
Could you take a look at test results to see what's breaking them?

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java
```diff
@@ -107,18 +107,34 @@ public String getTaskForm(String containerId, Long taskId, String language) {
...
+
+    private String getTaskFormByTypeAsUser(String userId, String containerId, Long taskId, String language, String formType, boolean marshallContent ) {
         if( config.isRest() ) {
-            Map<String, Object> valuesMap = new HashMap<String, Object>();
+            Map<String, Object> valuesMap = new HashMap<>();
             valuesMap.put(RestURI.CONTAINER_ID, containerId);
             valuesMap.put(RestURI.TASK_INSTANCE_ID, taskId);
 
-            StringBuffer params = new StringBuffer();
+            String userQuery = userId.equals("") ? "" : getUserQueryStr(userId);
```
sutaakar: This code can throw NPE when null userId is supplied. 
It would be good to replace userId default value in this class from "" to null, here you can check null value using StringUtils.isEmpty .

fsou: Will do that too.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
sutaakar: there is just one test failure, seems to be an unstable test. Lets see what will be the result once you implement the changes and job will retrigger.

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/impl/UIServicesClientImpl.java
```diff
@@ -130,12 +146,14 @@ private String getTaskFormByType( String containerId, Long taskId, String langua
...
-                    build(loadBalancer.getUrl(), FORM_URI + "/" + TASK_FORM_GET_URI, valuesMap) + "?" + params.toString());
+                    build(loadBalancer.getUrl(), FORM_URI + "/" + TASK_FORM_GET_URI, valuesMap) + params.toString());
 
         } else {
-            CommandScript script = new CommandScript( Collections.singletonList(
-                    (KieServerCommand) new DescriptorCommand( "FormService", "getFormDisplayTask", new Object[]{containerId, taskId, StringUtils.defaultString( language ), !StringUtils.isEmpty( language ), formType } )) );
-            ServiceResponse<String> response = (ServiceResponse<String>) executeJmsCommand( script, DescriptorCommand.class.getName(), "BPM-UI", containerId ).getResponses().get(0);
+            CommandScript script = new CommandScript(Collections.singletonList(
+                    (KieServerCommand) new DescriptorCommand("FormService", "getFormDisplayTask",
+                            containerId, taskId, userId, StringUtils.defaultString(language), !StringUtils.isEmpty(language), formType)));
```
fsou: @sutaakar  Sonar analysis advised to remove Object[]. I was making local testing but it went with this commit. Please advise if you see any problem with this change or if I should get back the Object[].

sutaakar: Well, I would say you can keep there the Object[] to make implementation consistent with other methods and ignore this SonarCloud warning.

fsou: Ok, will revert that change.

sutaakar:
Hmm, build failed due to some issue with Jenkins, will retrigger it

sutaakar:
Jenkins retest this

sutaakar:
The test failures are related to the fact that marshalling cannot correctly process null values as parameters. For REST it is handled by checking if the userId is empty, however for JMS it isn't. Can you please adjust implementation making sure that if null userId is passed then empty String is used for JMS?

Sorry for this, I didn't realize this issue before.

fsou:
Hi @sutaakar .
I'll make that change, no problem at all.
I am trying to increase code coverage also.

fsou:
@sutaakar 
Can you review the last changes, please?
Now all checks have passed.

sutaakar:
Looks good to me, thanks.
@mswiderski Can you please take a look?

sutaakar:
The verification job seems stuck, retriggering.

sutaakar:
Jenkins retest this

fsou:
> The verification job seems stuck, retriggering.

It seems something is wrong gathering data from jenkis to here, as almost all open PR are with the yellow dot and this changes we're already tested along with the SonarCloud analysis.

Let's see how it goes this time...

PR #1955, opened 2019-12-04
Use Nexus cache server to speed up Kie server integration tests

sutaakar:


sutaakar:
Jenkins retest this

sutaakar:
Seems like using cache server doesn't have any significant effect on Kie server test execution time in Jenkins PR builder.
The container build time for common containers is same with cache server and without. 
The difference can be spotted when complex container is built. For example in OptaplannerIntegrationTest the cache server saved roughly 7 seconds on container installation time.

The only significant possible advantage of using cache server is from my point of view isolation of our test execution from default Maven repository.

PR #1956, opened 2019-12-04
JBPM-8934 - PIM Service cannot create a database schema on Oracle dat…

MarianMacik:
…abase


ruromero:
@MarianMacik I prefer to give it a different name rather than escape it with quotes

MarianMacik:
@ruromero why? Delimiting is a relevant approach, I have already tested it on all supported databases. Moreover, a different column name would break the backward compatibility.

MarianMacik:
@ruromero I renamed the column to `success`. Let me know what you think about it.

PR #1957, opened 2019-12-04
PLANNER-1739: Quick fix for OptaPlanner test broken by PLANNER-241

yurloc:
OptaPlanner no longer throws exception if the entity collection is empty. That caused the test to fail because no exception was thrown. But it still throws exception if the value provider returns an empty collection. We can exploit this behavior to keep the negative test.

I personally don't see much value in this test but it is possible that it covers exception propagation in the Fuse environment. Jiri is on a leave this week so we cannot ask him so I'm proposing a fix that keeps the test in place and makes it pass. We can discuss with Jiri and decide to remove the test later.


ge0ffrey:
Thanks Jiri for fixing my mistake :)

yurloc:
I don't think it was a mistake. We don't know all tests in optaplanner's downstream and we don't run them in the PR check so the failing daily build was the only way to find out.

PR #1962, opened 2019-12-09
JBPM-8936 ConcurrentModificationException when retrieving server temp…

bxf12315:
…late

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
@@ -18,14 +18,17 @@ package org.kie.server.controller.api.model.spec;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
+import java.util.Collections;
+import java.util.EnumMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
 
 import org.kie.server.api.model.KieServerMode;
 import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
@@ -37,13 +40,16 @@ public class ServerTemplate extends ServerTemplateKey {
     @XmlElement(name = "container-specs")
     private Collection<ContainerSpec> containersSpec = new ArrayList<ContainerSpec>();
     @XmlElement(name = "server-config")
-    private Map<Capability, ServerConfig> configs = new HashMap<Capability, ServerConfig>();
+    private Map<Capability, ServerConfig> configs = new EnumMap<>(Capability.class);
     @XmlElement(name = "server-instances")
     private Collection<ServerInstanceKey> serverInstances = new ArrayList<ServerInstanceKey>();
     @XmlElement(name="capabilities")
     private List<String> capabilities = new ArrayList<String>();
     @XmlElement(name="mode")
     private KieServerMode mode;
+
+    @XmlTransient
+    private final Object lock = new Object();
 
     public ServerTemplate() {
     }
@@ -76,9 +82,9 @@ public class ServerTemplate extends ServerTemplateKey {
 
     public Map<Capability, ServerConfig> getConfigs() {
         if (configs == null) {
-            configs = new HashMap<Capability, ServerConfig>();
+            configs = new EnumMap<>(Capability.class);
         }
-        return new HashMap<Capability, ServerConfig>( configs );
+        return new EnumMap<>(configs);
     }
 
     public Collection<ContainerSpec> getContainersSpec() {
@@ -176,13 +182,8 @@ public class ServerTemplate extends ServerTemplateKey {
         if (serverInstances == null) {
             return;
         }
-        Iterator<ServerInstanceKey> iterator = serverInstances.iterator();
-
-        while(iterator.hasNext()) {
-            ServerInstanceKey serverInstanceKey = iterator.next();
-            if (serverInstanceId.equals(serverInstanceKey.getServerInstanceId())) {
-                iterator.remove();
-            }
+        synchronized (lock) {
+            serverInstances.removeIf((ServerInstanceKey key) -> key.getServerInstanceId().equals(serverInstanceId));
         }
     }
 
@@ -240,5 +241,17 @@ public class ServerTemplate extends ServerTemplateKey {
         int result = super.hashCode();
         result = 31 * result + (capabilities != null ? capabilities.hashCode() : 0);
         return result;
+    }
+
+    public ServerTemplate cloneServerTemplate() {
+        synchronized (lock) {
+            ServerTemplate serverTemplate = new ServerTemplate(this.getId(), this.getName(), new ArrayList<>(capabilities),
+                                                               getConfigs(), Collections.EMPTY_LIST, new ArrayList<ServerInstanceKey>(serverInstances));
+            containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), new EnumMap<>(c.getConfigs()))));
+
+            serverTemplate.setMode(mode);
+
+            return serverTemplate;
+        }
     }
 }
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.controller.api.model.spec;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Test;
+import org.kie.server.api.model.KieContainerStatus;
+import org.kie.server.api.model.KieScannerStatus;
+import org.kie.server.api.model.KieServerMode;
+import org.kie.server.api.model.ReleaseId;
+import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+public class ServerTemplateTest {
+
+    @Test
+    public void testClone() {
+        ServerTemplate src = new ServerTemplate();
+        src.setId("test1");
+        src.setName("test1");
+        src.setMode(KieServerMode.DEVELOPMENT);
+        List<String> capabilities = new ArrayList<String>();
+        capabilities.add(Capability.RULE.toString());
+        src.setCapabilities(capabilities);
+        Map<Capability, ContainerConfig> containerSpecRule = new HashMap<>();
+        containerSpecRule.put(Capability.RULE, new RuleConfig(1L, KieScannerStatus.STARTED));
+        src.addContainerSpec(new ContainerSpec("test", "test1", new ServerTemplateKey("test", "test"),
+                                                new ReleaseId("test", "test", "1"), KieContainerStatus.STARTED, containerSpecRule));
+
+        Map<Capability, ContainerConfig> containerSpecProcess = new HashMap<>();
+        containerSpecProcess.put(Capability.PROCESS, new ProcessConfig("test1", "test1", "test1", "test1"));
+        src.addContainerSpec(new ContainerSpec("test1", "test1", new ServerTemplateKey("test1", "test1"),
+                                               new ReleaseId("test1", "test1", "2"), KieContainerStatus.STARTED, containerSpecProcess));
+
+        Map<Capability, ServerConfig> cloneServerConfig = new HashMap<Capability, ServerConfig>();
+        cloneServerConfig.put(Capability.RULE,new ServerConfig());
+        src.setConfigs(cloneServerConfig);
+        src.addServerInstance(new ServerInstanceKey("test1", "test1", "test1", "test1"));
+
+        ServerTemplate dest = src.cloneServerTemplate();
+
+        assertEquals(src, dest);
+        assertFalse(src == dest);
+
+    }
+}
--- a/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
@@ -212,9 +212,10 @@ public class ControllerUtils {
                 break;
             }
         }
-
+        if (entity instanceof ServerTemplate) {
+            return marshaller.marshall(((ServerTemplate) entity).cloneServerTemplate());
+        }
         return marshaller.marshall(entity);
-
     }
 
     public static String getUser() {
--- a/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.server.controller.rest;
+
+import java.util.stream.IntStream;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.controller.api.storage.KieServerTemplateStorage;
+import org.kie.server.controller.impl.storage.InMemoryKieServerTemplateStorage;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+import static org.kie.server.controller.rest.ControllerUtils.marshal;
+
+public class ConcurrentModificationExceptionTest {
+
+    private KieServerTemplateStorage templateStorage = InMemoryKieServerTemplateStorage.getInstance();
+    private static final String serverTemplateId = "id";
+
+    @Before
+    public void setUp() {
+        ServerTemplate serverTemplate = new ServerTemplate(serverTemplateId, "name");
+        IntStream.range(0, 1000).boxed().map(i -> Integer.toString(i)).map(s -> new ServerInstanceKey(s, s, s, ""))
+                .forEach(instance -> serverTemplate.addServerInstance(instance));
+        templateStorage.store(serverTemplate);
+    }
+
+    private void disconnect() {
+        ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+
+        try {
+            for (ServerInstanceKey instanceKey : serverTemplate.getServerInstanceKeys()) {
+                serverTemplate.deleteServerInstance(instanceKey.getServerInstanceId());
+            }
+        } catch (Exception e) {
+            fail(e.getMessage());
+        }
+    }
+
+    @Test
+    public void testConcurrency() {
+        final ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+        marshal("application/xml", serverTemplate);
+
+        Thread disconnect = new Thread(() -> disconnect());
+        disconnect.start();
+
+        String response = marshal("application/xml", serverTemplate);
+        assertNotNull(response);
+    }
+}
--- a/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ContainerSpec.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ContainerSpec.java
@@ -15,7 +15,7 @@
 
 package org.kie.server.controller.api.model.spec;
 
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.Map;
 
 import javax.xml.bind.annotation.XmlAccessType;
@@ -33,7 +33,7 @@ public class ContainerSpec extends ContainerSpecKey  {
     @XmlElement(name = "release-id")
     private ReleaseId releasedId;
     @XmlElement(name = "configuration")
-    private Map<Capability, ContainerConfig> configs = new HashMap<>();
+    private Map<Capability, ContainerConfig> configs = new EnumMap<>(Capability.class);
     @XmlElement(name = "status")
     private KieContainerStatus status = KieContainerStatus.STOPPED;
 
@@ -58,7 +58,7 @@ public class ContainerSpec extends ContainerSpecKey  {
 
     public Map<Capability, ContainerConfig> getConfigs() {
         if (configs == null) {
-            configs = new HashMap<>();
+            configs = new EnumMap<>(Capability.class);
         }
         return configs;
     }


bxf12315:
dependent on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1140

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Thread disconnect = new Thread(() -> {
+            try {
+                deleteServerTemplate(serverTemplate);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        });
+        disconnect.start();
+
+        String response = marshal("application/xml", serverTemplate);
```
MarianMacik: This scenario can be overcome by overriding the marshal method to countDown the latch just before we delegate to the parent's implementation? Maybe we don't need Byteman at all.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/ControllerUtils.java
```diff
@@ -212,15 +212,17 @@ public static String marshal(String marshallingFormat, Object entity) {
                 break;
             }
         }
-
+        if (entity instanceof ServerTemplate) {
+            return marshaller.marshall(((ServerTemplate) entity).cloneServerTemplate());
```
MarianMacik: So there is still a tiny chance of a concurrent modification if this thread and the delete thread iterate over the `serverInstances` colletion, isn't it? Just if I get it right...

bxf12315: I think yes, last version code I added LOCK for it, but I find if remove it, the test also can pass, So I change it.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +236,48 @@ public int hashCode() {
...
+
+        serverTemplate.setMode(mode);
+
+        return serverTemplate;
+    }
+
+    protected Map<Capability, ContainerConfig> cloneContainerConfigMap(final Map<Capability, ContainerConfig> configs) {
+        final Map<Capability, ContainerConfig> cloneConfigs = new HashMap<>();
+        configs.entrySet().stream().forEach(entry -> {
+            cloneConfigs.put(entry.getKey(), cloneContainerConfig(entry.getValue()));
```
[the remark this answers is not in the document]
bxf12315: shallow copy may be good enough, but it need to change any object of ServerTemplate, My code just change one class.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
```diff
@@ -0,0 +1,59 @@
...
+                                                new ReleaseId("test1", "test1", "1"), KieContainerStatus.STARTED, containerSpec));
+
+        Map<Capability, ServerConfig> cloneServerConfig = new HashMap<Capability, ServerConfig>();
+        cloneServerConfig.put(Capability.RULE,new ServerConfig());
+        src.setConfigs(cloneServerConfig);
+        src.addServerInstance(new ServerInstanceKey("test1", "test1", "test1", "test1"));
+
+        ServerTemplate dest = src.cloneServerTemplate();
+
+        assertEquals(src, dest);
```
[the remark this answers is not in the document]
bxf12315: no,here I just check cloned object equal to original one.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test(expected = Test.None.class)
+    public void testConcurrency() throws InterruptedException {
+        final ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+
+        Thread disconnect = new Thread(() -> disconnect());
+        disconnect.start();
+
+        // Sync before marshalling
+        await().untilTrue(atomic);
```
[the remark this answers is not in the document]
bxf12315: Sonar check suggest me use awaitility.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
bxf12315: It have a sonar check issue,https://stackoverflow.com/questions/36283334/how-to-avoid-thread-sleep-in-unit-tests. So I think byteman is better way.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +241,48 @@ public int hashCode() {
...
+
+    public ServerTemplate cloneServerTemplate() {
+        synchronized (LOCK) {
+            ServerTemplate serverTemplate = new ServerTemplate();
+            serverTemplate.setId(this.getId());
+            serverTemplate.setName(this.getName());
+            containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), cloneContainerConfigMap(c.getConfigs()))));
+
+            final Map<Capability, ServerConfig> cloneServerConfig = new HashMap<>();
+            configs.entrySet().stream().forEach(entry -> cloneServerConfig.put(entry.getKey(), entry.getValue()));
```
MarianMacik: This can be still made much simpler using `final Map<Capability, ServerConfig> cloneServerConfig = new HashMap<>(configs);`. The result will be the same and it's much easier to read.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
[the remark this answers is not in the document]
MarianMacik: I understand what you check here, but you don't check if the cloned server template is really another cloned object.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
MarianMacik: But with CountDownLatch you don't use Thread.sleep() at all, it's a completely fine approach which is built-in in JDK. You just create a CountDownLatch and then use `await()` and `countDown()` methods.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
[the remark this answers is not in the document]
MarianMacik: Same here + AtomicBoolean is true from the start so this "waiting" won't wait at all.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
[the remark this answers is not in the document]
MarianMacik: Not sure I understand that. Did you mean that in case a collection is added into them, we will need to reimplement the clone method? We will need to reimplement it anyway even with this code.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +236,48 @@ public int hashCode() {
...
     }
+
+    public ServerTemplate cloneServerTemplate() {
+        ServerTemplate serverTemplate = new ServerTemplate();
+        serverTemplate.setId(this.getId());
+        serverTemplate.setName(this.getName());
+        containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), cloneContainerConfigMap(c.getConfigs()))));
+
+        final Map<Capability, ServerConfig> cloneServerConfig = new HashMap<>();
+        configs.entrySet().stream().forEach(entry -> cloneServerConfig.put(entry.getKey(), entry.getValue()));
```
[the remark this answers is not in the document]
MarianMacik: This can be still made much simpler using final Map<Capability, ServerConfig> cloneServerConfig = new HashMap<>(configs);. The result will be the same and it's much easier to read.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
bxf12315: "with CountDownLatch you don't use Thread.sleep() at all" , can not reproduce  the issue.so ...
I create three way to test,just want to run all of them in sonar.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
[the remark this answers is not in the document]
bxf12315: it's make sense.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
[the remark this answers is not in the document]
bxf12315: added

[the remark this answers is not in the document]
MarianMacik: `assertEquals(false, src == dest);` can be simplified (it is also more readable) to `assertFalse(src == dest)` or `assertTrue(src != dest)`. I also meant to compare fields inside the template, but I guess it's better than nothing now :)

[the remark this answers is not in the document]
bxf12315: Sure, but I want to wait for the sonar check's result. It will definitely bring more issues.

elguardian:
@MarianMacik @bxf12315 Just a quick look.
I don't see how this is going to fix the issue at hand. 

1) doing this new ArrayList(this.config) is not thread safe at all
2) Any collection in the server template must be in sync - block  -> the same goes for any other operation (get/set/clone) operation

3) naming convetions for cloning. Using .clone or public constructor with the same class public ServerTemplate(ServerTemplate) instead of naming it with non-standard names (it is more difficult to follow logic)

MarianMacik:
> @MarianMacik @bxf12315 Just a quick look.
> I don't see how this is going to fix the issue at hand.
> 
>     1. doing this new ArrayList(this.config) is not thread safe at all
> 
>     2. Any collection in the server template must be in sync - block  -> the same goes for any other operation (get/set/clone) operation

That's why the lock was added - either you can clone the template or you can call `deleteServerInstance`, nothing in between. This will certainly help with the issue the jira is about - it will not allow both operations to be done at the same time. But yes, we can synchronize on the whole template each time we access it, though not sure if we want to do it as we don't do it now and the issue is about these 2 particular methods called at the same time.

bxf12315:
> > @MarianMacik @bxf12315 Just a quick look.
> > I don't see how this is going to fix the issue at hand.
> > ```
> > 1. doing this new ArrayList(this.config) is not thread safe at all
> > 
> > 2. Any collection in the server template must be in sync - block  -> the same goes for any other operation (get/set/clone) operation
> > ```
> 
> That's why the lock was added - either you can clone the template or you can call `deleteServerInstance`, nothing in between. This will certainly help with the issue the jira is about - it will not allow both operations to be done at the same time. But yes, we can synchronize on the whole template each time we access it, though not sure if we want to do it as we don't do it now and the issue is about these 2 particular methods called at the same time.

About sync block, I agree with @MarianMacik. No special reason does not need to synchronize the entire ServerTemplate.

last one, about .clone and public constructor, In sonar suggest do not use .clone
,it suggest use copy method way and public constructor, it is both ok.

elguardian:
> > @MarianMacik @bxf12315 Just a quick look.
> > I don't see how this is going to fix the issue at hand.
> > ```
> > 1. doing this new ArrayList(this.config) is not thread safe at all
> > 
> > 2. Any collection in the server template must be in sync - block  -> the same goes for any other operation (get/set/clone) operation
> > ```
> 
> That's why the lock was added - either you can clone the template or you can call `deleteServerInstance`, nothing in between. This will certainly help with the issue the jira is about - it will not allow both operations to be done at the same time. But yes, we can synchronize on the whole template each time we access it, though not sure if we want to do it as we don't do it now and the issue is about these 2 particular methods called at the same time.

No. The issue is not about two methods being called at the same time. The issue is that a mutable object instance is being used in a non-thread safe way. 

Regarding the current PR.
https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R103

that is not thread safe and it is not consistent with the treatment with 

https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R187

You cannot have it half way with concurrency. If you let this to happen at some point some process will get an inconsistent state of the object or exceptions.

bxf12315:
> > > @MarianMacik @bxf12315 Just a quick look.
> > > I don't see how this is going to fix the issue at hand.
> > > ```
> > > 1. doing this new ArrayList(this.config) is not thread safe at all
> > > 
> > > 2. Any collection in the server template must be in sync - block  -> the same goes for any other operation (get/set/clone) operation
> > > ```
> > 
> > 
> > That's why the lock was added - either you can clone the template or you can call `deleteServerInstance`, nothing in between. This will certainly help with the issue the jira is about - it will not allow both operations to be done at the same time. But yes, we can synchronize on the whole template each time we access it, though not sure if we want to do it as we don't do it now and the issue is about these 2 particular methods called at the same time.
> 
> No. The issue is not about two methods being called at the same time. The issue is that a mutable object instance is being used in a non-thread safe way.
> 
> Regarding the current PR.
> [kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R103](https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R103)
> 
> that is not thread safe and it is not consistent with the treatment with
> 
> [kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R187](https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R187)
> 
> You cannot have it half way with concurrency. If you let this to happen at some point some process will get an inconsistent state of the object or exceptions.
yes, you are right, for the case When delete ServerInstanceKey and receive ServerInstanceKey execute at the same time,ConcurrentModificationException will happen. 
But the case happen in jaxb marshall, it means jaxb mashall will get all fields of ServiceTemplete,it's means other Collections field also have the issue. so I think add sync lock in copy method it makes sense.

MarianMacik:
> > 
> > 
> > That's why the lock was added - either you can clone the template or you can call `deleteServerInstance`, nothing in between. This will certainly help with the issue the jira is about - it will not allow both operations to be done at the same time. But yes, we can synchronize on the whole template each time we access it, though not sure if we want to do it as we don't do it now and the issue is about these 2 particular methods called at the same time.
> 
> No. The issue is not about two methods being called at the same time. The issue is that a mutable object instance is being used in a non-thread safe way.
> 
> Regarding the current PR.
> https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R103
> 
> that is not thread safe and it is not consistent with the treatment with
> 
> https://github.com/kiegroup/droolsjbpm-integration/pull/1962/files#diff-9e4d13a201f4713a892ef4109229f719R187
> 
> You cannot have it half way with concurrency. If you let this to happen at some point some process will get an inconsistent state of the object or exceptions.

That's what I was trying to explain - my understanding is that we don't mind this type of thread-unsafety because this was also here before. This JIRA is about making thread-safe just one combination of methods:

"ConcurrentModificationException is thrown in controller in case of race condition when Kie server is disconnected at the same time as server template is retrieved."

So the combination is retrieving the template (marshalling it) and deleting the server instance from the template.

But of course, if we want to fix the thread-safety completely, let's just lock the whole template (or make every method synchronized for the future + in case we make there more than one change at a time - so we keep the atomicity of the operation) or just make methods for adding/deleting/getting of ContainerSpec and ServerInstanceKey synchronized so we are sure we iterate over the consistent collections. Not sure what all kinds of things we do with the template.

But by looking at the `SpecManagementServiceImpl` get/read methods are not synchronized, `deleteServerInstance` is not synchronized as well, but other add/delete/update methods are. So I would probably start there. If we just synchronize it on the server template level, we might get template (when calling unsynchronized `getServerTemplate`) which is partially updated/inconsistent if we obtain template while other synchronized methods are being executed, or even the `deleteServerInstance` which is now unsynchronized.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertEquals(false, src == dest);
```
cristianonicolai: assertNotEquals

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,69 @@
...
+        templateStorage.store(serverTemplate);
+    }
+
+    private void disconnect() {
+        ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+
+        try {
+            deleteServerTemplate(serverTemplate);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
jstastny-cz: shouldn't we rather log this instead of stdout?

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test(expected = Test.None.class)
+    public void testConcurrencyWithByteman() {
+        final ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+        marshal("application/xml", serverTemplate);
+
+        Thread disconnect = new Thread(() -> disconnect());
+        disconnect.start();
+
+        String response = marshal("application/xml", serverTemplate);
+        System.out.println(response);
```
jstastny-cz: again, we should rather log this.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +243,48 @@ public int hashCode() {
...
+            RuleConfig ruleConfig = (RuleConfig) containerConfig;
+            return new RuleConfig(ruleConfig.getPollInterval(), ruleConfig.getScannerStatus());
+        }
+
+        if (containerConfig instanceof ProcessConfig) {
+            ProcessConfig processConfig = (ProcessConfig) containerConfig;
+            return new ProcessConfig(processConfig.getRuntimeStrategy(), processConfig.getKBase(), processConfig.getKSession(), processConfig.getMergeMode());
+        }
+
+        return null;
```
sutaakar: I think that this option is not valid, right? There is no other container config type apart from rule or process at this moment.
In such case I would recommend throwing here an exception. That way we make sure than if we add another container config type in the future then we won't forget to add it here.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.kie.server.controller.api.model.runtime.ServerInstanceKey;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.controller.api.storage.KieServerTemplateStorage;
+import org.kie.server.controller.impl.storage.InMemoryKieServerTemplateStorage;
+
+import static org.kie.server.controller.rest.ControllerUtils.marshal;
+
+public class ConcurrentModificationExceptionTest {
```
sutaakar: Can this test be moved to implementation module? It isn't an integration test.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -90,10 +97,10 @@ public ServerTemplate( final String id,
     }
 
     public Collection<ServerInstanceKey> getServerInstanceKeys() {
-        if (serverInstances == null) {
-            serverInstances = new ArrayList<ServerInstanceKey>();
-        }
-        return new ArrayList<ServerInstanceKey>( serverInstances );
+            if (serverInstances == null) {
```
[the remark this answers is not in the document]
bxf12315: I do not change here's logic. Just reformat it. I will redo.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
bxf12315: ok

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
bxf12315: I think you are right, but we need new type Exception firstly, what do you think WrongContainerConfigTypeException? 
@cristianonicolai @jstastny-cz

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
bxf12315: ok, which module is better? I think the test does not belong to other modules.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
bxf12315: can not use it. The code means, src equal to dest but not the same instance.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
sutaakar: I don't think we need to have some specialized exception. This is not something what should be handled on application layer, rather we need to indicate that something is wrong in application, something what should be fixed by developer.

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/src/test/java/org/kie/server/integrationtests/controller/ConcurrentModificationExceptionTest.java
sutaakar: What about kie-server-controller-impl module?

bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-impl/pom.xml
```diff
@@ -68,6 +68,11 @@
       <groupId>org.kie.server</groupId>
       <artifactId>kie-server-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-controller-rest</artifactId>
```
sutaakar: This brings a cyclic dependency between impl and rest modules.

bxf12315: 1 if move the unit test to kie-server-controller-impl, need the dependncy.
2 the dependency scope is test.
So back to old module?

bxf12315: I move the unit test to kie-server-controller-rest.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
```
cristianonicolai: 2020

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    private void deleteServerTemplate(ServerTemplate serverTemplate) throws InterruptedException {
+        for (ServerInstanceKey instanceKey : serverTemplate.getServerInstanceKeys()) {
+            serverTemplate.deleteServerInstance(instanceKey.getServerInstanceId());
+        }
+    }
+
+    @Test(expected = Test.None.class)
```
cristianonicolai: remove expected

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
cristianonicolai: makes sense

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/pom.xml
```diff
@@ -194,6 +194,7 @@
         </exclusion>
       </exclusions>
     </dependency>
+
```
cristianonicolai: remove change

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
bxf12315: done

kie-server-parent/kie-server-tests/kie-server-integ-tests-controller/pom.xml
bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +244,48 @@ public int hashCode() {
...
+            containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), cloneContainerConfigMap(c.getConfigs()))));
+
+            final Map<Capability, ServerConfig> cloneServerConfig = new EnumMap<>(Capability.class);
+            cloneServerConfig.putAll(configs);
+            serverTemplate.setConfigs(cloneServerConfig);
+
+            serverInstances.forEach(serverInstance -> serverTemplate.addServerInstance(new ServerInstanceKey(serverInstance.getServerTemplateId(), serverInstance.getServerName(), serverInstance.getServerInstanceId(), serverInstance.getUrl())));
+
+            final List<String> cloneCapabilities = new ArrayList<>();
+            cloneCapabilities.addAll(capabilities);
```
MarianMacik: Again, just use copy constructor.

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+        Map<Capability, ServerConfig> cloneServerConfig = new HashMap<Capability, ServerConfig>();
+        cloneServerConfig.put(Capability.RULE,new ServerConfig());
+        src.setConfigs(cloneServerConfig);
+        src.addServerInstance(new ServerInstanceKey("test1", "test1", "test1", "test1"));
+
+        ServerTemplate dest = src.cloneServerTemplate();
+
+        assertEquals(src, dest);
+        assertFalse(src == dest);
```
MarianMacik: `assertNotEquals` will be better here. You can also check that inner fields are equal but not the same.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -176,13 +184,8 @@ public void deleteServerInstance(String serverInstanceId) {
...
         }
-        Iterator<ServerInstanceKey> iterator = serverInstances.iterator();
-
-        while(iterator.hasNext()) {
-            ServerInstanceKey serverInstanceKey = iterator.next();
-            if (serverInstanceId.equals(serverInstanceKey.getServerInstanceId())) {
-                iterator.remove();
-            }
+        synchronized (lock) {
+            serverInstances.removeIf((ServerInstanceKey key) -> key.getServerInstanceId().equals(serverInstanceId));
```
MarianMacik: What about the other methods which modify the collections here? Namely `addServerInstance`, `addContainerSpec`, `deleteContainerSpec`? They can create this exception as well.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +244,48 @@ public int hashCode() {
...
+
+    public ServerTemplate cloneServerTemplate() {
+        synchronized (lock) {
+            ServerTemplate serverTemplate = new ServerTemplate();
+            serverTemplate.setId(this.getId());
+            serverTemplate.setName(this.getName());
+            containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), cloneContainerConfigMap(c.getConfigs()))));
+
+            final Map<Capability, ServerConfig> cloneServerConfig = new EnumMap<>(Capability.class);
+            cloneServerConfig.putAll(configs);
```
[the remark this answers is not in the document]
bxf12315: if use clone constructor,we cannot use instance LOCK.

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    public void testConcurrency() {
+        final ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
+        marshal("application/xml", serverTemplate);
+
+        Thread disconnect = new Thread(() -> disconnect());
+        disconnect.start();
+
+        String response = marshal("application/xml", serverTemplate);
+        LOGGER.info(response);
```
cristianonicolai:
```suggestion
        assertNotNull(response);
```

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @Before
+    public void setUp() {
+        ServerTemplate serverTemplate = new ServerTemplate(serverTemplateId, "name");
+        IntStream.range(0, 1000).boxed().map(i -> Integer.toString(i)).map(s -> new ServerInstanceKey(s, s, s, ""))
+                .forEach(instance -> serverTemplate.addServerInstance(instance));
+        templateStorage.store(serverTemplate);
+    }
+
+    private void disconnect() {
+        ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);
```
cristianonicolai:
```suggestion
        ServerTemplate serverTemplate = templateStorage.load(serverTemplateId);

        try {
            for (ServerInstanceKey instanceKey : serverTemplate.getServerInstanceKeys()) {
                serverTemplate.deleteServerInstance(instanceKey.getServerInstanceId());
            }
        } catch (Exception e) {
            fail(e.getMessage());
        }
```

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        try {
+            deleteServerTemplate(serverTemplate);
+        } catch (InterruptedException e) {
+            LOGGER.error(e.getMessage());
+        }
+    }
+
+    private void deleteServerTemplate(ServerTemplate serverTemplate) throws InterruptedException {
```
cristianonicolai: you can remove this method.

bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/test/java/org/kie/server/controller/rest/ConcurrentModificationExceptionTest.java
bxf12315: done

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +244,48 @@ public int hashCode() {
...
+            serverTemplate.setConfigs(cloneServerConfig);
+
+            serverInstances.forEach(serverInstance -> serverTemplate.addServerInstance(new ServerInstanceKey(serverInstance.getServerTemplateId(), serverInstance.getServerName(), serverInstance.getServerInstanceId(), serverInstance.getUrl())));
```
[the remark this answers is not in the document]
bxf12315: https://github.com/kiegroup/droolsjbpm-integration/blob/4fa524a2fc11228738fbd62c8ad4d7f848178f3d/kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java#L104
In here, it will return new instance,so we can use getServerInstanceKeys()

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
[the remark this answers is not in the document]
MarianMacik: No, I mean this:

`final Map<Capability, ServerConfig> cloneServerConfig = new EnumMap<>(configs);`
This is the same as doing:
```
final Map<Capability, ServerConfig> cloneServerConfig = new EnumMap<>(Capability.class);
cloneServerConfig.putAll(configs);
```


By clone/copy constructor I mean a constructor which has the collection as argument.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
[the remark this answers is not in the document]
MarianMacik: But there is no setter for the whole collection. Only constructor where we can set serverInstances as a whole.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
bxf12315: what's your means?

kie-server-parent/kie-server-controller/kie-server-controller-impl/src/test/java/org/kie/server/controller/api/model/spec/ServerTemplateTest.java
bxf12315: assertNotEquals(src,dest);?

MarianMacik: Sorry, my bad, I meant `assertNotSame`. This checks for identity.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
MarianMacik: I mean that these 3 methods manipulate with collections as well, so in theory they can cause concurrent modification issues as well. But given that this JIRA is specifically about use case when KIE Server disconnects, it depends on you or other stakeholders if we want to go further.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
```diff
@@ -241,4 +244,48 @@ public int hashCode() {
         result = 31 * result + (capabilities != null ? capabilities.hashCode() : 0);
         return result;
     }
+
+    public ServerTemplate cloneServerTemplate() {
+        synchronized (lock) {
+            ServerTemplate serverTemplate = new ServerTemplate();
+            serverTemplate.setId(this.getId());
+            serverTemplate.setName(this.getName());
+            containersSpec.forEach(c -> serverTemplate.addContainerSpec(new ContainerSpec(c.getId(), c.getContainerName(), c.getServerTemplateKey(), c.getReleasedId(), c.getStatus(), cloneContainerConfigMap(c.getConfigs()))));
```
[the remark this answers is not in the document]
MarianMacik: This still holds. Just delete the `cloneContainerConfigMap()` and use `new EnumMap<>(c.getConfigs())`.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
bxf12315: I think the issue's scope just has this. so we can keep it is simple.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/spec/ServerTemplate.java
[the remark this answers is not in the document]
bxf12315: yes you are right.

elguardian:
out of date

PR #1969, opened 2019-12-19
Refactor Kie server test profiles to get rid of default profile activation

sutaakar:


sutaakar:
@mbiarnes FYI this should fix the build issues.

PR #1970, opened 2019-12-19
[JBPM-8943] business central cannot show process details when configured with multiple kie-servers through smart router.

elguardian:
Added logic to avoid deep merge when the call is single result

### DIFF
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
@@ -18,6 +18,7 @@ package org.kie.server.remote.rest.common.util;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.text.MessageFormat;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
@@ -32,6 +33,7 @@ import org.kie.server.api.KieServerEnvironment;
 import org.kie.server.api.marshalling.MarshallingFormat;
 import org.kie.server.api.model.KieContainerStatus;
 import org.kie.server.api.model.ReleaseId;
+import org.kie.server.api.model.instance.Aggregatable;
 import org.kie.server.common.rest.RestEasy960Util;
 import org.kie.server.remote.rest.common.Header;
 import org.kie.server.services.api.KieServerRegistry;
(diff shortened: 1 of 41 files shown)


elguardian:
@mswiderski @MarianMacik not very convince about this. 

Several paths:
1) build a list of single result calls to kie-server
2) build a list of list result calls to kie-server

3) build dynamically of list result calls (search for implementors of ItemList; needs a dependency) -> not sure we want that

ideas are welcome

mswiderski:
@elguardian how about flagging responses with some headers about type of the response - list or single item. If not given assuming it's list.... wdyt?

elguardian:
@mswiderski works for me

elguardian:
@mswiderski changed. this approach seems cleaner but the problem is we need to go one by one call. check and tell me wdyt

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/utils/MediaTypeUtil.java
```diff
@@ -0,0 +1,25 @@
+package org.kie.server.router.utils;
```
mswiderski: missing license header

MarianMacik:
jenkins retest this

elguardian:
sure, will change the implementation.

kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/src/main/java/org/kie/server/remote/rest/common/util/RestUtils.java
```diff
@@ -94,12 +97,23 @@ public static Response createResponse(Object responseObj, Variant v, javax.ws.rs
...
+        return getVariant(headers, null);
+    }
+
+    public static Variant getVariant(HttpHeaders headers, Object obj) {
         Variant v = RestEasy960Util.getVariant(headers);
         if( v == null ) {
             v = Variant.mediaTypes(getMediaType(headers)).add().build().get(0);
         }
+
+        if (obj instanceof ItemList || (obj instanceof ServiceResponse && ((ServiceResponse) obj).getResult() instanceof ItemList)) {
```
elguardian: @mswiderski I calculated the objects being ItemList as the only ones aggregatable. I don't see if others must be in that regard.

elguardian:
added @Aggregatable annotation to define whether an object response could be aggregate or not.

gmunozfe:
jenkins retest this

elguardian:
@gom

> It looks good to me.
> 
> Just one question and one consideration @elguardian:
> 
> * Is ProcessInstance the unique model for adding Aggregatable(false) annotation? Are there no more entities that should include this annotation to avoid being aggregated?
> * Regarding aggregated elements: for these cases, requested limit applies to each kie-server behind the router, returning this one the sum of all (requested limit is not maintained, perhaps this should be documented anyway).

regarding your first bullet. Only this one is the goal of this bug.
regarding your second bullet. That is not the goal of the fix (prior condition)

elguardian:
@kiegroup/gatekeepers  can anybody take care of the merge ?

PR #1971, opened 2019-12-19
PLANNER-1745: Eliminate deprecated OptaPlanner API, clean up test

yurloc:
https://issues.redhat.com/browse/PLANNER-1745

Follows up on #1957.

- Removed negative test that has little value and would break soon again.
- Removed deprecated API.
- Using fluent config API.
- Removed raw type usage, plus other cleanup.


jiripetrlik:
Jenkins retest this

PR #1999, opened 2020-02-03
[DROOLS-5003]  Avoid generating the executable model sources if drools-model-compiler is not in the pom.xml file

lucamolteni:


lucamolteni:
jenkins test again

PR #2001, opened 2020-02-05
[DROOLS-5008] Implement tests for PMML in exec. server (nn, naive bay…

jiripetrlik:
…es, kmeans)


sutaakar:
@jiripetrlik Should the executable model be enabled also for PMML?

jiripetrlik:
@sutaakar 
Executable model should be disabled for PMML. I've set "generateModel" to NO in kjars poms.

PR #2016, opened 2020-02-18
[REPRODUCER] Reproducer for JBPM-9031

sutaakar:


sutaakar:
Fix is implemented, closing reproducer.

PR #2014, opened 2020-02-17
DROOLS-5020 DMN wire Validator to kie-maven-plugin Maven build

tarilabs:
please note tests are provided by takari framework here: https://github.com/kiegroup/droolsjbpm-integration/tree/master/kie-plugins-testing/src/test/projects/kjar-7-with-dmn

@etirelli @baldimir @hellowdan 
/cc @DuncanDoyle 

If I introduce an Error which is subject of the model analysis:
image

Then, if I really want to try anyway, by disabling the kie-maven-plugin flag:
image

I will get the same trouble now because the DMN compilation would, naturally, fail:
image

Once review approved, I will write the doc N&N

PR #2019, opened 2020-02-18
Refactor KJar builds for Kie server tests to use installed Maven binary

sutaakar:

The purpose is to get rid of MavenCLI which brings undesired undeclared dependencies like Xerces, causing issues like https://github.com/kiegroup/droolsjbpm-integration/pull/2014#issuecomment-587384823

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
@@ -18,15 +18,16 @@ package org.kie.server.integrationtests.shared;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.stream.Collectors;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.maven.cli.MavenCli;
 import org.apache.maven.project.MavenProject;
 import org.appformer.maven.integration.MavenRepository;
 import org.appformer.maven.integration.embedder.MavenProjectLoader;
@@ -61,14 +62,9 @@ public class KieServerDeployer {
     }
 
     public static void buildAndDeployMavenProject(String basedir) {
-        // need to backup (and later restore) the current class loader, because the Maven/Plexus does some classloader
-        // magic which then results in CNFE in RestEasy client
         // run the Maven build which will create the kjar. The kjar is then either installed or deployed to local and
         // remote repo
         logger.debug("Building and deploying Maven project from basedir '{}'.", basedir);
-        ClassLoader classLoaderBak = Thread.currentThread().getContextClassLoader();
-        System.setProperty("maven.multiModuleProjectDirectory", basedir); // required by MavenCli 3.3.0+
-        MavenCli cli = new MavenCli();
         List<String> mvnArgs;
         if (TestConfig.isLocalServer()) {
             // just install into local repository when running the local server. Deploying to remote repo will fail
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
```diff
@@ -61,24 +62,19 @@ public static void buildAndDeployMavenProjectFromResource(String resourcePath) {
...
-        MavenCli cli = new MavenCli();
         List<String> mvnArgs;
         if (TestConfig.isLocalServer()) {
             // just install into local repository when running the local server. Deploying to remote repo will fail
             // if the repo does not exist.
 
             // wrapping explicitly in ArrayList as we may need to update the list later (and Arrays.toList() returns
             // just read-only list)
-            mvnArgs = new ArrayList<String>(Arrays.asList("-B", "clean", "install"));
+            mvnArgs = new ArrayList<>(Arrays.asList("mvn", "--quiet", "clean", "install"));
```
lucamolteni: I usually have problem with embedded maven not showing error messages.
Can this --quiet worsen the situation?

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
```diff
@@ -87,13 +83,15 @@ public static void buildAndDeployMavenProject(String basedir) {
...
-        Thread.currentThread().setContextClassLoader(classLoaderBak);
+        // Execute the Maven build using installed Maven binary
+        try (ProcessExecutor executor = new ProcessExecutor()) {
+            String command = mvnArgs.stream().collect(Collectors.joining(" "));
+            boolean executionSuccessful = executor.executeProcessCommand(command, Paths.get(basedir));
 
-        if (mvnRunResult != 0) {
-            throw new RuntimeException("Error while building Maven project from basedir " + basedir +
-                    ". Return code=" + mvnRunResult);
+            if (!executionSuccessful) {
```
lucamolteni: We need some better error messages here. This one is too generic and when it occured to me I had to debug to discover the real cause

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
sutaakar: The errors should be still shown. From the maven help:
`-q,--quiet                             Quiet output - only show errors`

I have tried that manually, the error is shown if any error happens.

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
sutaakar: Ideally the real issue should be visible from the invoked process log - any error should be shown there.
Here in the code we can get just process exit value, that won't tell us much.

If it is not enough I can adjust the code to print complete process log (not just error) in case maven build fails. What do you think about it?

lucamolteni: If the errors are shown before it's fine, we don't need to print the log again. Thanks

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/KieServerDeployer.java
lucamolteni: Ok perfect

tarilabs:
Currently checked out purely this PR, without putting it on top by cherry-pick on the linked PR.
The problem manifest the same

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 7.495 s <<< FAILURE! - in org.kie.server.integrationtests.drools.StatelessSessionUsageIntegrationTest
[ERROR] org.kie.server.integrationtests.drools.StatelessSessionUsageIntegrationTest  Time elapsed: 7.494 s  <<< ERROR!
java.lang.NoClassDefFoundError: org/drools/compiler/kie/builder/impl/KieRepositoryImpl
	at org.kie.server.integrationtests.drools.StatelessSessionUsageIntegrationTest.deployArtifacts(StatelessSessionUsageIntegrationTest.java:56)
Caused by: java.lang.ClassNotFoundException: org.drools.compiler.kie.builder.impl.KieRepositoryImpl
	at org.kie.server.integrationtests.drools.StatelessSessionUsageIntegrationTest.deployArtifacts(StatelessSessionUsageIntegrationTest.java:56)
```

Will try again now by erasing the `.m2` for kie/drools.

PR #2020, opened 2020-02-19
[JBPM-9031] Fix condition evaluating Server template capabilities

sutaakar:


sutaakar:
@mswiderski Can you please take a look/merge?

PR #2022, opened 2020-02-20
[JBPM-9015] KIE-Server rendererd forms bind data to incorrect process variable names

elguardian:
Associate the model with the top nested form field binding name

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
@@ -29,9 +29,13 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import freemarker.cache.StringTemplateLoader;
+import freemarker.template.Configuration;
+import freemarker.template.Template;
 import org.jbpm.casemgmt.api.model.CaseDefinition;
 import org.jbpm.casemgmt.api.model.CaseRole;
 import org.jbpm.services.api.model.ProcessDefinition;
@@ -45,10 +49,6 @@ import org.kie.server.services.jbpm.ui.form.render.model.LayoutRow;
 import org.kie.server.services.jbpm.ui.form.render.model.TableInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import freemarker.cache.StringTemplateLoader;
-import freemarker.template.Configuration;
-import freemarker.template.Template;
 
 public abstract class AbstractFormRenderer implements FormRenderer {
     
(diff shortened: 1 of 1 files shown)


elguardian:
@mswiderski not really sure about this one. There too many possible cases

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
```diff
@@ -293,16 +293,21 @@ protected void processFormLayout(FormInstance topLevelForm,
...
-            jsonTemplate                
-                .append("'")
-                .append(form.getModel().getName())
-                .append("' : ")
-                .append("{")
-                .append("'")
-                .append(form.getModel().getClassName())                
-                .append("' : ")
-                .append("{");
+            Optional<FormField> fieldForm = topLevelForm.getFields().stream().filter(e -> e.getNestedForm().equals(form.getId())).findAny();
```
mswiderski: could we use this `org.kie.server.services.jbpm.ui.form.render.model.FormInstance.getNestedForm(String)` instead here?

meaning `topLevelForm..getNestedForm(form.getId())`

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
```diff
@@ -293,16 +293,21 @@ protected void processFormLayout(FormInstance topLevelForm,
...
+            jsonTemplate.append("'");
+            if (fieldForm.isPresent()) {
+                jsonTemplate.append(fieldForm.get().getBinding());
```
mswiderski: this would be ideal to be already added to the FormModel class next to classname and name. Have you looked if this can be achieved?

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
elguardian: I am lost with this comment. 
fieldForm.getNestedForm().equals(form.getId()) is not equivalent to
topLevelForm..getNestedForm(form.getId())

what exactly you want to change ?

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
elguardian: The problem is that FormModel is "clean" in the sense that the binding is introduced in the top level form and not here so you only have local information. Those comes from unmarshalling files.

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
mswiderski: wouldn't getNestedForm return the form or null for given formId? U mean to replace the entire stream, filter, etc

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
mswiderski: ok, fair enough

kie-server-parent/kie-server-services/kie-server-services-jbpm-ui/src/main/java/org/kie/server/services/jbpm/ui/form/render/AbstractFormRenderer.java
elguardian: The idea is to find the binding in the top level form. Each field has the nested form id that belongs to (AFAIK)

mswiderski: there should be a resolver that can look through all children forms so might be able to find any that is defined for the top level one

elguardian:
any progress with this @mswiderski ?

sutaakar:
Looks ok for though I am not much familiar with form rendering.
But I don't see any test covering this.

PR #2025, opened 2020-02-20
Retrieve Maven binary for Kie server tests using maven.home property

sutaakar:


sutaakar:
@lucamolteni @tarilabs This should fix the issue with downstream build (and on Windows). Maven binary is retrieved using maven.home property (pointing to Maven home of Maven instance which runs this build).

sutaakar:
In case user triggers the tests using embedded server then mvn command from the path is used.

PR #2031, opened 2020-02-28
Set correct Kie server mode in policy test suite for SpringBoot

sutaakar:
When the mode was configured for SpringBoot then this change was forgotten.


sutaakar:
Jenkins retest this

sutaakar:
Jenkins retest this

PR #2036, opened 2020-03-04
[JBPM-8910] How to use a "OR" query operator in a custom query in the query-parameters

elguardian:
supporting omision in cond-column when the expression is logical

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
@@ -729,7 +729,7 @@ public class JSONMarshaller implements Marshaller {
                 if (jp.nextToken() != JsonToken.FIELD_NAME) { // single entry; but we want modifiable
                     LinkedHashMap<String, Object> result = new LinkedHashMap<String, Object>(4);
                     result.put(field1, value1);
-                    return result;
+                    return wrapCustomObject(result);
                 }
                 String field2 = jp.getText();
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -781,10 +781,20 @@ protected Object mapObject(JsonParser jp, DeserializationContext ctxt) throws IO
...
                     return new QueryParam(column, op, condValues);
                 }
-                return result;
+
+                return wrapCustomObject(result);
             }
         }
 
+        public Object wrapCustomObject(Map<String, Object> result) {
+            if (result.containsKey("cond-operator")) {
```
gmunozfe: I think this condition always matches, so it shouldn't be needed, wdyt?

elguardian: I don't think so contains key only will be true if the object has that property

gmunozfe: Right, forget my comment, I was only thinking in json with queryparams

gmunozfe: BTW, the code-smell that SonarCloud complains, is related to convert "cond-operator" (4 times) into a constant. It may be also applied to "cond-column" and "cond-values", but it's up to you as nice-to-have.

PR #2043, opened 2020-03-12
Enable SpringBoot swagger for kie-server-integ-tests-all module

sutaakar:
As Swagger is tested in SwaggerIntegrationTest


sutaakar:
@MarianMacik Can you please take a look/merge?

PR #2048, opened 2020-03-20
[RHPAM-2810] Non potential owner of task able to update task variable through REST api

elguardian:
giving error 40x from server when it is not authorized

PR #2056, opened 2020-04-07
[BAQE-1367] administrator user should be lowercase to match keycloak

gmunozfe:
provision tests

Needed after [droolsjbpm-integration#2048](https://github.com/kiegroup/droolsjbpm-integration/pull/2048) due to in Keycloak we have "administrator" user in lowercase as Keycloak doesn't allow using capital letters.


gmunozfe:
@sutaakar please review. I will add also the bxms-qe-tests PR to align keycloak tests

gmunozfe:
@kiegroup/gatekeepers could you merge please?

PR #2067, opened 2020-04-17
[JBPM-9121] REST Process APIs should return 403 when user has no permissions

gmunozfe:
Security (based on required roles) is performed only on process operations, but REST APIs are responding with an internal error (500) instead of forbidden (403).
Added also annotations for Swagger.


gmunozfe:
Unrelated test failures.
jenkins retest this

elguardian:
wait for CI to be green. Sonar cloud is not relevant

elguardian:
Jenkins retests this please.

gmunozfe:
jenkins retest this

PR #2073, opened 2020-04-23
Kie-camel tests no longer depends on exact order of rules evaluation

jiripetrlik:
This fix is related to DROOLS-2965 where order was reversed. It is no longer
possible to compare exact content of XMLs. Only unmarshalled values are compared.


mareknovotny:
Jenkins retest this please

PR #2076, opened 2020-04-24
PLANNER-1933: Remove error message informing about an update to a solution outside of planning

wmedvede:
…

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-task-assigning/kie-server-services-task-assigning-planning/src/main/java/org/kie/server/services/taskassigning/planning/SolutionSynchronizer.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-task-assigning/kie-server-services-task-assigning-planning/src/main/java/org/kie/server/services/taskassigning/planning/SolutionSynchronizer.java
@@ -178,7 +178,7 @@ public class SolutionSynchronizer extends RunnableBase {
             LOGGER.debug("Solution Synchronizer will recover the solution from the jBPM runtime for starting the solver.");
             if (!solverExecutor.isStopped()) {
                 LOGGER.debug("Previous solver instance has not yet finished, let's wait for it to stop." +
-                                     " Next attempt will be in {} milliseconds.", syncInterval.toMillis());
+                                     " Next attempt will be in a period of {}.", syncInterval);
                 nextAction = Action.INIT_SOLVER_EXECUTOR;
             } else {
                 final TaskAssigningSolution recoveredSolution = recoverSolution();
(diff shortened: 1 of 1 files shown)


wmedvede:
Hi @rsynek would you mind review this PR?
Thanks!

wmedvede:

kie-server-parent/kie-server-services/kie-server-services-task-assigning/kie-server-services-task-assigning-planning/src/main/java/org/kie/server/services/taskassigning/planning/SolutionSynchronizer.java
```diff
@@ -178,7 +178,7 @@ Action doInitSolverExecutor() {
             LOGGER.debug("Solution Synchronizer will recover the solution from the jBPM runtime for starting the solver.");
             if (!solverExecutor.isStopped()) {
                 LOGGER.debug("Previous solver instance has not yet finished, let's wait for it to stop." +
-                                     " Next attempt will be in {} milliseconds.", syncInterval.toMillis());
+                                     " Next attempt will be in a period of {}.", syncInterval);
```
rsynek, line 181: The message is missing a time unit of the period (seconds, milliseconds?).

wmedvede: This was left this way by intention since the sync interval is now a period and not milliseconds nor seconds any more. Instead now the toString of the period will be printed and the log will print "Next attempt will be in a period of PT2S" depending on the configured value. I think this is much better.

rsynek: I see. On the other hand, should we rely on users having to understand the ISO_8601 format?

wmedvede: Well, potential users interested on scanning the server log normally are system administrators, developers or integrators, etc., looking for issues. I believe this audience can understand it. Note that it could be the same people that previously has setup/configured the installation, etc., where this value is configured via the task assigning properties exactly with the value that is printed in the log. From my point of view it's ok. In the end, you see in the log what you previously has configured.
I think it's ok, thanks.

wmedvede:

rsynek:
Merging as this PR has no potential of breaking the FDB (which is currently broken).

PR #2079, opened 2020-05-04
[JBPM-9121] Add missing jaxb dependency in integration test

sutaakar:
Required for JDK 11.


sutaakar:
@MarianMacik Can you please take a look/merge?

PR #2087, opened 2020-05-12
[Drools-5321] Randomly failing test org.kie.hacep.LocalStorageKieSessionTest.fireUntilHaltTest on timeout

desmax74:
@mariofusco

### DIFF
--- a/drools-ha/ha-core-infra/src/test/java/org/kie/hacep/LocalStorageKieSessionTest.java
+++ b/drools-ha/ha-core-infra/src/test/java/org/kie/hacep/LocalStorageKieSessionTest.java
@@ -127,7 +127,7 @@ public class LocalStorageKieSessionTest {
         assertEquals((Long) 0L, session.getFactCount().get());
     }
 
-    @Test(timeout = 10000)
+    @Test(timeout = 20000)
     public void fireUntilHaltTest() throws ExecutionException, InterruptedException {
 
         assertEquals((Long) 0L, session.getFactCount().get());


drools-ha/ha-core-infra/src/test/java/org/kie/hacep/LocalStorageKieSessionTest.java
```diff
@@ -127,7 +127,7 @@ public void deleteTest() throws ExecutionException, InterruptedException {
         assertEquals((Long) 0L, session.getFactCount().get());
     }
 
-    @Test(timeout = 10000)
+    @Test(timeout = 20000)
```
mareknovotny, line 130: I don't think  increasing timeout is good fix here

desmax74: Locally it spent 5 sec, on failure it spent 10.645 sec on pr build all six test in the class spent 24 sec and the fireuntil halt is under 10 sec and no failure, the execution time seems depend on the build execution.

mareknovotny: well there is some concurrent processing and this will be always very dependent on the environment, so I would like to see 1. eliminate concurrent access in Tests generally or 2. do not include such concurrency tests in default unit/integration run at all.

desmax74: 3 Remove the timeouts because now are useless

mareknovotny: if that resolves the problem from opposite direction, why not

mareknovotny: Are you preparing follow up PR?

desmax74: Waiting the result of build 81 with this change, if green then a PR with LocalStorage without timeouts

mareknovotny: `test timed out after 20000 milliseconds` in [#82](https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/upstream-downstream/job/droolsjbpm-build-bootstrap.downstream/82/)

PR #2033, opened 2020-03-02
[JBPM-9044] Upgrade Wildfly to 18.0.1.Final

mareknovotny:
https://issues.redhat.com/browse/JBPM-9044

Parent PR kiegroup/droolsjbpm-build-bootstrap/pull/1203

PR #2088, opened 2020-05-12
JBPM-9152: Add missing netty dependencies

jomarko:
This PR alligns dependecies of wildfly and eap7 profiles. We recently updated wildfly profily but seems same updates are needed alos for eap7.

For more details see:
- https://github.com/kiegroup/droolsjbpm-integration/pull/2033/files
- https://issues.redhat.com/browse/JBPM-9152


jomarko:

mareknovotny:
I guess we don't need to wait for FDB as there is not eap profile used @sutaakar @MarianMacik ?

sutaakar:
right

PR #2106, opened 2020-05-26
[RHPAM-2911] Null fields should not be included in JSON response

fjtirado:
New fields "parameter" in content type. If value is non-null alternative
object mapper is used to prevent null serialization

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
@@ -17,6 +17,7 @@
 package org.kie.server.api.marshalling;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Collectors;
 
@@ -57,19 +58,24 @@ public enum MarshallingFormat {
         }
     }
 
+    
     public static boolean isStrictType(String type) {
-        int idx;
-        if ((idx = type.indexOf(';')) < 0 || (idx + 1) == type.length()) {
-            return false;
-        }
-        // we map parameter=1, pararmeter=2,.... into a map
-        Map<String, String> parameters = Arrays.asList(type.substring(idx + 1).split(","))
-                                               .stream()
-                                               .filter(e -> e.split("=").length > 1) // remove bad parameters
-                                               .map(e -> e.split("="))
-                                               .collect(Collectors.toMap(e -> ((String[]) e)[0].trim(), e -> ((String[]) e)[1]));
-        return Boolean.parseBoolean(parameters.get(Marshaller.MARSHALLER_PARAMETER_STRICT));
+        String strictParam  = (String)buildParameters(type).get(Marshaller.MARSHALLER_PARAMETER_STRICT);
+        return strictParam != null && Boolean.parseBoolean(strictParam);
     }
+    
+    public static Map<String,Object> buildParameters(String contentType)
+    {
+        int idx;
+        return ((idx = contentType.indexOf(';')) < 0 || (idx + 1) == contentType.length()) 
+                ? Collections.emptyMap() 
+                : Arrays.stream(contentType.substring(idx + 1).split(",")).
+                    filter(e -> e.split("=").length > 1) // remove bad parameters
+                    .map(e -> e.split("="))
+                    .collect(Collectors.toMap(e -> e[0].trim(), e ->  e[1].trim()));
+    }
+    
+
 
     public static MarshallingFormat fromType(String type) {
         if (startsWithIgnoreCase(type, "xstream") || startsWithIgnoreCase(type, "application/xstream")) {
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
@@ -41,6 +41,7 @@ import javax.xml.bind.annotation.adapters.XmlAdapter;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapters;
 
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
@@ -93,6 +94,8 @@ public class JSONMarshaller implements Marshaller {
     private static final Logger logger = LoggerFactory.getLogger(MarshallerFactory.class);
 
     private static final boolean STRICT_ID_FORMAT = Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_SERVER_STRICT_ID_FORMAT, "false"));
+    private static final String FIELDS = "fields";
+    private static final String NOT_NULL = "not_null";
 
     private boolean formatDate = Boolean.parseBoolean(System.getProperty("org.kie.server.json.format.date", "false"));
     private String dateFormatStr = System.getProperty("org.kie.server.json.date_format", "yyyy-MM-dd'T'hh:mm:ss.SSSZ");
@@ -102,6 +105,8 @@ public class JSONMarshaller implements Marshaller {
         private boolean stripped;
 
         private boolean wrap;
+        
+        private boolean writeNull;
 
         public JSONContext() {
             this.reset();
@@ -110,6 +115,7 @@ public class JSONMarshaller implements Marshaller {
         public void reset() {
             stripped = false;
             wrap = false;
+            writeNull = true;
         }
 
         public boolean isStripped() {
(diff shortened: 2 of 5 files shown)


kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/marshal/MarshallerHelper.java
```diff
@@ -84,9 +83,23 @@ public String marshal(String marshallingFormat, Object entity) {
         	serverMarshallers.put(format, marshaller);
         }
 
-        return marshaller.marshall(entity, Collections.singletonMap(MARSHALLER_PARAMETER_STRICT, isStrictType(marshallingFormat)));
+        return marshaller.marshall(entity, buildParameters(marshallingFormat));
+    }
+    
+    
+    private  Map<String,Object> buildParameters(String contentType)
```
elguardian, line 90: reuse and set this in to the that class
https://github.com/kiegroup/droolsjbpm-integration/blob/56c0abc4a3d676ae24170642c05537577c32dad7/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java#L60

kie-server-parent/kie-server-services/kie-server-services-common/src/test/java/org/kie/server/services/impl/marshal/MarshallerHelperTest.java
```diff
@@ -218,6 +217,25 @@ public void testJsonUnmarshallProcessInstanceQueryFilterSpec() {
...
 	}
 	
+	
+	@Test
+    public void testJsonUnmarshallNull() throws JSONException {   
+	    KieServerRegistry kieServerRegistryMock = Mockito.mock(KieServerRegistry.class);
+        Set<Class<?>> extraClasses = new HashSet<>();
+        extraClasses.add(QueryFilterSpec.class);
+        Mockito.when(kieServerRegistryMock.getExtraClasses()).thenReturn(extraClasses);
+        JSONAssert.assertEquals("{\"order-by\" : null, \"order-asc\" : false, \"query-params\" : null, \"result-column-mapping\" : null}", new MarshallerHelper(kieServerRegistryMock).marshal("application/json", new QueryFilterSpecBuilder().get()), false);
```
elguardian, line 227: do some extra calls to check the reset of the thead local variable is working.

fjtirado:
After latest change the unit case is pretty similar to what the server will do
1) Instantiante a MarshallerHelper
2) Registry.getClasses will return Date and JaxbByteArray
3) Marshall will be invoked once the operation is performed 
Ill add integration test soon

fjtirado:
fields = not_null added to json integration test

gmunozfe:
jenkins retest this

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/basetests/RestJmsSharedBaseIntegrationTest.java
```diff
@@ -47,16 +50,14 @@
     @Parameterized.Parameters(name = "{index}: {0} {1}")
     public static Collection<Object[]> data() {
         KieServicesConfiguration restConfiguration = createKieServicesRestConfiguration();
-
         Collection<Object[]> parameterData = new ArrayList<Object[]>(Arrays.asList(new Object[][]
                         {
                                 {MarshallingFormat.JAXB, restConfiguration},
-                                {MarshallingFormat.JSON, restConfiguration},
+                                {MarshallingFormat.JSON,  createKieServicesRestConfiguration(Collections.singletonMap(KieServerConstants.KIE_CONTENT_TYPE_HEADER, "application/json; fields=not_null"))},
```
elguardian, line 56: This will make this very slow at integration test level (veeeeeeeeery slow).  and you are testing the full logic of marshalling not really checking whether those fields are being retrieved or not.

You need to check directly against the kie-server and see the string response whether thoses fields are returned o not.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
```diff
@@ -57,19 +58,24 @@ public static MarshallingFormat fromId(int id) {
...
-        // we map parameter=1, pararmeter=2,.... into a map
-        Map<String, String> parameters = Arrays.asList(type.substring(idx + 1).split(","))
-                                               .stream()
-                                               .filter(e -> e.split("=").length > 1) // remove bad parameters
-                                               .map(e -> e.split("="))
-                                               .collect(Collectors.toMap(e -> ((String[]) e)[0].trim(), e -> ((String[]) e)[1]));
-        return Boolean.parseBoolean(parameters.get(Marshaller.MARSHALLER_PARAMETER_STRICT));
+        return ((idx = contentType.indexOf(';')) < 0 || (idx + 1) == contentType.length()) 
+                ? Collections.emptyMap() 
+                : Arrays.stream(contentType.substring(idx + 1).split(",")).
```
elguardian, line 72: @gmunozfe how was your trick to make the split(",") more robust ? split("\\s*,\\s*) ?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -93,6 +94,8 @@
     private static final Logger logger = LoggerFactory.getLogger(MarshallerFactory.class);
 
     private static final boolean STRICT_ID_FORMAT = Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_SERVER_STRICT_ID_FORMAT, "false"));
+    private static final String FIELDS = "fields";
+    private static final String NOT_NULL = "not_null";
```
elguardian, line 98: this is nitpicking.... what about calling the parameter filter : none | not_null or something of the sort. wdyt ?

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/basetests/RestJmsSharedBaseIntegrationTest.java
fjtirado: That nulls are not returned when not null are specified is already tested in Unit test. The purposse of modifiying integration test is to test that the entity is still created when null parameters are skipped.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
fjtirado: , and ;  are not a valid character in content type, this proposed changed will not make the code more robust, but slower

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
fjtirado: you mean none being equivalent to returning everything?

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/MarshallingFormat.java
gmunozfe: `trim().split("\\s*,\\s*")`
trim() method removes leading and trailing spaces in the input string, and the regex itself handles the extra spaces around delimiter.

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/basetests/RestJmsSharedBaseIntegrationTest.java
sutaakar: I think that we should keep the default settings in here.
Testing of skipped null fields could be done in some specific integration test only, for example something like  https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/src/test/java/org/kie/server/integrationtests/jbpm/rest/ProcessServiceRestOnlyIntegrationTest.java

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
fjtirado: NA

kie-server-parent/kie-server-tests/kie-server-integ-tests-common/src/main/java/org/kie/server/integrationtests/shared/basetests/RestJmsSharedBaseIntegrationTest.java
fjtirado: Added new test class under specified directory. Thanks for the advice

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -306,6 +323,10 @@ public String marshall(Object input, Map<String, Object> parameters) {
             if (parameters.containsKey(MARSHALLER_PARAMETER_STRICT)) {
                 jsonContext.get().setWrap((boolean) parameters.get(MARSHALLER_PARAMETER_STRICT));
```
gmunozfe, line 324: This casting to boolean is failing when receiving a String parameter (true) due to changes for invoking _buildParameters_ instead of _isStrictType_ .
Failing test is _IdentifierJbpmRestIntegrationTest.testBasicJbpmRequest_ 
If always receives a String, it should be:
```suggestion
                jsonContext.get().setWrap(Boolean.parseBoolean((string) parameters.get(MARSHALLER_PARAMETER_STRICT)));
```

fjtirado: Done

PR #2107, opened 2020-05-27
JBPM-9173: Create reproducer test for dmn date on kie server evaluation

jomarko:
This PR is not a fix.
For more details see: https://issues.redhat.com/browse/JBPM-9173

### DIFF
--- a/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/src/test/filtered-resources/kjars-sources/local-date-unary-check/pom.xml
+++ b/kie-server-parent/kie-server-tests/kie-server-integ-tests-dmn/src/test/filtered-resources/kjars-sources/local-date-unary-check/pom.xml
@@ -11,7 +11,18 @@
   <version>1.0.0.Final</version>
   <packaging>kjar</packaging>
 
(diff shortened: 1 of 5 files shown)


sutaakar:
@jomarko The PR looks ok, though I would keep it open until a fix is ready.
If the fix will affect just runtime then I would prefer to have this test as unit test in runtime. If the fix will include Kie server marshalling then it is ok to have this test here as part of the fix.

jomarko:
@sutaakar one more question. Is the concept I used for passing person object (I represent as a Map) correct? Or should I pass somehow a Person java class directly?

sutaakar:
I am not sure, I don't have much knowledge of dmn.

jomarko:
@sutaakar thank you for your answer. I have added do not merge label as you suggested not to merge sooner than fix is ready.

tarilabs:
> I am not sure, I don't have much knowledge of dmn.

Just to be clear, there is nothing specific of DMN here. This is just incidentally used as part of the example.

tarilabs:
green :)

jomarko:
@sutaakar may I ask you to review when you have a moment?

jomarko:
@kiegroup/gatekeepers please merge

PR #2109, opened 2020-05-27
[JBPM-9129] Add tests for security container

gmunozfe:
Added a couple of tests for the new APIs to create process from nodeIds in the security container to check 403 Response for non-allowed users.


elguardian:
LGTM thx for the contrib

PR #2096, opened 2020-05-20
[Drools 5321] Failing test org.kie.hacep.LocalStorageKieSessionTest.fireUntilHaltTest on timeout

desmax74:
Debug on LocalMessageSystem 
@mariofusco @danielezonca

PR #2110, opened 2020-05-27
Revert "[Drools 5321] Failing test org.kie.hacep.LocalStorageKieSessionTest.fireUntilHaltTest on timeout"

danielezonca:
Reverts kiegroup/droolsjbpm-integration#2096


danielezonca:
@mareknovotny 
I think we should revert  this PR because it raises some log at debug to let @desmax74 debug the problem but it was not ready for merge.
Actually I think it was supposed to be draft
@desmax74 
Can you please confirm?

mareknovotny:
we tried to find the root cause of that timed out test @danielezonca

PR #2105, opened 2020-05-26
Refactor dependencies for Kie server integration tests

sutaakar:
@rsynek This is just an initial work in jbpm integration test module, others will follow.

PR #2114, opened 2020-06-01
Use provided scope for kjar dependencies for Kie server tests

sutaakar:
System scope was used in a try to speed up test execution, however without significant effect.
Provided scope is aligned with purpose of kjar dependencies.

Changes in https://github.com/kiegroup/droolsjbpm-integration/pull/2105 made build if these kjars failing unless artefacts with system scope were available in local repository. With this fix these artefacts are properly downloaded when kjar is built.


sutaakar:
Jenkins retest this

PR #2115, opened 2020-06-01
[JBPM-9174] CXF warning exception because of class loading

fjtirado:
Including CXF jar that was reported as missing


elguardian:
Not sure about this. Have you tried this on wildfly ?

elguardian:
Issues related.
https://issues.redhat.com/browse/JBPM-8088

PR #2118, opened 2020-06-02
[RHPAM-2782] Add reproducer as springboot test

gmunozfe:
Just changed from the original reproducer, this line in the bpmn:
`  <bpmn2:error id="java.lang.Exception" errorCode="java.lang.Exception"/>`
by
`<bpmn2:error id="java.lang.Exception" />`

If no value is specified for "errorCode", then it will catch any error that occurs in the subprocess.


gmunozfe:
jenkins retest this

elguardian:
jira closed. PR out of date.

PR #2133, opened 2020-06-16
[RHPAM-3000] Attach classes for Process migration service war

jakubschwan:
Signed-off-by: Jakub Schwan <jschwan@redhat.com>


ruromero:
I think @MarianMacik or @gmunozfe might be able to help on this review more than me.

PR #2137, opened 2020-06-22
[JBPM-9196] configure port in properties

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

https://issues.redhat.com/browse/JBPM-9196

### DIFF
--- a/process-migration-service/pom.xml
+++ b/process-migration-service/pom.xml
@@ -43,7 +43,8 @@
     <kie.server.password>usetheforce123@</kie.server.password>
 
     <!-- PIM service configuration -->
-    <pim.server.http>http://${container.hostname}:8180/</pim.server.http>
+    <pim.server.http.port>18080</pim.server.http.port>
+    <pim.server.http>http://${container.hostname}:${pim.server.http.port}/</pim.server.http>
     <pim.server.http.rest>${pim.server.http}/rest</pim.server.http.rest>
     <pim.server.http.health>${pim.server.http}/health</pim.server.http.health>
     <pim.username>kermit</pim.username>
--- a/process-migration-service/src/test/filtered-resources/project-test.yml
+++ b/process-migration-service/src/test/filtered-resources/project-test.yml
@@ -1,6 +1,6 @@
 thorntail:
-  port:
-    offset: 100
+  http:
+    port: ${pim.server.http.port}
   logging:
     root-logger:
       level: INFO


ruromero:
jenkins retest this

mareknovotny:

Testing the same change on FDB just to see port setting really fixed it in FDB too

mareknovotny:

mareknovotny:
Jenkins retest this please

mareknovotny:

ruromero:

ruromero:
jenkins retest this

ruromero:

ruromero:
jenkins please retest this

mareknovotny:

process-migration-service/pom.xml
```diff
@@ -43,7 +43,8 @@
     <kie.server.password>usetheforce123@</kie.server.password>
 
     <!-- PIM service configuration -->
-    <pim.server.http>http://${container.hostname}:8180/</pim.server.http>
+    <pim.server.http.port>18080</pim.server.http.port>
+    <pim.server.http>http://${container.hostname}:${pim.server.http.port}/</pim.server.http>
```
MarianMacik, line 47: This is just to make tests more stable? As I think that the KIE Server runs on 8080, so the conflict shouldn't be happening. Although from other tests/modules it can...

mareknovotny: the build log also shows some cargo problems with starting the service or something before ProcessMigrationIntegrationTest

MarianMacik: Yeah, that would signal that something before wasn't properly stopped. But still a good precaution to use a very different port.

ruromero: I can't explain why I can't reproduce it locally. My only guess is that maybe some other service is using port 8180 or that certain ports are not allowed.

MarianMacik: Yes, I think it can also happen if the jenkins slave is reused after the previous run wasn't successfully finished etc.

ruromero:
jenkins retest this

mareknovotny:

mareknovotny:
@ruromero @MarianMacik do you want to keep this port change too in addition what we fixed on settings.xml in CI job?

ruromero:
@mareknovotny yes I think it's better like this

PR #2170, opened 2020-07-29
[DROOLS-5534] MarshallingException occurs during REST request (JSON)

tkobayas:
…unmarshalling in KIE server.

- Use Thread Context ClassLoader. Replace only when URLClassLoader


tkobayas:
@elguardian @mariofusco 

Please review the fix. Thanks! (I will check Code Smells tomorrow)

At first, I tested with 2 approaches:

1) Set JSONMarshaller.classLoader to deserializeObjectMapper's TypeFactory. (This will take priority over thread context classloader)
```
    protected void configureMarshaller(Set<Class<?>> classes, final ClassLoader classLoader) {
        ...
        deserializeObjectMapper.setTypeFactory(TypeFactory.defaultInstance().withClassLoader(classLoader));
```
However, it results in several test failures because of classloader changes (e.g. kjar update)

2) Set JSONMarshaller.classLoader to thread context classloader in JSONMarshaller.CustomAsWrapperTypeDeserializer methods only when _baseType.getRawClass().getClassLoader() is URLClassLoader (= dependency jar)
```
        public Object deserializeTypedFromObject(JsonParser jp, DeserializationContext ctxt) throws IOException, JsonProcessingException {
            ClassLoader current = Thread.currentThread().getContextClassLoader();
            try {
                Thread.currentThread().setContextClassLoader(getClassLoaderForDeserialization());
...
        private ClassLoader getClassLoaderForDeserialization() {
            ClassLoader baseTypeClassLoader = _baseType.getRawClass().getClassLoader();
            if (baseTypeClassLoader instanceof URLClassLoader) {
                return classLoader;
            } else {
                return baseTypeClassLoader;
            }
        }
```
However, it's not yet stable (failed in MultiModuleProjectIntegrationTest).

Actually, _baseType.getRawClass().getClassLoader() is the most suitable classloader for deserialization so I should not replace it. Rather I want to fallback to a ProjectClassLoader if the class (mentioned in the JIRA) is not found by _baseType.getRawClass().getClassLoader().

So I introduced FallbackableTypeFactory. This loads a class by the specified fallback classloader only when it's not found by thread context classloader. So the effect is minimized.

tkobayas:
Jenkins retest this please

elguardian:
Not very convince. Still checking.
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-services/kie-server-services-drools/src/main/java/org/kie/server/services/drools/DroolsKieContainerCommandServiceImpl.java#L60

The types should be in the KCI IMO so not sure why it is not resolving the type when it belongs to a different class loader.

elguardian:
on hold till this PR is checked

https://github.com/kiegroup/droolsjbpm-integration/pull/2176

PR #2214, opened 2020-08-28
[JBPM-9311] KIE server will not always deploy services, stay in sync …

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9311

moving the logic for pluggable persistent template storage to the base class so every controller class behaves the same way.

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java
@@ -17,10 +17,13 @@ package org.kie.server.controller.impl;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/KieServerControllerImpl.java
```diff
@@ -62,10 +67,50 @@
...
+        ServiceLoader<PersistingServerTemplateStorageService> storageServices = ServiceLoader.load(PersistingServerTemplateStorageService.class);
+        Iterator<PersistingServerTemplateStorageService> storageServicesIterator = storageServices.iterator();
+        while (storageServicesIterator.hasNext()) {
+            PersistingServerTemplateStorageService storageService = storageServicesIterator.next();
+            logger.debug("Server template storage for kie server controller found {}", storageService.getTemplateStorage().toString());
+            servicesFound.put(storageService.getClass().getName(), storageService);
+        }
+
+        String storageServiceClassName = System.getProperty(KieServerConstants.KIE_SERVER_SERVER_TEMPLATE_STORAGE_SERVICE, "org.kie.server.controller.impl.service.FileBasedServerTemplateStorageService");
+        String []servicesClassNames = storageServiceClassName.split(",");
```
sutaakar, line 87: Does it mean that user can specify several storage classes and one of those will be used when available on classparh?

elguardian: @sutaakar Correct. I am still not sure about the approach but yes.. It will search in order and It will be pick the first one.  I didn't put any QE guy just checking if the approach would be backward compatible. We have plenty of places were the service loader works not very determinstic way. I need to check which deployments deploys what jars to see if this make sense or just setting one of them is enough.

As I didn't test if open shift controller will be enough to sync among BC instances when servers deploy... I thinks so. Maybe you have more info I didn't have the time to test.

Any ideas are welcome

sutaakar: There shouldn't be any significant problem with OpenShift controller storage as it is backed by ConfigMap which is shared with all Business central instances in openShift namespace.

sutaakar: IMHO setting one storage service should be enough.

elguardian: removed that. by default file system is used.

sutaakar (approved):
Looks good to me.
@rhtevan When this is merged the Business central image will need to be updated to set this system property.

rhtevan:
@spolti Just a heads up for the required BC image change coming down the pipe.

elguardian:
Not working as expected.

PR #2221, opened 2020-09-03
[DROOLS-5616] Be able to set with an Env Variable that KIE using XStr…

tkobayas:
…eam does not complain on unknown fields

**JIRA**: 

https://issues.redhat.com/browse/DROOLS-5616
https://issues.redhat.com/browse/RHDM-1444

<pre>
How to retest a PR or trigger a specific build:

</pre>

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
@@ -164,5 +164,7 @@ public class KieServerConstants {
     // System variable to store the enabled packages for the XStreamMarshaller
     public static final String SYSTEM_XSTREAM_ENABLED_PACKAGES = "org.kie.server.xstream.enabled.packages";
 
+    public static final String XSTREAM_IGNORE_UNKNOWN_ELEMENTS = "org.kie.server.xstream.ignore.unknown.elements";
+
     public static final String RESET_CONTAINER_BEFORE_UPDATE = "resetBeforeUpdate";
 }
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -28,6 +28,7 @@ import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
 import org.drools.core.runtime.help.impl.XStreamXML;
+import org.kie.server.api.KieServerConstants;
 import org.kie.server.api.commands.CallContainerCommand;
 import org.kie.server.api.commands.CommandScript;
 import org.kie.server.api.commands.CreateContainerCommand;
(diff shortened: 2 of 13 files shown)


tarilabs (approved):
👍  I like this A LOT :) 

_Arigatouu gozaimashita_ Toshiya, I was getting curious on this topic and I just found out you have already done it 🤣

tkobayas:
Grazie, Matteo!

@elguardian  @sutaakar  Please review and merge this if it's okay. Thanks!

sutaakar:
@tkobayas Can you please add `<argument>-Dorg.kie.server.xstream.ignore.unknown.elements=true</argument>` to https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/pom.xml#L1468 to be able to run this test also for SpringBoot? Thanks.
Otherwise looks good.

elguardian:
Not very happy about this. We are using media type parameters for this sort of things.
example
https://github.com/kiegroup/droolsjbpm-integration/commit/933120d759fae5f521cb8fb39f9d5edb16589d5e#diff-63a038f8aee4aca5fd3dad3f72359355

elguardian (changes requested):
At least add the media type parameter, for ignoring unkown fields.

tarilabs:
> At least add the media type parameter, for ignoring unkown fields.

interesting comment https://github.com/kiegroup/droolsjbpm-integration/pull/2221#pullrequestreview-482435733, since what is being discussed here is a marshaller-wide configuration..
F

elguardian:
> interesting comment [#2221 (review)](https://github.com/kiegroup/droolsjbpm-integration/pull/2221#pullrequestreview-482435733), since what is being discussed here is a marshaller-wide configuration..
> F

Marshalling wide for xml stream is never a good idea so would try to minimize that as much as possible (it is not how xml works).
In any way this is how we are treating this sort of problems with marshaller through the system.

Either this is done that way or I will reject it.

tarilabs:
I agree! What I'm saying is that for `XStream` this is a marshaller-instance-wise setting, and since you mentioned:

> We are using media type parameters for this sort of things.

I am interested to understand how you deal with this libraries limitations in a case like this one?
I have a similar issue in another project using `XStream`, and I'm hoping to learn here how to deal with this limitation in a new way. 

Do you have a link to share which deal with an _analogous_ problem such as this one, please?
The link you previously shared, IIUC, relies on the ability for Jackson to support marshaller cloning but AFAIK this feature is not available for `XStream`. How did you solved this problem?

tkobayas:
I created another PR which adds media type "ignoreUnknownElements".

https://github.com/kiegroup/droolsjbpm-integration/pull/2223/

I think the point is aligned with what @tarilabs said. As there is no way to revert xstream.ignoreUnknownElements(), I recreate an xstream instance when the media type "ignoreUnknownElements" changed from "true" to "false".

https://github.com/kiegroup/droolsjbpm-integration/pull/2223/files#diff-1f557254ee4636142b42b01b7577c95aR239

Is it good enough? Even though I introduced media type, I hope that the system property is still available for user's convenience.

jonathanvila:
I would add that doing a system wide flag is a nice and not intrusive way of configure this behavior to make KIE behave the same on XStream, JAXB or JSON ... so not hard xml validation.
Adding a request header makes current clients to recode ( I don't know if it's needed also to modify the KIE client code and the kie-camel component ? ), it's ok if wanted to have this feature request wise if needed.

elguardian:
Ok. Let's try to clarify things. 

This jira should not exist in first place. It is clearly an integration problem not a kie server problem.

You are trying to solve an integration problem in kie server. XML have a schema that must be fullifiled. It you are sending more info to the kie server this needs to be fixed in the integration part and not kie server. I understand this must be convenient but it is not the right approach to do this. If you send a xml that does not fit what is expected in xml it must throw an error. No room for discussion there, otherwise it would be difficult for troubleshot if someone sends data and gets ignored silently.

The reason I am insistent about this being fixed in another way it is because xml is not though for this and can cause problems at GSS at some point.

Parameter in media type specifying that is a way to explicityly force the integration parts to tell kie server to ignore those fields. 

Also wide flag will impact in operations at kie server level so user expectations can really get messy using kie server (not just user input)

elguardian:
reached an agreement with Jonathan Vila Lopez about this. It will require a bit more tweaking as this should only apply to the user payloads (user space) and not the entire objet.

PR #2261, opened 2020-10-05
[JBPM-9402] AuditDataReplicationKieServerTest is in probably in deadlock

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9402

Set to embedded broker instead of using full server with tcp connection


mareknovotny:
all works on my local env too.
Thanks @elguardian

mareknovotny (approved):

PR #2264, opened 2020-10-06
[JBPM-9407] Spring Boot - Multiple Data Sources and Pluggable Variable Persistence

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9407

Overwrite for spring boot for adding custom entity managers

depends on:
https://github.com/kiegroup/jbpm/pull/1769

### DIFF
--- a/kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
+++ b/kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
@@ -15,7 +15,10 @@
 
 package org.jbpm.springboot.services;
 
+import java.util.HashMap;
 import java.util.Map;
+
+import javax.persistence.EntityManagerFactory;
 
 import org.drools.core.event.AbstractEventSupport;
 import org.jbpm.casemgmt.api.event.CaseEventListener;
@@ -50,7 +53,17 @@ public class SpringKModuleDeploymentService extends KModuleDeploymentService {
         return factory;
     }
 
-    
+    @Override
+    protected Map<String, Object> buildContextParameters(KieContainer kieContainer) {
+        Map<String, Object> contextParams = new HashMap<>();
+        Map<String, EntityManagerFactory> emfsFound = context.getBeansOfType(EntityManagerFactory.class);
+        for(Map.Entry<String, EntityManagerFactory> entry : emfsFound.entrySet()) {
+            contextParams.put(entry.getKey(), entry.getValue());
+        }
+        contextParams.putAll(super.buildContextParameters(kieContainer));
+        return contextParams;
+    }
+
     public void setContext(ApplicationContext context) {
         this.context = context;
     }


kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
```diff
@@ -50,7 +53,16 @@ protected RegisterableItemsFactory getRegisterableItemsFactory(AuditEventBuilder
         return factory;
     }
 
-    
+    protected Map<String, Object> buildContextParameters(KieContainer kieContainer) {
```
gmunozfe, line 56: Though it's not mandatory, perhaps it's more clear if "Override" annotation is explicitly shown.

gmunozfe (commented):
Looks good, just a minor comment for enhancing clarity

kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
```diff
@@ -50,7 +53,16 @@ protected RegisterableItemsFactory getRegisterableItemsFactory(AuditEventBuilder
...
+        Map<String, Object> contextParams = new HashMap<>();
+        Map<String, EntityManagerFactory> emfsFound = context.getBeansOfType(EntityManagerFactory.class);
+        for(Map.Entry<String, EntityManagerFactory> entry : emfsFound.entrySet()) {
+            contextParams.put(entry.getKey(), entry.getValue());
+        }
+        contextParams.putAll(super.buildContextParameters(kieContainer));
```
gmunozfe, line 62: In the parent invocation, ("entityManagerFactory", emf) is also added, shouldn't it be redundant?

gmunozfe (commented):
I've got also a question regarding parent invocation which also adds the emf.

afalhambra (commented):
Looks good to me though it would be very useful to have this particular scenario in a test case.
Good work, thanks

kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
elguardian: done

kie-spring-boot/kie-spring-boot-autoconfiguration/jbpm-spring-boot-autoconfiguration/src/main/java/org/jbpm/springboot/services/SpringKModuleDeploymentService.java
elguardian: not really. 
Not all deployment service use this spring for instance kie server. so the parents needs to add the entity manager.
Also we need the parameters defined in there.
There is no point to do any check away just to remove the jbpm entity manager as it cannot be override.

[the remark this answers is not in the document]
elguardian: not really. 
Not all deployment service use this spring for instance kie server. so the parents needs to add the entity manager.
Also we need the parameters defined in there.
There is no point to do any check away just to remove the jbpm entity manager as it cannot be override.

afalhambra (commented):
Nicely done. Will you add a test for this scenario? That's the only thing I'm missing here. Thanks

PR #2265, opened 2020-10-06
[RHDM-1461] allow to define an AgendaFilter when starting a process

mariofusco:


elguardian:
depends on 
https://github.com/kiegroup/droolsjbpm-knowledge/pull/470
https://github.com/kiegroup/drools/pull/3144
https://github.com/kiegroup/jbpm/pull/1771

elguardian:
remove the package-lock.json from the commit.

@mariofusco  I can see only one "but"
if the StartProcessCommand now accepts a filter I don't see in any commit how that can be used from the integration point of view. You added a new attribute there and this can be used as command. so how do we plan to support that when we are sending a command to kie server ? I guess something in the api server code should allow to pass the AgendaFilter right ?

mariofusco:
@elguardian I added the AgendaFilter to the StartProcess command. It should work exactly in the same way of the FireAllRules command. So the kie-server should receive that command with the serialized AgendaFilter and act accordingly. Am I missing something?

elguardian:
@mariofusco Just curious about what sort of AgendaFilter. Maybe I am missing something but I did not see any implementation that could be serialize/deserialize to be used in a marshaller

elguardian (approved):

mariofusco:
@elguardian we have some predefined and serializable AgendaFilter implementations, see for instance https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/base/RuleNameEqualsAgendaFilter.java and https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/base/RuleNameStartsWithAgendaFilter.java

In the same way I think that users can define their own AgendaFilter implementations and make them serializable as we did with those ones.

PR #2269, opened 2020-10-07
[JBPM-9416] node ids missing from ProcessServiceBase  causing process just being created.

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9416

missing node id list when starting from node ids without correlation key.


elguardian:
Jenkins retest this please.

gmunozfe (commented):
Fix looks good to me, but shouldn't we also update the test _StartProcessServiceIntegrationTest#testStartProcessFromNodeId_ to check that the process is really launched from the targeted node and verifying then that works as expected?

elguardian:
@gmunozfe done

PR #1495, opened 2018-06-14
DROOLS-1663 Kie DMN support for IMPORT decisions between DMN files

tarilabs:
/cc @mswiderski really sorry I missed this!

PR #2284, opened 2020-10-15
[JBPM-9436]  AMQ Streams (Kafka) listener functionality / consumer

fjtirado:
Adding Kafka dependency to server common so it is added to j2ee war and spring boot starters

**JIRA**: 

[JBPM-9436](https://issues.redhat.com/browse/JBPM-9436)

Depends on: 
[#1495](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1495)

### DIFF
--- a/drools-ha/pom.xml
+++ b/drools-ha/pom.xml
@@ -75,17 +75,6 @@
       <!-- kafka -->
       <dependency>
         <groupId>org.apache.kafka</groupId>
-        <artifactId>kafka-clients</artifactId>
-        <version>${version.org.apache.kafka}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.kafka</groupId>
-        <artifactId>kafka-clients</artifactId>
-        <version>${version.org.apache.kafka}</version>
-        <classifier>test</classifier>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.kafka</groupId>
         <artifactId>kafka_2.12</artifactId>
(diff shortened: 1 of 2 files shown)


drools-ha/pom.xml
```diff
@@ -24,7 +24,7 @@
     <surefire.forkCount>1</surefire.forkCount>
     <version.javax.el>3.0.1-b08</version.javax.el>
     <version.javax.ws.rs>2.0.0.Final</version.javax.ws.rs>
-    <version.org.apache.kafka>2.2.1</version.org.apache.kafka>
+    <version.org.apache.kafka_2>2.2.1</version.org.apache.kafka_2>
```
elguardian, line 27: why is this change required ?

elguardian:
I am going to reject this change. You are packaging a dependency that is not needed for this JIRA. So there is no point to packaging in kie server when the event emitter is something optional.

fjtirado:
> I am going to reject this change. You are packaging a dependency that is not needed for this JIRA. So there is no point to packaging in kie server when the event emitter is something optional.
> 
> I might be the case that for a another jira (listener) this could be needed. but let's not mix stuff.

The version of kafka client to be use on jbpm-workitem should match the one used in emitter, thats the reason for this PR. The version for kafka should be on kie parent

drools-ha/pom.xml
fjtirado: This is a typo

PR #2288, opened 2020-10-19
[JBPM-9436] AMQ Streams (Kafka) listener functionality / consumer

fjtirado:
Implementing Kafka server extension 

**JIRA**: 

[JBPM-9436](https://issues.redhat.com/browse/JBPM-9436)

Depends on: 
[#drools-build-bootstrap:1502](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1502) and [#jbpm:1785](https://github.com/kiegroup/jbpm/pull/1785)

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.server</groupId>
+    <artifactId>kie-server-services</artifactId>
+    <version>7.47.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>kie-server-services-kafka</artifactId>
+
+  <name>KIE :: Execution Server :: Services :: Kafka Extension</name>
+  <description>KIE Kafka jBPM Execution Server Extension</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-internal</artifactId>
+    </dependency>
+   
+
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-api</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.activation</groupId>
+          <artifactId>activation</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-services-common</artifactId>
+      <exclusions>
+        <exclusion>
+          <!-- Collides with xml-apis:xml-apis -->
+          <groupId>javax.xml.stream</groupId>
+          <artifactId>stax-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.activation</groupId>
+          <artifactId>activation</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-runtime-manager</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-shared-services</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-services-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-kie-services</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-services-jbpm</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+       <groupId>org.apache.kafka</groupId>
+       <artifactId>kafka-clients</artifactId>
+    </dependency>
+    <!-- TEST -->
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/CloudEvent.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/CloudEvent.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.server.services.jbpm.kafka;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class CloudEvent<T> {
+
+    private static ObjectMapper mapper = new ObjectMapper()
+            .setDateFormat(new SimpleDateFormat(System.getProperty(
+                    KafkaServerExtension.KAFKA_EXTENSION_PREFIX + "json.date_format", System.getProperty(
+                            "org.kie.server.json.date_format",
+                            "yyyy-MM-dd'T'HH:mm:ss.SSSZ"))));
+
+    private String specVersion;
+    private Date time;
+    private String id;
+    private String type;
+    private String source;
+    private T data;
+
+    public static <T> CloudEvent<T> read(byte[] bytes, Class<T> type) throws IOException, ParseException {
+        JsonNode node = mapper.readTree(bytes);
+        CloudEvent<T> cloudEvent = new CloudEvent<>();
+        if (node.has("id")) {
+            cloudEvent.id = node.get("id").asText();
+        }
+        if (node.has("source")) {
+            cloudEvent.source = node.get("source").asText();
+        }
+        if (node.has("type")) {
+            cloudEvent.type = node.get("type").asText();
+        }
+        if (node.has("specversion")) {
+            cloudEvent.specVersion = node.get("specversion").asText();
+        }
+        if (node.has("time")) {
+            cloudEvent.time = mapper.getDateFormat().parse(node.get("time").asText());
+        }
+        if (node.has("data")) {
+            cloudEvent.data = mapper.treeToValue(node.get("data"), type);
+        }
+        return cloudEvent;
+    }
+
+    private CloudEvent() {}
+
+    public String getSpecVersion() {
+        return specVersion;
+    }
+
+    public Date getTime() {
+        return time;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public String getType() {
+        return type;
+    }
+
+    public String getSource() {
+        return source;
+    }
+
+    public T getData() {
+        return data;
+    }
+
+    @Override
+    public String toString() {
+        return "CloudEvent [specversion=" + specVersion + ", time=" + time + ", id=" + id + ", type=" + type +
+               ", source=" + source + ", data=" + data + "]";
+    }
+}
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
@@ -1,0 +1,504 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+package org.kie.server.services.jbpm.kafka;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.BiConsumer;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.IsolationLevel;
+import org.apache.kafka.common.errors.WakeupException;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.jbpm.services.api.DeploymentEvent;
+import org.jbpm.services.api.DeploymentEventListener;
+import org.jbpm.services.api.DeploymentService;
+import org.jbpm.services.api.ListenerSupport;
+import org.jbpm.services.api.ProcessService;
+import org.jbpm.services.api.model.DeployedAsset;
+import org.jbpm.services.api.model.MessageDesc;
+import org.jbpm.services.api.model.ProcessDefinition;
+import org.jbpm.services.api.model.SignalDesc;
+import org.jbpm.services.api.model.SignalDescBase;
+import org.kie.server.api.KieServerConstants;
+import org.kie.server.services.api.KieContainerInstance;
+import org.kie.server.services.api.KieServerExtension;
+import org.kie.server.services.api.KieServerRegistry;
+import org.kie.server.services.api.SupportedTransports;
+import org.kie.server.services.impl.KieServerImpl;
+import org.kie.server.services.jbpm.JbpmKieServerExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KafkaServerExtension implements KieServerExtension, DeploymentEventListener, Runnable {
+
+    public static final String EXTENSION_NAME = "Kafka";
+    static final String KAFKA_EXTENSION_PREFIX = "org.kie.server.jbpm-kafka.ext.";
+    static final String TOPIC_PREFIX = KAFKA_EXTENSION_PREFIX + "topics.";
+    private static final Logger logger = LoggerFactory.getLogger(KafkaServerExtension.class);
+
+    private AtomicBoolean initialized = new AtomicBoolean();
+    // Kafka consumer
+    private Consumer<String, byte[]> consumer;
+    // JBPM services
+    private ListenerSupport deploymentService;
+    private ProcessService processService;
+    // Executor thread for dispatching signals to jbpm core
+    private AtomicReference<ExecutorService> notifyService = new AtomicReference<>();
+    // registration and classes information
+    private Map<String, Map<SignalDesc, Collection<String>>> topic2Signal = new HashMap<>();
+    private Map<String, Map<MessageDesc, Collection<String>>> topic2Message = new HashMap<>();
+    private Map<String, Collection<Class<?>>> classes = new ConcurrentHashMap<>();
+    // synchronization variables
+    private AtomicBoolean consumerReady = new AtomicBoolean();
+    private Lock changeRegistrationLock = new ReentrantLock();
+    private Lock consumerLock = new ReentrantLock();
+    private Condition isSubscribedCond = changeRegistrationLock.newCondition();
+
+    @Override
+    public boolean isInitialized() {
+        return initialized.get();
+    }
+
+    @Override
+    public boolean isActive() {
+        return !Boolean.parseBoolean(System.getProperty(KieServerConstants.KIE_KAFKA_SERVER_EXT_DISABLED, "true"));
+    }
+
+    @Override
+    public void init(KieServerImpl kieServer, KieServerRegistry registry) {
+        if (initialized.get()) {
+            logger.warn("Kafka extension already initialized");
+            return;
+        }
+
+        KieServerExtension jbpmExt = registry.getServerExtension(JbpmKieServerExtension.EXTENSION_NAME);
+        if (jbpmExt == null) {
+            logger.warn("Extension " + JbpmKieServerExtension.EXTENSION_NAME + " is required");
+            return;
+        }
+
+        for (Object service : jbpmExt.getServices()) {
+            if (deploymentService == null && DeploymentService.class.isAssignableFrom(service.getClass())) {
+                deploymentService = (ListenerSupport) service;
+            } else if (processService == null && ProcessService.class.isAssignableFrom(service.getClass())) {
+                processService = (ProcessService) service;
+            }
+            if (deploymentService != null && processService != null) {
+                break;
+            }
+        }
+        if (deploymentService == null) {
+            throw new IllegalStateException("Cannot find deployment service");
+        }
+        if (processService == null) {
+            throw new IllegalStateException("Cannot find process service");
+        }
+        deploymentService.addListener(this);
+        initialized.set(true);
+    }
+
+    @Override
+    public void destroy(KieServerImpl kieServer, KieServerRegistry registry) {
+        if (deploymentService != null) {
+            deploymentService.removeListener(this);
+        }
+        if (consumerReady.compareAndSet(true, false)) {
+            notifyService.getAndSet(null).shutdownNow();
+            consumer.wakeup();
+            consumerLock.lock();
+            try {
+                consumer.unsubscribe();
+                consumer.close(Duration.ofSeconds(Long.getLong(KAFKA_EXTENSION_PREFIX + "close.timeout", 30L)));
+            } finally {
+                consumerLock.unlock();
+            }
+            consumer = null;
+        }
+        changeRegistrationLock.lock();
+        try {
+            topic2Signal.clear();
+            topic2Message.clear();
+            isSubscribedCond.signal();
+        } finally {
+            changeRegistrationLock.unlock();
+        }
+        classes.clear();
+        processService = null;
+        deploymentService = null;
+        initialized.set(false);
+    }
+
+    protected Consumer<String, byte[]> getKafkaConsumer() {
+        Map<String, Object> props = new HashMap<>();
+        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, System.getProperty(
+                KAFKA_EXTENSION_PREFIX + "boopstrap.servers", "localhost:9092"));
+        // read only committed events
+        props.put(ConsumerConfig.ISOLATION_LEVEL_CONFIG, IsolationLevel.READ_COMMITTED.toString().toLowerCase());
+        // do not automatically create topics by default
+        props.put(ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG, Boolean.getBoolean(
+                KAFKA_EXTENSION_PREFIX + "auto.create.topics"));
+        props.put(ConsumerConfig.GROUP_ID_CONFIG, System.getProperty(KAFKA_EXTENSION_PREFIX + "group.id",
+                "jbpm-consumer"));
+        return new KafkaConsumer<>(props, new StringDeserializer(), new ByteArrayDeserializer());
+    }
+
+    @Override
+    public void onDeploy(DeploymentEvent event) {
+        updateRegistration(event, this::updateTopics);
+    }
+
+    @Override
+    public void onUnDeploy(DeploymentEvent event) {
+        updateRegistration(event, this::removeTopics);
+    }
+
+    @Override
+    public void onActivate(DeploymentEvent event) {
+        updateRegistration(event, this::updateTopics);
+    }
+
+    @Override
+    public void onDeactivate(DeploymentEvent event) {
+        updateRegistration(event, this::removeTopics);
+    }
+
+    private void updateTopics(String deploymentId, ProcessDefinition processDefinition) {
+        updateTopics(topic2Signal, deploymentId, processDefinition.getSignalsDesc());
+        updateTopics(topic2Message, deploymentId, processDefinition.getMessagesDesc());
+    }
+
+    private void removeTopics(String deploymentId, ProcessDefinition processDefinition) {
+        removeTopics(topic2Signal, deploymentId, processDefinition.getSignalsDesc());
+        removeTopics(topic2Message, deploymentId, processDefinition.getMessagesDesc());
+    }
+
+    private <T extends SignalDescBase> void updateTopics(Map<String, Map<T, Collection<String>>> topic2SignalBase,
+                                                         String deploymentId,
+                                                         Collection<T> signals) {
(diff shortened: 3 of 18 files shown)


kie-server-parent/kie-server-services/kie-server-services-kafka/pom.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.server</groupId>
+    <artifactId>kie-server-services</artifactId>
+    <version>7.45.0-SNAPSHOT</version>
```
mareknovotny, line 9: fix this to 7.46.0-SNAPSHOT

mareknovotny (changes requested):
there is used old snapshot version instead of the one already on master branch

kie-server-parent/kie-server-services/kie-server-services-kafka/pom.xml
fjtirado: > there is used old snapshot version instead of the one already on master branch

Thanks for pointing out.

mareknovotny:
Jenkins retest this please

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
+package org.kie.server.services.jbpm.kafka;
```
afalhambra, line 1: Red Hat copyright is missing.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
+
+public class KakfaServerExtensionTest {
+
+    private static class MockKafkaServerExtension extends KafkaServerExtension {
+
+        private MockConsumer<String, byte[]> consumer = new MockConsumer<String, byte[]>(OffsetResetStrategy.EARLIEST);
```
afalhambra, line 44: Can be replaced with empty diamonds
```suggestion
        private MockConsumer<String, byte[]> consumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST);
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+import org.mockito.verification.VerificationMode;
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
+
+public class KakfaServerExtensionTest {
```
afalhambra, line 40: Typo
```suggestion
public class KafkaServerExtensionTest {
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        when(processDefinition.getSignalsMetadata()).thenReturn(Collections.singletonList(SignalDescImpl.from(
+                new Signal("MySignal", "MySignal", "String"))));
+        when(deployedUnit.getDeployedAssets()).thenReturn(Collections.singletonList(processDefinition));
+        extension.onDeploy(new DeploymentEvent("MyDeploy", deployedUnit));
+        publishEvent("MySignal", "{\"id\":\"javi\",\"type\":\"one\",\"source\":\"pepe\",\"data\":\"javierito\"}");
+        verify(processService, getTimeout()).signalEvent("MyDeploy", "MySignal", "javierito");
+    }
+
+    @Test
+    public void testKafkaServerExecutorMessage() throws InterruptedException {
```
afalhambra, line 97: Checked exception is not thrown within this method
```suggestion
    public void testKafkaServerExecutorMessage() {
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        msg.setName("Hello");
+        msg.setType("String");
+        when(processDefinition.getSignalsMetadata()).thenReturn(Collections.singletonList(SignalDescImpl.from(msg)));
+        extension.onDeploy(new DeploymentEvent("MyDeploy", deployedUnit));
+        publishEvent("Hello", "{\"id\":\"javi\",\"type\":\"one\",\"source\":\"pepe\",\"data\":\"pepe\"}");
+        verify(processService, getTimeout()).signalEvent("MyDeploy", "Message-Hello", "pepe");
+    }
+
+    @Test
+    public void testKafkaServerExecutorMessageTopic() throws InterruptedException {
```
afalhambra, line 109: Checked exception is not thrown within this method

```suggestion
    public void testKafkaServerExecutorMessageTopic() {
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        when(deployedUnit.getDeployedAssets()).thenReturn(Collections.singletonList(processDefinition));
+    }
+
+    @After
+    public void close() {
+        extension.destroy(server, registry);
+    }
+
+    @Test
+    public void testKafkaServerExecutorSignal() throws InterruptedException {
```
afalhambra, line 87: Checked exception is not thrown within this method

```suggestion
    public void testKafkaServerExecutorSignal() {
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        return timeout(Long.getLong(KafkaServerExtension.KAFKA_EXTENSION_PREFIX + "poll.interval", 1L) * 2000).times(1);
+    }
+
+    private void publishEvent(String topic, String cloudEventText) {
+        Set<String> topics = extension.getMockConsumer().subscription();
+        assertTrue("Topic " + topic + " not found", topics.contains(topic));
+        List<TopicPartition> partitions = Collections.singletonList(new TopicPartition(topic, 0));
+        extension.getMockConsumer().rebalance(partitions);
+        Map<TopicPartition, Long> partitionsBeginningMap = new HashMap<TopicPartition, Long>();
+        Map<TopicPartition, Long> partitionsEndMap = new HashMap<TopicPartition, Long>();
```
afalhambra, line 137: Can be replaced with empty diamonds
```suggestion
        Map<TopicPartition, Long> partitionsBeginningMap = new HashMap<>();
        Map<TopicPartition, Long> partitionsEndMap = new HashMap<>();
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        long records = 10L;
+        for (TopicPartition partition : partitions) {
+            partitionsBeginningMap.put(partition, 0l);
+            partitionsEndMap.put(partition, records);
+        }
+        extension.getMockConsumer().updateBeginningOffsets(partitionsBeginningMap);
+        extension.getMockConsumer().updateEndOffsets(partitionsEndMap);
+        extension.getMockConsumer().addRecord(new ConsumerRecord<String, byte[]>(topic, 0, 0L, "",
```
afalhambra, line 145: same as above
```suggestion
        extension.getMockConsumer().addRecord(new ConsumerRecord<>(topic, 0, 0L, "",
```

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
```diff
@@ -0,0 +1,149 @@
...
+    private void publishEvent(String topic, String cloudEventText) {
+        Set<String> topics = extension.getMockConsumer().subscription();
+        assertTrue("Topic " + topic + " not found", topics.contains(topic));
+        List<TopicPartition> partitions = Collections.singletonList(new TopicPartition(topic, 0));
+        extension.getMockConsumer().rebalance(partitions);
+        Map<TopicPartition, Long> partitionsBeginningMap = new HashMap<TopicPartition, Long>();
+        Map<TopicPartition, Long> partitionsEndMap = new HashMap<TopicPartition, Long>();
+        long records = 10L;
+        for (TopicPartition partition : partitions) {
+            partitionsBeginningMap.put(partition, 0l);
```
afalhambra, line 140:
```suggestion
            partitionsBeginningMap.put(partition, 0L);
```

afalhambra (changes requested):
Excellent work! - only minor comments, mostly cosmetic ones.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/CloudEvent.java
```diff
@@ -0,0 +1,96 @@
...
+    private Date time;
+    private String id;
+    private String type;
+    private String source;
+    private T data;
+
+    public static <T> CloudEvent<T> read(byte[] bytes, Class<T> type) throws IOException, ParseException {
+        JsonNode node = mapper.readTree(bytes);
+        CloudEvent<T> cloudEvent = new CloudEvent<>();
+        if (node.has("id")) {
```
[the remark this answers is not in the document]
fjtirado: I believe readTree hardly retursn null, hence there is not point in checking

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+import org.kie.server.services.api.KieServerRegistry;
+import org.kie.server.services.api.SupportedTransports;
+import org.kie.server.services.impl.KieServerImpl;
+import org.kie.server.services.jbpm.JbpmKieServerExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KafkaServerExtension implements KieServerExtension, DeploymentEventListener, Runnable {
+
+    public static final String EXTENSION_NAME = "Kafka";
```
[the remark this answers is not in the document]
fjtirado: Here Im just following what has been done in other extensions. It makes sense in static context, as for example, in the code it is used the EXTENSION_NAME of BPMNServerExtension

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+            consumer.close();
+            consumer = null;
+        }
+        initialized.set(false);
+    }
+
+    protected Consumer<String, byte[]> getKafkaConsumer() {
+        Map<String, Object> props = new HashMap<>();
+        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, System.getProperty(
+                KAFKA_EXTENSION_PREFIX + "boopstrap.servers", "localhost:9092"));
```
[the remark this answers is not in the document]
fjtirado: for Kafka, host and port are specified together.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KakfaServerExtensionTest.java
fjtirado: good catch

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
elguardian, line 137: If you are setting by default this extension as true
https://github.com/kiegroup/droolsjbpm-integration/pull/2288/files#diff-821631efbb543afdf97147bcc31174b3fa3f48fc3b97159a0536a484468d0c4cR82
(disable false)

this is not necessaraly true as this can be active without any kafka server in the environment. Have you checked ?

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+                updater.accept(event.getDeploymentId(), (ProcessDefinition) asset);
+            }
+            synchronized (consumer) {
+                consumer.subscribe(topic2Signal.keySet());
+            }
+        }
+    }
+
+    private String topicFromSignal(SignalDesc signal) {
+        return System.getProperty(TOPIC_PREFIX + signal.getName(), signal.getName());
```
elguardian, line 193: this is far to generic and it is not documented in the BAPL. 
signal names within process does not necessarily mean the same among process I would argue this is a good idea in the general case. 
While I can see the benefit (and indeed you can leave it like this). You would need to search first for more particular signal definition

The most particular case (this should be mandatory at least)
TOPIC_PREFIX.<deploymentId>.<processId>.signalName=signalName

more general (by deployment)
TOPIC_PREFIX.<deploymentId>.signalName = signalName

or your case (by entire server -> this is ok too)
TOPIC_PREFIX.signalName = signalName

Also requires to say that some signal is bounded. For instance is there is a signal registed by a process there is need to set an info kjar - process id - signal has been bounded to topic (whatever topic is mapped) through system property whaever.

The spec how to define the mapping should be part of the documentation of your BAPL as well.

This would add a lot of info regarding the mapping and allow diagnostics.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+
+    private void processEvent(ConsumerRecord<String, byte[]> event) {
+        SignalInfo signalInfo = topic2Signal.get(event.topic());
+        if (signalInfo != null) {
+            try {
+                String type = signalInfo.getSignalDesc().getName();
+                if (signalInfo.getSignalDesc().getSignalType() == SignalType.MESSAGE) {
+                    type = "Message-" + type;
+                }
+                processService.signalEvent(signalInfo.getDeploymentId(), type, CloudEvent.read(event.value(),
```
elguardian, line 284: debug levels regarding what signal you received and where are you sending it.
trace would be the message content.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+        return !Boolean.getBoolean(KieServerConstants.KIE_KAFKA_SERVER_EXT_DISABLED);
+    }
+
+    @Override
+    public void init(KieServerImpl kieServer, KieServerRegistry registry) {
+        if (initialized.getAndSet(true)) {
+            return;
+        }
+        pollService = Executors.newSingleThreadScheduledExecutor();
+        notifyService = Executors.newCachedThreadPool();
```
elguardian, line 91: make this configurable (number of threads)

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    @Override
+    public void onDeactivate(DeploymentEvent event) {
+        updateRegistration(event, this::removeTopics);
+    }
+
+    private void updateTopics(String deploymentId, ProcessDefinition processDefinition) {
+        for (SignalDesc signal : processDefinition.getSignalsMetadata()) {
+            topic2Signal.computeIfAbsent(topicFromSignal(signal), t -> new SignalInfo(deploymentId,
```
elguardian, line 170: if we have several deployments with the same signal name, wouldn't be overriding the deployment Id ?

elguardian (changes requested):
Aside from the comments that needed to be addressed.

* Lack of integration test
* I don't see anything related to spring boot support

drools-ha/pom.xml
```diff
@@ -73,17 +73,6 @@
...
-        <groupId>org.apache.kafka</groupId>
-        <artifactId>kafka-clients</artifactId>
-        <version>${version.org.apache.kafka}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.kafka</groupId>
-        <artifactId>kafka-clients</artifactId>
-        <version>${version.org.apache.kafka}</version>
-        <classifier>test</classifier>
-      </dependency>
```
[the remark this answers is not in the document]
fjtirado: This is forcing drools-ha to use the same kafka dependency than consumer and emitter. It is done int he same jira to avoid a window of time in which several kafka client version might be colliding

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+import org.kie.server.services.api.KieContainerInstance;
+import org.kie.server.services.api.KieServerExtension;
+import org.kie.server.services.api.KieServerRegistry;
+import org.kie.server.services.api.SupportedTransports;
+import org.kie.server.services.impl.KieServerImpl;
+import org.kie.server.services.jbpm.JbpmKieServerExtension;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KafkaServerExtension implements KieServerExtension, DeploymentEventListener, Runnable {
```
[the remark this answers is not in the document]
fjtirado: I prefer to save object creation but making the extension implementing them

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,307 @@
...
+                events = consumer.subscription().isEmpty() ? ConsumerRecords.empty() : consumer.poll(Duration.ZERO);
+            }
+            for (ConsumerRecord<String, byte[]> event : events) {
+                notifyService.submit(() -> processEvent(event));
+            }
+        }
+    }
+
+    private void processEvent(ConsumerRecord<String, byte[]> event) {
+        SignalInfo signalInfo = topic2Signal.get(event.topic());
```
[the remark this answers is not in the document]
fjtirado: Implementation changed to deal with two different methods: getSignalsDesc and getMessagesDesc. This certainly complicates implementation here, but allow message and signal diverge on future (right now, they have the same data, thats why the implementation was done assuming identity between them)

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: Map structure changed to fix this

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: I think the usage of newCachedThreadPool matches the use case, as per documentation
* Creates a thread pool that creates new threads as needed, but
     * will reuse previously constructed threads when they are
     * available.  These pools will typically improve the performance
     * of programs that execute many short-lived asynchronous tasks.
     * Calls to {@code execute} will reuse previously constructed
     * threads if available. If no existing thread is available, a new
     * thread will be created and added to the pool. Threads that have
     * not been used for sixty seconds are terminated and removed from
     * the cache. Thus, a pool that remains idle for long enough will
     * not consume any resources. Note that pools with similar
     * properties but different details (for example, timeout parameters)
     * may be created using {@link ThreadPoolExecutor} constructors.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: do we really need to check that? The kafka broker might be down when we start the server. I think this refers as if we want this extension to be active or not, regardless the status of the kafka broker

fjtirado: Problem was that kie server wont start when extension is enabled and kafka broker down. Consumer is now lazily initialized to preven that. Deployment with signals/messages will fail if attempted when kafka broker is down (so user might retry)

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: done

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: Upper limit to avoid thread exhaustion set

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/CloudEvent.java
```diff
@@ -0,0 +1,96 @@
...
+                            "yyyy-MM-dd'T'HH:mm:ss.SSSZ"))));
+
+    private String specVersion;
+    private Date time;
+    private String id;
+    private String type;
+    private String source;
+    private T data;
+
+    public static <T> CloudEvent<T> read(byte[] bytes, Class<T> type) throws IOException, ParseException {
```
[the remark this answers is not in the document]
fjtirado: Using list of classes returned by DepployedEvent

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: Leaving at it is for now. Opened https://issues.redhat.com/browse/JBPM-9455 to allow fine-grainer approach.

fjtirado:
jenkins retest this

drools-ha/pom.xml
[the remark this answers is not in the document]
elguardian: what a about just removing 
https://github.com/kiegroup/droolsjbpm-integration/blob/a95e2687e0fe78678090b8ee6dcb30f1a09dfdb6/drools-ha/pom.xml#L27 ?

and remove the version element of those dependencies ?

this modules depends on kie parent

process-migration-service/frontend/package-lock.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@babel/cli": {
       "version": "7.5.5",
-      "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/@babel/cli/-/cli-7.5.5.tgz",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.5.5.tgz",
```
elguardian, line 9: change not required. remove this file from the commit

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,453 @@
...
+                            eventsPerTopic.compute(event.topic(), (k, v) -> v == null ? 1 : v++);
+                        }
+                        logger.debug("Number of events received per topic {}", eventsPerTopic);
+                    }
+                    for (ConsumerRecord<String, byte[]> event : events) {
+                        notifyService.submit(() -> processEvent(event));
+                    }
+                }
+            }
+        }
```
elguardian, line 375: if we have consumer but not topics this loop does not sleep. we will eat CPU non stop without interruption.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,453 @@
...
+                                                         ConsumerRecord<String, byte[]> event,
+                                                         Signaller signaller) {
+        Map<T, Collection<String>> signalInfo = topic2SignalBase.get(event.topic());
+        if (signalInfo != null) {
+            for (Map.Entry<T, Collection<String>> entry : signalInfo.entrySet())
+                try {
+                    String signalName = entry.getKey().getName();
+                    for (String deploymentId : entry.getValue()) {
+                        CloudEvent<?> cloudEvent = CloudEvent.read(event.value(), getDataClass(deploymentId, entry
+                                .getKey()));
```
elguardian, line 419: this is ok but has a limitation. we won't support in kafka events typed json unmarshaller. 
```
{
    map : {
         "key1" : {
               "com.my.simple.type" : {
                       "field" : "hello"
               }
          }
    }

```
(for instance when we have Map, List so the values are not clear and require types being in the json. 
This limitation should be in the BAPL.

elguardian (changes requested):
cpu eater needs to be addressed and document the limitation of marshalling.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: good catch!! This bug was provoked by changing from ScheduledService to backgroung thread. Fixed used condition lock

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: Limitation docummented in bapl

process-migration-service/frontend/package-lock.json
fjtirado: removed

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
fjtirado: good catch, fixed with a condition object

mareknovotny:
Jenkins retest this please

afalhambra (commented):
Excellent job, only thing I'm missing is:

- Some tests when Kakfa is down and check for expected behaviour.
- Some tests to check the behaviour when using these new KafkaServerExtension properties.
- Integration tests using springboot, though I believe Gonzalo will work on that

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,492 @@
...
+    public void destroy(KieServerImpl kieServer, KieServerRegistry registry) {
+        if (deploymentService != null) {
+            deploymentService.removeListener(this);
+        }
+        if (consumerReady.compareAndSet(true, false)) {
+            notifyService.getAndSet(null).shutdownNow();
+            consumer.wakeup();
+            consumerLock.lock();
+            try {
+                consumer.close();
```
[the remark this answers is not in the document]
fjtirado: Im not sure is required, but probably does not harm, so adding

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,497 @@
...
+        return new KafkaConsumer<>(props, new StringDeserializer(), new ByteArrayDeserializer());
+    }
+
+    @Override
+    public void onDeploy(DeploymentEvent event) {
+        updateRegistration(event, this::updateTopics);
+    }
+
+    @Override
+    public void onUnDeploy(DeploymentEvent event) {
```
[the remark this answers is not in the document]
fjtirado: No, we are undeploying one container, there might be other containers listening for events

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KafkaServerExtensionTest.java
```diff
@@ -0,0 +1,237 @@
...
+        extension.destroy(server, registry);
+        mockConsumer = new MockConsumer<String, byte[]>(OffsetResetStrategy.EARLIEST);
+        when(processDefinition.getSignalsDesc()).thenReturn(Collections.emptyList());
+        when(processDefinition.getMessagesDesc()).thenReturn(Collections.singletonList(MessageDescImpl.from(
+                msg)));
+        extension.setKafkaConsumer(mockConsumer);
+        extension.init(server, registry);
+        extension.onDeploy(new DeploymentEvent("MyDeploy5", deployedUnit));
+        publishEvent("Hello2", "{\"id\":\"javi\",\"type\":\"one\",\"source\":\"pepe\",\"data\":\"pepe\"}");
+        verify(processService, getTimeout()).signalEvent("MyDeploy5", "Message-Hello2", "pepe");
```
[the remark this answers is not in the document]
fjtirado: Sure, test added

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -0,0 +1,497 @@
...
+        return events;
+    }
+
+    private void dispatchEvents(ConsumerRecords<String, byte[]> events) {
+        if (consumerReady.get() && !events.isEmpty()) {
+            if (logger.isDebugEnabled()) {
+                printEventsLog(events);
+            }
+            for (ConsumerRecord<String, byte[]> event : events) {
+                if (notifyService.get() != null) {
```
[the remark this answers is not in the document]
fjtirado: this will be null only when destroying the extension. so Im removing the check and if the exception ever happen, it will catched and printed

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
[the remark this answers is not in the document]
fjtirado: Property added, default 30 seconds, like no parameter close

kie-server-parent/kie-server-services/kie-server-services-kafka/src/test/java/org/kie/server/services/jbpm/kafka/KafkaServerExtensionTest.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+
+    private VerificationMode getTimeout() {
+        return getTimeout(1);
+    }
+
+    private VerificationMode getTimeout(int times) {
+        return timeout(Long.getLong(KafkaServerExtension.KAFKA_EXTENSION_PREFIX + "poll.interval", 1L) * 2000).times(
```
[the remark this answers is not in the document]
fjtirado: Ill test that one, other I think should be tested by integration test

fjtirado:
Kafka Integration test will be designed by QE (Gonzalo) both for emitter and consumer. I agree with him because he was already familiar with testcontainers

elguardian:
LGTM but let's wait for CI clean. I don't think changes at jbpm side caused this but just in case let's run again. If it goes red needs to be checked.

elguardian:
Jenkins retest this

mareknovotny:
there are like 12 failed tests with "just" Comparison failure, so it means the test fails on expected value which are in this PR case not null or uninitialized, but just different. Please look at these tests.

fjtirado:
> there are like 12 failed tests with "just" Comparison failure, so it means the test fails on expected value which are in this PR case not null or uninitialized, but just different. Please look at these tests.

Yes, they were faulty test, trying to fix them, first attempt failed, so I take a closer look at the flow. Reason they are failing now is that we change hascode and equals of ProcessAssetDesc, so this is side effect that make the faulty test to arise now

elguardian:
failures not related this time.
Jenkins retest this

fjtirado:
@kiegroup/cloud-gatekeepers Finally green ;)

fjtirado:
@kiegroup/gatekeepers  Please merge this one

elguardian (approved):

afalhambra (approved):

PR #2296, opened 2020-10-25
DROOLS-5756 DMN kie-camel local/embedded execution capability and demo

tarilabs:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5756

**referenced Pull Requests**: none

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-camel/src/test/java/org/kie/camel/embedded/camel/component/KieEmbeddedEndpointDMNTest.java
+++ b/kie-camel/src/test/java/org/kie/camel/embedded/camel/component/KieEmbeddedEndpointDMNTest.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


kie-camel/src/test/java/org/kie/camel/embedded/camel/component/KieEmbeddedEndpointDMNTest.java
```diff
@@ -0,0 +1,120 @@
...
+        final Processor dmnToHeader = DecisionsToHeadersProcessor.builder("dmnResult",
+                                                                          "canDrinkHeader",
+                                                                          "Can Drink?")
+                                                                 .build();
+
+        return new RouteBuilder() {
+            public void configure() throws Exception {
+                from("direct:start").claimCheck(ClaimCheckOperation.Push)
+                                    .process(toMap)
+                                    .process(toDMNCommand)
```
danielezonca, line 82: What about merge `ToMapProcessor` logic with `ToDMNEvaluateAllCommandProcessor`? Something like an additional optional parameter for the constructor to specify where to put the exchange body. If this value is not provided and the body is a `Map` we can use the whole object as input.
Wdyt?

danielezonca (approved):
LGTM, approving with an optional comment :+1:

kie-camel/src/test/java/org/kie/camel/embedded/camel/component/KieEmbeddedEndpointDMNTest.java
tarilabs: I thought about that indeed, but at least in my experience Camel route is very similar to ETL, so you want the basic components/block to not do "too much" if possible.
This way:

- if the user want to collapse the two steps into one, can always do it since they are very trivial
- if the user has other steps or different steps, can compose the steps in the best fit: example, the user might be looking to transform the exchange into a Map as part of another ETL step.

So I see your point, but providing the basic steps and the basic version of the components, is the best approach in my limited Camel experience

hellowdan (approved):

PR #2303, opened 2020-10-30
[JBPM-9453] Exclusions at spring-boot-starter-test after junit bumped to 4.13.1

gmunozfe:
**JIRA**: [JBPM-9453](https://issues.redhat.com/browse/JBPM-9453)


afalhambra (approved):
nice catch!

mareknovotny (approved):

mareknovotny:
Could you check the test results if they are executed properly now? I can see 75 tests were skipped, but in different test classes. Do not want to go from left to right side and do have skipped another set of tests ;) Otherwise it looks ready to be merged

gmunozfe:
@mareknovotny yes, tests are executed properly in this branch (skipped are those marked to be disabled, meanwhile current behaviour at master branch after bumping junit is not even to execute runner, so total tests are fewer)

PR #2316, opened 2020-11-16
[JBPM-9437] Updated data object is not take into account when the project is redeployed in the "development mode"

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9437

SNAPSHOT is determined like the same version which is a problem.


elguardian:
Jenkins retest this

PR #2369, opened 2021-01-13
[JBPM-9558]  KIE using XStream does not complain on unknown fields

fjtirado:
In order to "relax" deserialization requirements, user can specifiy as parameter of contentype, ignoreUnknown=true, which will ignore unknown fields durign xstream deserialization. 

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9558)

Depends on
https://github.com/kiegroup/kie-soup/pull/173

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/Marshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/Marshaller.java
@@ -25,19 +25,23 @@ public interface Marshaller {
 
     static String MARSHALLER_PARAMETER_STRICT = "strict";
 
-    public default String marshall(Object input, Map<String, Object> parameters) {
+    default String marshall(Object input, Map<String, Object> parameters) {
         return marshall(input);
     }
 
-    public String marshall(Object input);
+    String marshall(Object input);
 
-    public <T> T unmarshall(String input, Class<T> type);
+    <T> T unmarshall(String input, Class<T> type);
 
-    public void dispose();
+    default <T> T unmarshall(String input, Class<T> type, Map<String, Object> parameters) {
+        return unmarshall(input, type);
+    }
 
-    public MarshallingFormat getFormat();
+    void dispose();
 
-    public void setClassLoader(ClassLoader classloader);
+    MarshallingFormat getFormat();
 
-    public ClassLoader getClassLoader();
+    void setClassLoader(ClassLoader classloader);
+
+    ClassLoader getClassLoader();
 }
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
@@ -26,6 +26,8 @@ import java.util.Set;
 
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider;
+import com.thoughtworks.xstream.mapper.ElementIgnoringMapper;
+import com.thoughtworks.xstream.mapper.Mapper;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
 import org.drools.core.runtime.help.impl.XStreamXML;
 import org.kie.server.api.commands.CallContainerCommand;
@@ -81,10 +83,13 @@ import static org.kie.soup.xstream.XStreamUtils.createNonTrustingXStream;
 
 public class XStreamMarshaller implements Marshaller {
 
+    public static final String XSTREAM_IGNORE_UNKNOWN = "ignore-unknown-elements";
     private static final Logger logger = LoggerFactory.getLogger(XStreamMarshaller.class);
     protected XStream xstream;
     protected ClassLoader classLoader;
-    protected Map<String, Class> classNames = new HashMap<String, Class>();
+    protected Map<String, Class<?>> classNames = new HashMap<>();
+
+    private ThreadLocal<XStreamContext> xstreamContext = ThreadLocal.withInitial(XStreamContext::new);
 
     // Optional marshaller extensions to handle new types / configure custom behavior
     private static final List<XStreamMarshallerExtension> EXTENSIONS;
@@ -113,16 +118,42 @@ public class XStreamMarshaller implements Marshaller {
         EXTENSIONS.forEach(ext -> ext.extend(this));
     }
 
+    private static class XStreamContext {
+
+        private boolean ignoreUnknownElements;
+
+        public boolean isIgnoreUnknownElements() {
+            return ignoreUnknownElements;
+        }
+
+        public void setIgnoreUnknownElements(boolean ignoreUnknownElements) {
+            this.ignoreUnknownElements = ignoreUnknownElements;
+        }
+    }
+
+    protected class CustomElementIgnore extends ElementIgnoringMapper {
+
+        public CustomElementIgnore(Mapper wrapped) {
+            super(wrapped);
+        }
+
+        @Override
+        public boolean isIgnoredElement(String name) {
+            return xstreamContext.get().isIgnoreUnknownElements();
+        }
+    }
+
     protected void buildMarshaller(Set<Class<?>> classes,
                                    final ClassLoader classLoader) {
         this.xstream = XStreamXML.newXStreamMarshaller(createNonTrustingXStream(
                 new PureJavaReflectionProvider(),
                 next -> {
                     return new MapperWrapper(chainMapperWrappers(new ArrayList<>(EXTENSIONS),
-                                                                 next)) {
-                        public Class realClass(String elementName) {
+                            new CustomElementIgnore(next))) {
 
-                            Class customClass = classNames.get(elementName);
+                        @Override
+                        public Class<?> realClass(String elementName) {
+                            Class<?> customClass = classNames.get(elementName);
                             if (customClass != null) {
                                 return customClass;
                             }
@@ -151,7 +182,7 @@ public class XStreamMarshaller implements Marshaller {
         this.xstream.denyTypes(voidDeny);
 
         this.xstream.addPermission(new KieServerTypePermission(classes));
-        String classWildcards[] = {"org.kie.api.pmml.*", "org.kie.pmml.pmml_4_2.model.*"};
+        String[] classWildcards = {"org.kie.api.pmml.*", "org.kie.pmml.pmml_4_2.model.*"};
         this.xstream.allowTypesByWildcard(classWildcards);
 
         AbstractScoreXStreamConverter.registerScoreConverters(xstream);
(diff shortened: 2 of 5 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -143,86 +162,107 @@ private MapperWrapper chainMapperWrappers(List<XStreamMarshallerExtension> exten
...
 
     @Override
     public String marshall(Object objectInput) {
         return xstream.toXML(objectInput);
     }
 
+    private XStream getProperXstream(Map<String, Object> parameters) {
+        Object ignoreString = parameters.get(XSTREAM_IGNORE_UNKNOWN);
+        if (ignoreString != null && Boolean.parseBoolean(ignoreString.toString())) {
+            ignoreLock.lock();
```
fjtirado, line 234: Not sure if lazy initialization is the best aprroach here. 
We are certainly saving memory if nobody (as we expect) is using the parameter in a sunny day scenario. 
On the other hand, when the parameter is enable, threads execution will be serialized because of the lock. 
I think this is better because it turns out that xstream object is not really inmutable

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -250,6 +290,14 @@ public ClassLoader getClassLoader() {
     }
 
     public XStream getXstream() {
+        // if someone is invoking this method is very likely that xtream will change
+        // so a new one will need to be created
+        ignoreLock.lock();
```
fjtirado, line 295: xstream not mutable, for completeness we need to reinstantiate xstreamignore if needed

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
elguardian: locks are not needed.

elguardian:
marshaller is created upon deployment and cached so there is not need for multithread. You can add user space objects during creation to the marshaller and ignore them as upon deployment you will have access to any java class in the kjar

elguardian (changes requested):
needs another approach

fjtirado:
> needs another approach

Changed to use Custom mapper and thread local

fjtirado:
@elguardian Ready to re-review

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -81,10 +83,13 @@
 
 public class XStreamMarshaller implements Marshaller {
 
+    public static final String XSTREAM_IGNORE_UNKNOWN = "ignoreUnknown";
```
elguardian, line 86: change the name for 

type-validation: ignore-unkown-elements

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -220,12 +251,21 @@ public String marshall(Object objectInput) {
     @Override
     public <T> T unmarshall(String input,
                             Class<T> type) {
+        xstreamContext.get().setIgnoreUnknownElements(false);
```
elguardian, line 254: just call the other unmarshall with one parameter XSTREAM_IGNORE_UNKNOWN = false

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
```diff
@@ -220,12 +251,21 @@ public String marshall(Object objectInput) {
...
+        return (T) xstream.fromXML(input);
+    }
+
+    @Override
+    public <T> T unmarshall(String input, Class<T> type, Map<String, Object> parameters) {
+        Object ignoreString = parameters.get(XSTREAM_IGNORE_UNKNOWN);
+        xstreamContext.get().setIgnoreUnknownElements(ignoreString != null && Boolean.parseBoolean(ignoreString
+                .toString()));
         return (T) xstream.fromXML(input);
```
elguardian, line 263: reset in a finally block the flag  of ignore elements

elguardian (changes requested):
some minor clean up

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
fjtirado: ok, I was trying to boost up the performance of the non parameter unmarshall a bit. Changed

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
fjtirado: done

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/XStreamMarshaller.java
fjtirado: removing a boolean is equivalent to setting it to false, since this is set before each unmarshall, is it really worthing to include a finally block to reset if to false?
I do not think so, but done anyway.

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/CustomXstreamMarshallerBuilder.java
```diff
@@ -35,7 +35,8 @@ public Marshaller build(Set<Class<?>> classes, MarshallingFormat format, ClassLo
             return new XStreamMarshaller(classes, classLoader) {
                 @Override
                 protected void buildMarshaller(Set<Class<?>> classes, ClassLoader classLoader) {
-                    xstream = XStreamUtils.createNonTrustingXStream(new PureJavaReflectionProvider(), new DomDriver("UTF-8", new XmlFriendlyNameCoder("_-", "_")));
+                    xstream = XStreamUtils.createNonTrustingXStream(new PureJavaReflectionProvider(), new DomDriver("UTF-8", new XmlFriendlyNameCoder("_-", "_"))
+                            , next -> new CustomElementIgnore(next));
```
afalhambra, line 39: the only way to get a buildMarshaller with this `ignoreUnknownFields` feature is by using and passing this new `CustomElementIgnore` - so I'm just thinking whether it would be worth to make it `public` instead of `protected` - otherwise we are forcing customer to create a package with the same structure - or am I wrong? wdyt?

afalhambra (commented):
Just one comment, others are just cosmetic changes here.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/Marshaller.java
```diff
@@ -33,6 +33,10 @@ public default String marshall(Object input, Map<String, Object> parameters) {
 
     public <T> T unmarshall(String input, Class<T> type);
 
+    public default <T> T unmarshall(String input, Class<T> type, Map<String, Object> parameters) {
```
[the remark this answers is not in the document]
fjtirado: I removed all public modifiers in interface methods o be "symettric"

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/CustomXstreamMarshallerBuilder.java
fjtirado: This class is intender to be used only in inherited class, so protected is right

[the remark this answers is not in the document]
fjtirado: done

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/XStreamMarshallerTest.java
```diff
@@ -249,4 +253,30 @@ public void testUnmarshallDateObject() {
...
+    @Test
+    public void testUnmarshallDateObjectWithExtraField() {
+        String expectedString = "<date-object>\n" +
+                                "  <localDate>2017-01-01</localDate>\n" +
+                                "  <localDateTime>2017-01-01T10:10:10</localDateTime>\n" +
+                                "  <localTime>10:10:10</localTime>\n" +
+                                "  <offsetDateTime>2017-01-01T10:10:10+01:00</offsetDateTime>\n" +
+                                "  <unknownField>jojojojo</unknownField>\n" +
+                                "</date-object>";
+        Set<Class<?>> extraClasses = new HashSet<Class<?>>();
```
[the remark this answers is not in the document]
fjtirado: done

afalhambra (approved):
Great job @fjtirado!

fjtirado:
@elguardian I think we are ready to go

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/XStreamMarshallerTest.java
```diff
@@ -249,4 +253,30 @@ public void testUnmarshallDateObject() {
...
+        assertThrows(UnknownFieldException.class, () -> marshaller.unmarshall(expectedString, DateObject.class));
+        DateObject dateObject = marshaller.unmarshall(expectedString, DateObject.class, Collections.singletonMap(
+                XStreamMarshaller.XSTREAM_IGNORE_UNKNOWN, Boolean.TRUE));
+        assertNotNull(dateObject);
+        assertEquals(LocalDate.of(2017, 1, 1), dateObject.getLocalDate());
+        assertEquals(LocalDateTime.of(2017, 1, 1, 10, 10, 10), dateObject.getLocalDateTime());
+        assertEquals(LocalTime.of(10, 10, 10), dateObject.getLocalTime());
+        assertEquals(OffsetDateTime.of(LocalDateTime.of(2017, 1, 1, 10, 10, 10), ZoneOffset.ofHours(1)), dateObject
+                .getOffsetDateTime());
+        assertThrows(UnknownFieldException.class, () -> marshaller.unmarshall(expectedString, DateObject.class));
```
gmunozfe, line 280: this was checked some lines above. Perhaps, were you trying to add the false flag?
```suggestion
        assertThrows(UnknownFieldException.class, () -> marshaller.unmarshall(expectedString, DateObject.class), Collections.singletonMap(XStreamMarshaller.XSTREAM_IGNORE_UNKNOWN, Boolean.FALSE));
```

gmunozfe (commented):
Looks good to me, just a comment in the test for adding false flag

kie-server-parent/kie-server-api/src/test/java/org/kie/server/api/marshalling/XStreamMarshallerTest.java
fjtirado: The repeteated class is that the flag is properly reset to false after a call with the flag equal to true, hence no need to test the false value

elguardian (approved):

PR #2373, opened 2021-01-20
[JBPM-9566] Smart router remove container even if there are other server instances active

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9566

### DIFF
--- a/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/Configuration.java
+++ b/kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/Configuration.java
@@ -101,6 +101,9 @@ public class Configuration {
         List<String> hosts = hostsPerContainer.get(containerId);
         if (hosts != null) {
             hosts.remove(serverUrl);
+            if(hosts.isEmpty()) {
+                hostsPerContainer.remove(containerId);
+            }
         }
 
         this.listeners.forEach(l -> l.onContainerRemoved(containerId, serverUrl));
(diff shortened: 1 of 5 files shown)


elguardian:
Jenkins retest this

gmunozfe:
> support running several Kie servers running under one URL

@sutaakar your concerns are not clear to me. Let me show you a picture with the scenario of balancers:

Screenshot from 2022-03-24 11-08-35

There, we have 2 balancers and 3 networks:
- network1 with the smart-router and balancers 1 and 2
- network2 with balancer1 and kie-server1 and 2
- network3 with balancer2 and kie-server1 (same name)

With self-register feature, each kie-server should send the KIE_SERVER_LOCATION of the balancer for the smart-router to forward the requests, turning out the following routing table:

| containers  | server-ids | locations |
| ------------- | ------------- | ------------- |
| container1 | server1 | balancer1:8666 |
| container1 | server2 | balancer1:8666 |
| container1 | server1 | balancer2:8666 |

From my point of view, 2 first rows are redundant and we should not keep them repeated.

Smart-router just "talks" with balancers (in the figure). 
Suppose balancer1 forwards a request to server2 and at that moment, that connection gets lost:
- Smart Router will delete URL for balancer:8666, as no response was obtained for that request (circuit open)
- Then, Smart Router launches "pinging failed host" mechanism with balancer:8666, and this one forwards to server1, so the connection is recovered and again written down in the table.

Balancer1 should balance all the load to server1 until server2 connection comes back healthy, but this is a task for the balancer.

From the point of view of the router, not allowing duplicates in these scenarios should be correct, as requests are not sending duplicated to the same balancer. Wdyt?

sutaakar:
@gmunozfe The issue I see is with removal of Kie server.
Lets say that there is deployed your example and Kie server 2 from network 2 is stopped. In such case the Kie server send a request to Smart router to remove itself from routing table. In case Smart router contains only one entry for `balancer1:8666` then it is removed, making Kie server 1 from network 2 unavailable for users.
Smart router would have to contain some login to realize that there is some another Kie server in the removed URL to keep the entry.

sutaakar:
I just think that such limitation can be considered a regression. Such Kie server removal can be more common for example in OpenShift, manual interaction in every case when Kie server is scaled down looks to me like an overkill.

mareknovotny:
can we close this PR? it seems the request was closed in jira and it is in conflict anyway

elguardian:
yep, it can be closed.,

PR #2379, opened 2021-01-26
[JBPM-9541] Kafka extension to support custom format

fjtirado:
Fixing leak in default factory implementation
Adding JavaDoc
System property for default reader and writer

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9541)


elguardian:
please use another jira. otherwise This gets hard to track

PR #2400, opened 2021-02-19
[DROOLS-5702] Using rules sources generated inside pmml compiler.

gitgabrio:
@danielezonca @mariofusco @jiripetrlik

This PR removes Rules source generation from maven-plugin/GeneratePMMLMojo

Depends on https://github.com/kiegroup/drools/pull/3395

### DIFF
--- a/kie-plugins-testing/src/test/projects/kjar-12-with-pmml-trusty/pom.xml
+++ b/kie-plugins-testing/src/test/projects/kjar-12-with-pmml-trusty/pom.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.kie</groupId>
+  <artifactId>kie-maven-plugin-test-kjar-12</artifactId>
+  <version>1.0.0.Final</version>
+
+  <packaging>kjar</packaging>
+
+  <properties>
+<!--    <kie.version>7.51.0-SNAPSHOT</kie.version>-->
+    <generateModel>NO</generateModel>
+    <generateDMNModel>NO</generateDMNModel>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+  </properties>
+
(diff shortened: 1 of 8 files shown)


gitgabrio:
Jenkins retest this

mariofusco (approved):

danielezonca (approved):

kie-plugins-testing/src/test/projects/kjar-12-with-pmml-trusty/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.kie</groupId>
+        <artifactId>kie-maven-plugin</artifactId>
+        <version>${org.kie.version}</version>
+        <extensions>true</extensions>
+        <configuration>
+          <properties>
+            <drools.dialect.java.compiler.lnglevel>1.8</drools.dialect.java.compiler.lnglevel>
```
jiripetrlik, line 64: Is this necessary? I think that Java 1.8 is used by Kie by default isn't it?

jiripetrlik (commented):
@gitgabrio 
Thank you for your PR. I have a one question.

kie-plugins-testing/src/test/projects/kjar-12-with-pmml-trusty/pom.xml
gitgabrio: @jiripetrlik 
I found that definition in other pom and, since I could not know why it is there nor what may happen in whatever possible environment if I remove it, I would prefer to keep it. One thing that cross my mind is that all the Kogito environment needs Java 11 - do not know if somehow/somewhere the twos may interact

jiripetrlik: @gitgabrio 
Understand. So let's keep it this way. To be honest I'm not sure about whether and how is kie-maven-plugin used in Kogito. Maybe @mariofusco knows whether it can cause any problems in Kogito.

jiripetrlik (approved):

kie-plugins-testing/src/test/projects/kjar-12-with-pmml-trusty/pom.xml
gitgabrio: @jiripetrlik As far as I know it is not used inside Kogito - but that code may be executed (?) in a machine where also Kogito is executed. E.g. in my machine I set Java 11 as default to work with Kogito

gitgabrio:
@danielezonca Hi! Could you please merge this, squashing commits? Thanks!
Together with https://github.com/kiegroup/drools/pull/3395

PR #2417, opened 2021-03-09
[JBPM-9639]Add cypress dependencies

prashuts:
**Thank you for submitting this pull request**

**JIRA**: _(https://issues.redhat.com/browse/JBPM-9639)_ 

[link](https://issues.redhat.com/browse/JBPM-9639)

**referenced Pull Requests**: _(please edit the URLs of referenced pullrequests if they exist)_

* paste the link(s) from GitHub here
* link 2
* link 3 etc.

<details>
<summary>
</summary>

</details>


process-migration-service/frontend/package.json
```diff
@@ -13,7 +13,11 @@
     "start": "node src/bin/server.js",
     "test": "jest",
     "precommit": "lint-staged",
-    "locktt": "locktt"
+    "locktt": "locktt",
+    "cypress:run": "cypress run --project ./frontend",
+    "cypress:run-with-dev-server": "start-server-and-test cypress:run'",
```
jstastny-cz, line 18: please call this script `test:it`
I assume that start-server-and-test:
* `start` script is invoked by default
* the pim frontend is running on same port as is default for the tool

Can you verify?

jstastny-cz: And we need to invoke this script `test:it` from pom.xml, similarly to how `test` script is invoked. Only difference will be binding it to `integration-test` phase.

Additional task will be checking if failsafe plugin is configured for the project.

process-migration-service/frontend/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
```
jstastny-cz, line 2: I would remove this file if not planned to use - I mean in tests you could put here a pre-exported migration plan and test importing back. But this example.json is useless.

jstastny-cz (changes requested):

process-migration-service/frontend/cypress/fixtures/example.json
prashuts: ok

process-migration-service/frontend/package.json
prashuts: ok

prashuts: done

prashuts: > please call this script `test:it`
> I assume that start-server-and-test:
> 
> * `start` script is invoked by default
> * the pim frontend is running on same port as is default for the tool
> 
> Can you verify?

Yes updated scripts section and kept only test:it

process-migration-service/frontend/package.json
```diff
@@ -13,7 +13,8 @@
     "start": "node src/bin/server.js",
     "test": "jest",
     "precommit": "lint-staged",
-    "locktt": "locktt"
+    "locktt": "locktt",
+    "test:it": "cypress run 9000"
```
jstastny-cz, line 17: we've removed the part where the app is actually being started before the tests .... please put back the `start-server-with-test` script. Also please make sure that the initial test spec tries to access the exposed port to verify we do things right.

process-migration-service/frontend/cypress/integration/spec.js
```diff
@@ -0,0 +1,5 @@
+describe('First Test', function() {
+  it('Example Test', function() {
+    expect(true).to.equal(true)
```
jstastny-cz, line 3: Please put here a meaningful check of the started web app, at least simple cy.visit().

jstastny-cz (changes requested):
Pom integration seems good, but package.json needs changes.

process-migration-service/frontend/package.json
prashuts: ok

process-migration-service/frontend/cypress/integration/spec.js
jstastny-cz: No, make sure you start PIM UI before this test is executed. Here you just invoke the default endpoint of PIM UI to verify all is configured correctly.

prashuts: then i need config.yml with proper setup..is that okay to commit?

jstastny-cz: You should not need it.
The order of steps we're doing is:
1. navigate to ./process-migration-service/frontend
2. run `npm install`
3. run `npm run build`
4. run `npm run start`

And PIM UI is then started at localhost:3000, I've just verified these steps.

`config.yml` file is required by Thorntail, but here we should be able to test without thorntail integration.

prashuts: ok

jstastny-cz: You might be fighting existing failsafe configuration in the pom.xml - symptom would be `npm test:it` works fine when invoked in terminal, but failing during maven build. Just fyi that it might happen.

prashuts: > You should not need it.
> The order of steps we're doing is:
> 
> 1. navigate to ./process-migration-service/frontend
> 2. run `npm install`
> 3. run `npm run build`
> 4. run `npm run start`
> 
> And PIM UI is then started at localhost:3000, I've just verified these steps.
> 
> `config.yml` file is required by Thorntail, but here we should be able to test without thorntail integration.
i see no 'npm run start' if i add in pom.xml (as execution) , this will be called everytime..pls suggest me on this .

prashuts: start-server-with-test i hope this will suffice :_)

jstastny-cz:
Failing due to no spec.js integration test part of the commit.

prashuts:
> Failing due to no spec.js integration test part of the commit.

Done

jstastny-cz:
jenkins retest this

prashuts:
@jstastny-cz  All the requested changes are committed . I will update incase of any recent changes in master. Otherwise let me know any other changes are required.

jstastny-cz:
See the logs, cypress run failed. Unfortunately green build does not mean correct in this case.

There are missing deps in the node it seems, unless this can be worked around, we might need to run cypress tests using their docker container instead - so that we don't need to care about dependencies.

prashuts:
Obsolete PR , will be creating new one

PR #2421, opened 2021-03-09
[JBPM-9652] Disable http if port specified is invalid

fjtirado:
**JIRA**: _(please edit the JIRA link if it exists)_ 

[link](https://issues.redhat.com/browse/JBPM-9652)


kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -202,6 +202,8 @@ public void start(String host,
                            port.toString());
         System.setProperty(ROUTER_PORT_TLS,
                            portTls.toString());
+        
+        isHttpEnabled = port < 0;
```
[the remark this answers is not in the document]
sutaakar: A nitpick, would it have sense to convert it to (static) method isHttpEnabled so same condition could be reused in KieServerInfoHandler? That would reduce duplication and help keeping consistency.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -271,9 +273,13 @@ public void start(String host,
...
-        Undertow.Builder undertowBuilder = Undertow.builder()
-                .addHttpListener(port,
-                                 host);
+        Undertow.Builder undertowBuilder = Undertow.builder();
+        
+        
+        
+        if (isHttpEnabled) {
+            undertowBuilder.addHttpListener(port, host);
+        }
```
sutaakar, line 282: I think we can show warning in case both HTTP and TLS are disabled, as Smart router doesn't expose any port in such case.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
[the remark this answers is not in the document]
fjtirado: problem is that code is messy, port is used to set system property, so it does not make sense to read system property again in the instance scope (and also the comparison is different, you want to check that is enabled, as you realized and I make a mistake).
In static context the property is only read in one place and the comparison is < 0.
Thats why I did the comparison twice

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
fjtirado: I would say that in that case we should throw exception, since the server is useless, wdyt?

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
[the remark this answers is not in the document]
sutaakar: ok

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
sutaakar: throwing exception is fine for me too

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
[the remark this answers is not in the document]
fjtirado: I will add static method just for the comparison (when githun allow me to push)

sutaakar (approved):
Looks good to me.
Works as expected, in case HTTP port is set for example to 0 then it is disabled. In such case HTTPS is used (if it is defined).
In case only HTTPS is activated then HTTPS is used to register router in the Business central.

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -284,29 +290,28 @@ public void start(String host,
                                                                host,
                                                                sslContext);
         }
-
+        
+        if (!isHttpEnabled && !TLS_ENABLED) {
+            throw new IllegalStateException ("No port has been configured, review  "+ROUTER_PORT+ " and "+ROUTER_PORT_TLS+ " system properties");
```
sutaakar, line 295: Nitpick:
The error is actually thrown in case the HTTP port is explicitly set to <=0 and system property `org.kie.server.router.tls.keystore` is not defined. The exception text could be misleading.
What about something like:
`HTTP listener was disabled (by setting HTTP port to 0 or lower ) and TLS wasn't configured, no listener is available to handle requests.`

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
```diff
@@ -321,14 +328,28 @@ public void stop(boolean clean) {
...
+    
+    private void logServerInfo(String prefix, String host, int port, int portTls) {
+        StringBuilder sb = new StringBuilder(prefix);
+        if (isHttpEnabled) {
+            sb.append(host).append(':').append(port);
+        }
+        if (TLS_ENABLED) {
+            sb.append(host).append(':').append(portTls).append(" (TLS)");
+        }
+        log.info(sb);
```
afalhambra, line 350: if both `http` and `https/tls` are enabled, the output shown by this log might be a little difficult to read - am also missing the `date` as part of the output in the original code though it might be redundant as the log may display the date.

afalhambra (commented):
Look good to me, just one small and minor comment.
Well done!

kie-server-parent/kie-server-router/kie-server-router-proxy/src/main/java/org/kie/server/router/KieServerRouter.java
fjtirado: I slightly change the log to avoid that issue. Date was clearly redundant with log date and I intentionally remove it

afalhambra (approved):
Great job

fjtirado:
jenkins retest this

PR #2431, opened 2021-03-19
[JBPM-9663] Adding integration test for JBPM-9647 - Multiple humantask email notifications are sent

afalhambra:
**[JBPM-9663](https://issues.redhat.com/browse/JBPM-9663)**: Adding integration test for JBPM-9647 - Multiple humantask email notifications are sent

Depends on https://github.com/kiegroup/jbpm/pull/1893


sutaakar:
Why is this scenarios tested here, wouldn't it be enough to test it in jBPM module?
Or is it needed to check functionality on different application servers?

afalhambra:
Yes, you're right, not really needed in this repo.
Initially I planned to test the functionality of timers/notifications when server is bounced, but since it's managed by cargo, there is no chance to achieve that.
Moving them to jbpm repo.
Thanks @sutaakar

afalhambra:
Reopen - Need to double check since this functionality is only available with EJB timers, if I'm not wrong

fjtirado:
> Why is this scenarios tested here, wouldn't it be enough to test it in jBPM module?
> Or is it needed to check functionality on different application servers?

@sutaakar This needs to be done on integration because, although code is on jbpm, this needs to be tested using kie server

sutaakar:
Ok, in such case, would it be possible to reduce the thread sleep times?
Currently there is more than 30 seconds sleep time, that looks like too much for me for one test.

afalhambra:
> Ok, in such case, would it be possible to reduce the thread sleep times?
> Currently there is more than 30 seconds sleep time, that looks like too much for me for one test.

OK, I will see what I can do. Though not easy as there are several notifications involved in this process, some of which are repeatable ones.

afalhambra:
Closing this PR. Follow up on https://github.com/kiegroup/droolsjbpm-integration/pull/2434

PR #2441, opened 2021-03-26
rename sync process api

elguardian:
depends on https://github.com/kiegroup/jbpm/pull/1904

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/rest/RestURI.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/rest/RestURI.java
@@ -87,7 +87,7 @@ public class RestURI {
     // uris
     // process related prefixed by PROCESS_URI
     public static final String START_PROCESS_POST_URI = "{" + PROCESS_ID + "}/instances";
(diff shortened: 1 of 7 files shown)


kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/rest/RestURI.java
```diff
@@ -87,7 +87,7 @@
     // uris
     // process related prefixed by PROCESS_URI
     public static final String START_PROCESS_POST_URI = "{" + PROCESS_ID + "}/instances";
-    public static final String START_SYNC_PROCESS_POST_URI = "{" + PROCESS_ID + "}/syncInstances";
+    public static final String COMPUTE_PROCESS_OUTCOME_POST_URI = "{" + PROCESS_ID + "}/syncInstances";
```
gmunozfe, line 90: Shouldn't it be more accurate to use "computedInstances" in the path instead of syncInstances, with the new terminology?

gmunozfe (approved):
Looks good, just a question regarding REST URL

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/rest/RestURI.java
elguardian: done

PR #2455, opened 2021-04-14
[JBPM-9699] Register workitemhandler when Kafka extension is enabled

fjtirado:
If Kafka extension is enabled, When a kjar is deployed, it will automatically registered a send task work item handler.

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9699)

**Merge with**
https://github.com/kiegroup/jbpm/pull/1918
https://github.com/kiegroup/droolsjbpm-knowledge/pull/515

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
@@ -38,6 +38,7 @@ import org.jbpm.services.api.DeploymentEventListener;
 import org.jbpm.services.api.DeploymentService;
 import org.jbpm.services.api.ListenerSupport;
 import org.jbpm.services.api.ProcessService;
+import org.kie.internal.runtime.manager.deploy.DeploymentDescriptorManager;
 import org.kie.server.api.KieServerConstants;
 import org.kie.server.services.api.KieContainerInstance;
 import org.kie.server.services.api.KieServerExtension;
@@ -83,7 +84,10 @@ public class KafkaServerExtension implements KieServerExtension, DeploymentEvent
             logger.warn("Kafka extension already initialized");
             return;
         }
-
+        
+        DeploymentDescriptorManager.addDescriptorLocation(
+                "classpath:/META-INF/kafka-deployment-descriptor-defaults.xml");
+        
         KieServerExtension jbpmExt = registry.getServerExtension(JbpmKieServerExtension.EXTENSION_NAME);
         if (jbpmExt == null) {
             logger.warn("Extension " + JbpmKieServerExtension.EXTENSION_NAME + " is required");
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerProducer.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerProducer.java
@@ -20,12 +20,13 @@ import java.util.function.Supplier;
 
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.jbpm.bpmn2.handler.SendTaskHandler;
+import org.jbpm.runtime.manager.impl.SimpleRegisterableItemsFactory;
 import org.jbpm.services.api.DeploymentEvent;
 import org.kie.api.event.process.DefaultProcessEventListener;
 import org.kie.api.event.process.MessageEvent;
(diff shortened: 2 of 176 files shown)


kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerProducer.java
```diff
@@ -94,10 +95,13 @@ private void sendEvent(ProcessInstance processInstance,
...
     }
-
+    
+    
     public void activate(DeploymentEvent event) {
-        ((InternalRegisterableItemsFactory) ((InternalRuntimeManager) event.getDeployedUnit().getRuntimeManager())
-                .getEnvironment().getRegisterableItemsFactory()).addProcessListener(this);
+        SimpleRegisterableItemsFactory itemsFactory = (SimpleRegisterableItemsFactory) ((InternalRuntimeManager) event
+                .getDeployedUnit().getRuntimeManager()).getEnvironment().getRegisterableItemsFactory();
+        itemsFactory.addWorkItemHandler("Send Task", SendTaskHandler.class);
```
gmunozfe, line 103: The problem with this approach is that using the default _SendTaskHandler_ class, only String messages could be sent:
https://github.com/kiegroup/jbpm/blob/62f92c98f5d3f68f08fd519638259be97aa2d574/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/SendTaskHandler.java#L30

Then, this kind of tests with a custom message would fail:
https://github.com/kiegroup/droolsjbpm-integration/blob/dd06377f866ba814b45d9d654c64c3f754084622/kie-spring-boot/kie-spring-boot-samples/kie-server-spring-boot-sample/src/test/java/org/kie/server/springboot/samples/KafkaProducerHappyPathTest.java#L195

wdyt?

gmunozfe (commented):
A question regarding non string messages that would fail with the default SendTaskHandler

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerProducer.java
fjtirado: I will change the default handler to not assume string

elguardian:
wrong approach. Use deployment descriptor for this. This way you could override the Send Task if needed in case the use want to use their own stuff

fjtirado:
> wrong approach. Use deployment descriptor for this. This way you could override the Send Task if needed (by deployment) in case the use want to use their own stuff

I  think user definitions has preference over itemsFactory (basically because in the code they are added later). So the Send Task will only be used if user has not defined anything, which is what we want.

fjtirado:
jenkins retest this (to take updated example)

fjtirado:
Using prometeus extension approach

fjtirado:
@gmunozfe @elguardian  I think this is ready to go

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerExtension.java
```diff
@@ -83,7 +84,13 @@ public void init(KieServerImpl kieServer, KieServerRegistry registry) {
             logger.warn("Kafka extension already initialized");
             return;
         }
-
+        
+        // register work item handler
+        String desc = System.getProperty(DESCRIPTOR);
+        if(desc == null ) {
+            System.setProperty(DESCRIPTOR, "classpath:/META-INF/kafka-deployment-descriptor-defaults.xml");
```
elguardian, line 91: you are overriding this property therfore prometheus won't go in.

fjtirado: thats the property that is used by kie server here https://github.com/kiegroup/droolsjbpm-knowledge/blob/master/kie-internal/src/main/java/org/kie/internal/runtime/manager/deploy/DeploymentDescriptorManager.java#L61
I think is better to use initial proposal and avoid this clash

fjtirado: Initial proposal depends on JBPM behaviour while template is kie server one, so extending DeploymentDescriptorManager capabilities (adding add static method) to allow several templated

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

elguardian (approved):

gmunozfe (approved):
Looks good to me, splendid work @fjtirado !

PR #2478, opened 2021-05-11
[JBPM-9729] KIE-Server template is removed after a ping failure

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9729
adding kie server instance online flag and modifying alive ping so the server can have that possibility of comming back alive if there is a network glitch during ping is alive.


elguardian:
@sutaakar  it should be fixed now. Some implementations keep in memory mutable data. the copy constructor should fix that.

elguardian:
Jenkins retest this please

sutaakar (approved):
Looks good to me, seems to work as expected.

sutaakar:
@elguardian Can you please take a look on the test failures?

sutaakar:
Can you also provide some test coverage for healthcheck?

elguardian:
Jenkins retest this please.

PR #2486, opened 2021-05-19
[JBPM-9744] Avoid duplication notification in BC with KieServerInstanceManager

elguardian:


elguardian:
Jenkins retest this please

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/controller/DefaultRestControllerImpl.java
```diff
@@ -146,16 +146,21 @@ public KieServerSetup connect(KieServerInfo serverInfo) {
...
+            KieServerSetup kieServerSetup = null;
             for (String controllerUrl : controllers) {
 
                 if (controllerUrl != null && !controllerUrl.isEmpty()) {
-                    KieServerSetup kieServerSetup = connectToSingleController(serverInfo, config, controllerUrl);
-                    if (kieServerSetup != null && kieServerSetup.hasNoErrors()) {
-                        return kieServerSetup;
+                    KieServerSetup kieServerSetupController = connectToSingleController(serverInfo, config, controllerUrl);
+                    if (kieServerSetup == null && kieServerSetupController != null && kieServerSetupController.hasNoErrors()) {
+                        kieServerSetup = kieServerSetupController;
```
gmunozfe, line 155: With this change, it will try to connect to all controllers, returning the first succeeded one. In other implementations like `WebSocketKieServerControllerImpl` it returns with the first connected without errors, should it also be changed?

gmunozfe (commented):
Looks good to me, pretty awesome work @elguardian !
Just a couple of questions to be clarified.

kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/controller/DefaultRestControllerImpl.java
elguardian: not really. websockets has a different use case in this regard as it is a connected oriented protocol. So that would require another jira.

kie-server-parent/kie-server-controller/kie-server-controller-api/src/main/java/org/kie/server/controller/api/model/runtime/Container.java
```diff
@@ -54,7 +54,7 @@ public Container( final String containerSpecId,
         this.serverInstanceId = serverInstanceKey.getServerInstanceId();
         this.messages.addAll( messages );
         this.resolvedReleasedId = resolvedReleasedId;
-
+        setUrl(url);
```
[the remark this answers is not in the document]
elguardian: it is not exactly the same. The url comes from the server instance key and not the constructor

elguardian:
failures not related.

elguardian:
Jenkins retest this please

PR #2247, opened 2020-09-23
DROOLS-5663 align kie-maven-plugin default for generateModel config flag

tarilabs:
**JIRA**: https://issues.redhat.com/browse/DROOLS-5663

**referenced Pull Requests**: none

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #2505, opened 2021-06-10
[DROOLS-6292] perform in memory compilation of generated classes in k…

mariofusco:
…ie-maven-plugin

**JIRA**: https://issues.redhat.com/browse/DROOLS-6292

This is a first iteration to perform in memory compilation of generated sources. In a subsequent step I will consolidate all the mojos generating sources into a single one.

@gitgabrio as anticipated I gave a try creating the kjar for `trusty-drools-kjar` and it took 1:39 mins on my machine instead of the original 10+. It seems to me that everything has been generated correctly in that kjar, but please give a second look.

I also added a maven parameter `dumpGeneratedSources` (disabled by default) to optionally dump the generated sources which could be useful for debugging.

### DIFF
--- a/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/pom.xml
+++ b/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/pom.xml
@@ -30,6 +30,10 @@
 
   <packaging>kjar</packaging>
 
+  <properties>
+    <generateDMNModel>YES</generateDMNModel>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.kie</groupId>
--- a/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/src/test/java-filtered/org/kie/maven/plugin/ittests/KJarWithDMNIntegrationTestIT.java
+++ b/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/src/test/java-filtered/org/kie/maven/plugin/ittests/KJarWithDMNIntegrationTestIT.java
@@ -52,6 +52,10 @@ public class KJarWithDMNIntegrationTestIT {
     public void testComplexDMNModel() throws Exception {
         final URL targetLocation = KJarWithDMNIntegrationTestIT.class.getProtectionDomain().getCodeSource().getLocation();
         final KieSession kieSession = ITTestsUtils.getKieSession(targetLocation, GAV_ARTIFACT_ID, GAV_VERSION, KBASE_NAME);
+
+        // verify DMN file is present
+        assertThat(getClass().getClassLoader().getResourceAsStream("META-INF/kie/dmn")).isNotNull();
+
         try {
             final DMNRuntime dmnRuntime = kieSession.getKieRuntime(DMNRuntime.class);
             final DMNModel dmnModel = dmnRuntime.getModel("http://www.trisotech.com/definitions/_3068644b-d2c7-4b81-ab9d-64f011f81f47",
--- a/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/src/test/java-filtered/org/kie/maven/plugin/ittests/KJarWithDMNIntegrationTestIT.java
+++ b/kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/src/test/java-filtered/org/kie/maven/plugin/ittests/KJarWithDMNIntegrationTestIT.java
@@ -36,9 +36,9 @@ import org.kie.dmn.api.core.DMNModel;
 import org.kie.dmn.api.core.DMNResult;
 import org.kie.dmn.api.core.DMNRuntime;
 import org.kie.dmn.core.api.DMNFactory;
+import org.kie.dmn.core.compiler.execmodelbased.DMNRuleClassFile;
 
 import static org.assertj.core.api.Assertions.assertThat;
-
 
 public class KJarWithDMNIntegrationTestIT {
 
@@ -54,7 +54,7 @@ public class KJarWithDMNIntegrationTestIT {
         final KieSession kieSession = ITTestsUtils.getKieSession(targetLocation, GAV_ARTIFACT_ID, GAV_VERSION, KBASE_NAME);
 
         // verify DMN file is present
-        assertThat(getClass().getClassLoader().getResourceAsStream("META-INF/kie/dmn")).isNotNull();
+        assertThat(getClass().getClassLoader().getResourceAsStream(DMNRuleClassFile.RULE_CLASS_FILE_NAME)).isNotNull();
 
         try {
             final DMNRuntime dmnRuntime = kieSession.getKieRuntime(DMNRuntime.class);
--- a/kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
+++ b/kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
@@ -26,8 +26,8 @@ import java.util.stream.Collectors;
 
 public abstract class AbstractKieMojo extends AbstractMojo {
 
-    @Parameter(property = "dumpGeneratedSources", defaultValue = "false")
-    protected boolean dumpGeneratedSources;
+    @Parameter(property = "dumpKieSourcesFolder", defaultValue = "")
+    protected String dumpKieSourcesFolder;
 
     @Parameter(property = "javaCompiler", defaultValue = "ecj")
     private String javaCompiler;
--- a/kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
+++ b/kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
@@ -18,6 +18,7 @@ package org.kie.maven.plugin;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
+import org.kie.memorycompiler.JavaConfiguration;
 
 import java.util.List;
 import java.util.Map;
@@ -25,11 +26,14 @@ import java.util.stream.Collectors;
 
 public abstract class AbstractKieMojo extends AbstractMojo {
 
-    @Parameter(property = "dumpGeneratedSources", defaultValue = "false") // DROOLS-5663 align kie-maven-plugin default value for generateModel configuration flag
-    private boolean dumpGeneratedSources;
+    @Parameter(property = "dumpGeneratedSources", defaultValue = "false")
+    protected boolean dumpGeneratedSources;
 
-    protected boolean isDumpGeneratedSources() {
-        return dumpGeneratedSources;
+    @Parameter(property = "javaCompiler", defaultValue = "ecj")
+    private String javaCompiler;
+
+    protected JavaConfiguration.CompilerType getCompilerType() {
+        return javaCompiler.equalsIgnoreCase("native") ? JavaConfiguration.CompilerType.NATIVE : JavaConfiguration.CompilerType.ECLIPSE;
     }
 
     protected void setSystemProperties(Map<String, String> properties) {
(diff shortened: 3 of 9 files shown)


kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
```diff
@@ -0,0 +1,125 @@
...
+
+        Map<String, byte[]> compiledClassesMap = KieMemoryCompiler.compileNoLoad(classNameSourceMap, projectClassLoader, javaCompilerSettings, JavaConfiguration.CompilerType.ECLIPSE);
+
+        for (Map.Entry<String, byte[]> entry : compiledClassesMap.entrySet()) {
+            Path packagesDestinationPath = Paths.get(targetDirectory.getPath(), "classes", entry.getKey().replace('.', '/') + ".class");
+            writeFile(packagesDestinationPath, entry.getValue());
+        }
+    }
+
+    private static void dumpGeneratedSources(File targetDirectory, Map<String, String> classNameSourceMap) {
```
danielezonca, line 81: This code is not really equivalent to the old one: now all sources are generated in `generated-sources` folder without the additional specific folder (like `drools` or `alpha-network-compiler`).
We probably need to document this change because if a user has some post postprocessor that assume sources to be there this could break.
In addition, is this path a valid maven generated-sources content? Do we need `main/src` subfolder?

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
```diff
@@ -0,0 +1,125 @@
...
+    public static URLClassLoader getProjectClassLoader(MavenProject project, File outputDirectory, JavaCompilerSettings javaCompilerSettings) {
+        try {
+            Set<URL> urls = new HashSet<>();
+            for (String element : project.getCompileClasspathElements()) {
+                File file = new File(element);
+                javaCompilerSettings.addClasspath(file);
+                urls.add(file.toURI().toURL());
+            }
+
+            project.setArtifactFilter(new CumulativeScopeArtifactFilter(Arrays.asList("compile", "runtime")));
```
danielezonca, line 51: What about `provided`?

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateDMNModelMojo.java
```diff
@@ -94,22 +99,14 @@ private void generateDMNModel() throws MojoExecutionException {
...
-
-                    compiledClassNames.add(getCompiledClassName(fileNameRelative));
-
-                    final Path newFile = Paths.get(targetDirectory.getPath(),
-                                                   droolsModelCompilerPath,
-                                                   fileNameRelative.toString());
-
-                    createInvokerSourceFile(newFile, generatedFile.getSourceContent());
+                    getLog().info("Generating new DMN file: " + generatedFile);
+                    classNameSourceMap.put(getCompiledClassName(fileNameRelative), generatedFile.getSourceContent());
```
danielezonca, line 109: Are you sure this code is equivalent to the old one?
I see for example the code of method `createDMNFile` and I'm not sure the new code is equivalent

kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
```diff
@@ -24,6 +25,13 @@
 
 public abstract class AbstractKieMojo extends AbstractMojo {
 
+    @Parameter(property = "dumpGeneratedSources", defaultValue = "false") // DROOLS-5663 align kie-maven-plugin default value for generateModel configuration flag
```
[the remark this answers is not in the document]
lucamolteni: This comment was there before - was added here https://github.com/kiegroup/droolsjbpm-integration/pull/2247/files#diff-56c6e735378a2a62ff67cab2bced2750dd9016df3695b008d1825a6f4ecdb4e2R97

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static void compileAndWriteClasses(File targetDirectory, ClassLoader projectClassLoader, JavaCompilerSettings javaCompilerSettings, Map<String, String> classNameSourceMap, boolean dumpGeneratedSources) {
+        if (dumpGeneratedSources) {
+            dumpGeneratedSources(targetDirectory, classNameSourceMap);
+        }
+
+        Map<String, byte[]> compiledClassesMap = KieMemoryCompiler.compileNoLoad(classNameSourceMap, projectClassLoader, javaCompilerSettings, JavaConfiguration.CompilerType.ECLIPSE);
+
+        for (Map.Entry<String, byte[]> entry : compiledClassesMap.entrySet()) {
+            Path packagesDestinationPath = Paths.get(targetDirectory.getPath(), "classes", entry.getKey().replace('.', '/') + ".class");
```
gitgabrio, line 76: hi @mariofusco 
Could you please replace "/" with OS-specific path separator, all over this class ?  Thanks!

kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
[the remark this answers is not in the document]
mariofusco: Wild copy/paste strikes again :)

kie-maven-plugin/src/main/java/org/kie/maven/plugin/AbstractKieMojo.java
```diff
@@ -24,6 +25,13 @@
...
+    private boolean dumpGeneratedSources;
```
gitgabrio, line 29: @mariofusco 
Out of curiosity: why not making `dumpGeneratedSources` protected, thus avoiding the need for getter ?

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void compileAndWriteClasses(File targetDirectory, ClassLoader projectClassLoader, JavaCompilerSettings javaCompilerSettings, Map<String, String> classNameSourceMap, boolean dumpGeneratedSources) {
+        if (dumpGeneratedSources) {
+            dumpGeneratedSources(targetDirectory, classNameSourceMap);
+        }
+
+        Map<String, byte[]> compiledClassesMap = KieMemoryCompiler.compileNoLoad(classNameSourceMap, projectClassLoader, javaCompilerSettings, JavaConfiguration.CompilerType.ECLIPSE);
```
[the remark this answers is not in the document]
mariofusco: There are 2 problems here

1. ecj is much faster than the native java compiler (for some reasons)
2. (more important) [we never managed to make the native java compiler to work with JPMS](https://github.com/kiegroup/drools/blob/master/kie-memory-compiler/src/main/java/org/kie/memorycompiler/JavaCompilerFactory.java#L42) (Java 9+)

I could make this configurable but I'd leave ecj as the default.

gitgabrio (changes requested):
HI @mariofusco , thanks for the PR: beside anything else, it seems an overall improvement toward "unification" of different mojos.
I have just two small observations
1) usage of "/" as path separator: I think it is much safer to use OS-dependent one
2) there is not a single unit test on `GenerateCodeUtil`: if something broke (e.g. on filesystem access) the only way to find out is to debug the integration tests, that is extremely cumbersome

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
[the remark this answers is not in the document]
danielezonca: Fine for me to keep ecj as default but at least we let user decide what to do like we do also for drools compiler :)

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
mariofusco: This is not meant to be equivalent to the old code. We don't need anymore to dump the generated java sources since we're compiling them in memory, so by default we're not writing them anymore. I added a parameter to optionally dumping the generated sources only for debugging purposes.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
mariofusco: I honestly don't know, I didn't change this, just moved this line as it was from mojos to this utility class.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateDMNModelMojo.java
mariofusco: Sorry, I'm not sure what's the problem here. I'm actually replacing that `createInvokerSourceFile` method that dumped the source file with mine that put it in the map. Are you confusing the methods? Am I missing something?

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static String toClassName(String source) {
+        if (source.startsWith("./")) {
+            source = source.substring(2);
+        }
+        if (source.endsWith(".java")) {
+            source = source.substring(0, source.length()-5);
+        }
+        return source.replace('/', '.');
+    }
```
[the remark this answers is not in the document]
mariofusco: The methods are similar, but not identical. In particular from time to time the file name is passed it with a leading `./` and I haven't been able to understand why (some maven issue?), so I'm just checking this detail and eventually removing it. I wouldn't like to propagate this behavior into the other method.

kie-maven-plugin/src/main/java/org/kie/maven/plugin/GenerateCodeUtil.java
danielezonca: I understand the now the logic is different, my comment is related the fact that even if user enable this flag, this doesn't reproduce the same output of the old version and secondly just to be sure that if user enable this feature the output in `target` is not broken (like that if you run again `mvn compile` without `clean` it produce an error)

lucamolteni (approved):

kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/pom.xml
```diff
@@ -30,6 +30,10 @@
 
   <packaging>kjar</packaging>
 
+  <properties>
+    <generateDMNModel>YES</generateDMNModel>
```
tarilabs, line 34: thank you

kie-maven-plugin/src/it/kie-maven-plugin-test-kjar-7-exec-model/src/test/java-filtered/org/kie/maven/plugin/ittests/KJarWithDMNIntegrationTestIT.java
```diff
@@ -52,6 +52,10 @@
     public void testComplexDMNModel() throws Exception {
         final URL targetLocation = KJarWithDMNIntegrationTestIT.class.getProtectionDomain().getCodeSource().getLocation();
         final KieSession kieSession = ITTestsUtils.getKieSession(targetLocation, GAV_ARTIFACT_ID, GAV_VERSION, KBASE_NAME);
+
+        // verify DMN file is present
+        assertThat(getClass().getClassLoader().getResourceAsStream("META-INF/kie/dmn")).isNotNull();
```
danielezonca, line 57: What about use [DMNRuleClassFile.RULE_CLASS_FILE_NAME](https://github.com/kiegroup/drools/blob/master/kie-dmn/kie-dmn-core/src/main/java/org/kie/dmn/core/compiler/execmodelbased/DMNRuleClassFile.java#L15) instead of copy the string?

danielezonca (approved):

mariofusco:
jenkins retest this

mariofusco:
All test failures are caused by OSGi integration and not related to this.

PR #2511, opened 2021-06-14
[windows_codegenpath] test output path on windows

gitgabrio:
Testing source output on windows

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -6,9 +6,10 @@ jobs:
   build-chain:
     strategy:
       matrix:
+        os: [ubuntu-18.04, windows-latest]
         java-version: [8, 11]
       fail-fast: false
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
     name: Maven Build
     steps:
       - name: Free disk space
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -13,6 +13,7 @@ jobs:
     name: Maven Build
     steps:
       - name: Free disk space
+        if: ${{ matrix.os == 'ubuntu-18.04' }}
         shell: bash
         run: |
           sudo rm -rf /usr/share/dotnet
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -12,6 +12,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     name: Maven Build
     steps:
+      - name: Support longpaths
+            if: ${{ matrix.os == 'windows-latest' }}
+            run: git config --system core.longpaths true
       - name: Free disk space
         if: ${{ matrix.os == 'ubuntu-18.04' }}
         shell: bash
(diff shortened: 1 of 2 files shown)


gitgabrio:
Jenkins retest this

.github/workflows/pull_request.yml
```diff
@@ -6,12 +6,14 @@ jobs:
...
     strategy:
       matrix:
+        os: [ubuntu-18.04, windows-latest]
         java-version: [8, 11]
       fail-fast: false
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
     name: Maven Build
     steps:
       - name: Free disk space
```
Ginxo, line 15: An additional step should be added here as we do for drools https://github.com/kiegroup/drools/blob/7c64a322182f275c811a658763226858b2efa796/.github/workflows/pull_request.yml#L15
```suggestion
      - name: Support longpaths
        if: ${{ matrix.os == 'windows-latest' }}
        run: git config --system core.longpaths true
      - name: Free disk space
```

Ginxo (changes requested):
Please check

.github/workflows/pull_request.yml
gitgabrio: @Ginxo thx!

Ginxo:
@gitgabrio I'm afraid we have the same issue we had in the past
```
2021-06-15T07:31:40.1651042Z Caused by: java.io.IOException: Cannot run program "C:\hostedtoolcache\windows\jdk\8.0.292\x64\jre\..\bin\javadoc.exe" (in directory "D:\a\droolsjbpm-integration\droolsjbpm-integration\kiegroup\droolsjbpm_integration\kie-server-parent\kie-server-services\kie-server-services-task-assigning\kie-server-services-task-assigning-user-system\kie-server-services-task-assigning-user-system-api\target\apidocs"): CreateProcess error=267, The directory name is invalid
```

gitgabrio:
@Ginxo 
Yup, I see. Reason for that PR was exactly to verify what may happen in windows machine.
Some failures may be ignored (e.g. javadoc creation may happen just inside our "build" machines) but other code-path is expected to run on customer machine (e.g. maven-plugin execution).
Main issue is that currently it seems we are not verifying what happen on windows, and it is a bit of concern from my point of view. I think a more structured approach is needed.
@mariofusco @danielezonca   WDYT ?

danielezonca:
> @Ginxo
> Yup, I see. Reason for that PR was exactly to verify what may happen in windows machine.
> Some failures may be ignored (e.g. javadoc creation may happen just inside our "build" machines) but other code-path is expected to run on customer machine (e.g. maven-plugin execution).
> Main issue is that currently it seems we are not verifying what happen on windows, and it is a bit of concern from my point of view. I think a more structured approach is needed.
> @mariofusco @danielezonca WDYT ?

I think making the whole `-integration` repo compatible with a windows build will require quite a lot of work, it is probably easier create a specific profile and configure a windows job that uses that profile. For example I agree that `kie-maven-plugin` should be tested also with Windows.

mareknovotny:
Could we close this PR?

danielezonca:
I think so, it is quite old and the repo is sort of EOL

PR #2531, opened 2021-07-01
[7.52.X] [JBPM-9805] Fixing NPE on consumer

fjtirado:
**JIRA**
https://issues.redhat.com/browse/JBPM-9805


elguardian:
@gmunozfe @fjtirado have a look at those test. in this case seems related. Those test usually fail randomly. I know @gmunozfe has done some enhacements. any clue ?

fjtirado:
@elguardian According to the the traces, looks like test are failing with timeout because kafka broker is not available, so (with reservations) I would say it is unrelated . Im going to run this again. Jenkins test this

gmunozfe (approved):
Looks good to you, great work @fjtirado !
Just wait for CI to be green (locally, all kafka sb tests passed)

gmunozfe:
Failures are due to Jenkins takes the jbpm branch fjtirado/7.52.x which doesn't contain this commit https://github.com/kiegroup/jbpm/commit/054b2ab05a7b8902b04503231bd26010b24de1d7 
That's the reason for looking at the default bootstrap server.

PR #2595, opened 2021-09-08
DROOLS-6592 reproducer ONLY

tarilabs:
https://issues.redhat.com/browse/DROOLS-6592

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

PR #2599, opened 2021-09-09
[DROOLS-6592] Kie Server Marshalling changes Pojo field UpperCase begin and JSON

elguardian:
Jira: https://issues.redhat.com/browse/DROOLS-6592

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java
@@ -60,6 +60,7 @@ public class KieServerConstants {
     public static final String KIE_SERVER_MODE = "org.kie.server.mode";
     public static final String KIE_SERVER_INCLUDE_STACKTRACE = "org.kie.server.stacktrace.included";
     public static final String KIE_SERVER_STRICT_ID_FORMAT = "org.kie.server.strict.id.format";
+    public static final String KIE_SERVER_STRICT_JAVABEANS_SERIALIZERS = "org.kie.server.strict.javaBeans.serializers";
     public static final String KIE_SERVER_STRICT_JAXB_FORMAT = "org.kie.server.strict.jaxb.format";
     public static final String KIE_SERVER_IMAGESERVICE_MAX_NODES = "org.kie.server.service.image.max_nodes";
     public static final String KIE_SERVER_REST_MODE_READONLY = "org.kie.server.rest.mode.readonly";
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
@@ -95,14 +95,17 @@ public class JSONMarshaller implements Marshaller {
 
     private static final Logger logger = LoggerFactory.getLogger(MarshallerFactory.class);
 
-    private static final boolean STRICT_ID_FORMAT = Boolean.parseBoolean(
-            KieServerConstants.KIE_SERVER_STRICT_ID_FORMAT);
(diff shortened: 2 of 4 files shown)


tarilabs:
Can we please make sure the backward compatibility behaviour of the payload is respected ?

On suggestion of @tkobayas to consider B/C, I've added ffe6401 in https://github.com/kiegroup/droolsjbpm-integration/pull/2595
The new test ensures specifically that should ensure the expected payload used in previous releases is valid.

eg: please also cherry-pick https://github.com/tarilabs/droolsjbpm-integration/commit/ffe64011d91b377b10856d00fc160742aa249c6d here as needed.

elguardian:
Hi @tarilabs it is ensured. only activate the strictness under the system flag.
I will add your test.

elguardian:
instead of going overkilling (as flag is test with / without it in the full suite) just added a json marshaller test

kie-server-parent/kie-server-tests/kie-server-integ-tests-drools/pom.xml
```diff
@@ -152,6 +152,7 @@
         <configuration>
           <container>
             <systemProperties>
+              <org.kie.server.strict.javaBeans.serializers>true</org.kie.server.strict.javaBeans.serializers>
```
tarilabs, line 155: If I'm not mistaken, this changes ALL the kie-server Drools integration test to use the new configuration flag which was needed for JBPM-9784.

I am not sure this is desirable, that flag is for jBPM/RHPAM and it's best if only pertaining to the original scope of specific customer request.

I would keep the Drools Integration Test as we provided with
 - https://github.com/kiegroup/droolsjbpm-integration/pull/2595

and without the need to change the configuration behaviour of Drools, DMN, etc. kie-server integratio tests.

my2c
/cc @tkobayas @danielezonca @mariofusco @lucamolteni

gmunozfe: as an integration test, we cannot have both behaviors based on a property in the same module. Perhaps it's enough to test it at marshaller, removing these integration tests if you don't want to alter former behavior

tarilabs: the Kie Server Drools IntegrationTest we introduced on Drools side with 
 - https://github.com/kiegroup/droolsjbpm-integration/pull/2595

are needed because:
 - we did not have a test with a Business / Domain POJO with `CapitalizedField`
 - we did not have a test for ensuring non-regression of the JSON payload, as detailed in the original JIRA https://issues.redhat.com/browse/DROOLS-6592

I still am under the impression this flag would not be needed for Drools kie server IT tests (imho)

elguardian: @tarilabs it is not relevant whether that flag is active or not as long as we have tests covering both cases (which is the case). 
The flag is about kie server not just jBPM so it covers every aspect of kiesever not just the engine.

tarilabs: Let me try to bring a specific concern, with this highlighted line.

We always ask SA, Fields, and Users, to log-dump the Kie Server Client Java API request/response first, if they want to directly interact with a kie-server without using our java Client. ie: log-dump the request/response, and modify as needed from a base, not build a request payload from scratch.

If this line is merged, then people using these ITs in the directory
```
kie-server-parent/kie-server-tests/kie-server-integ-tests-drools
```
which is about Drools kie-server IT tests,
those using these ITs for their reference would start to use a payload shape which is NOT the default (I am assuming most of Users would NOT set the `org.kie.server.strict.javaBeans.serializers` overriding the default)

Hope this clarifies the concern, and the reasons of the original IT tests inside 
- https://github.com/kiegroup/droolsjbpm-integration/pull/2595

Thank you for considering this aspect too

elguardian: removing your reproducer then as JSON Marshaller is already having those tests in place.

gmunozfe (approved):
Looks good to me, outstanding work @elguardian !

PR #2606, opened 2021-09-15
[JBPM-9898] Container upgrade is not properly handled by Kafka server

fjtirado:
If message or signal is associated with start node, consider that two deployments with the same version are the same one and use the one that is deployed the latest

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9898)

**Referenced pull request**
https://github.com/kiegroup/droolsjbpm-knowledge/pull/552

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
@@ -14,30 +14,36 @@
 */
 package org.kie.server.services.jbpm.kafka;
 
-import java.util.ArrayList;
+import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.processMessages;
+import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.processSignals;
+import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.topicFromSignal;
+
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.function.Consumer;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.jbpm.kie.services.impl.KModuleDeploymentUnit;
 import org.jbpm.services.api.DeploymentEvent;
 import org.jbpm.services.api.model.DeployedAsset;
 import org.jbpm.services.api.model.MessageDesc;
 import org.jbpm.services.api.model.ProcessDefinition;
 import org.jbpm.services.api.model.SignalDesc;
 import org.jbpm.services.api.model.SignalDescBase;
-
-import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.processMessages;
-import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.processSignals;
-import static org.kie.server.services.jbpm.kafka.KafkaServerUtils.topicFromSignal;
+import org.jbpm.workflow.core.node.StartNode;
+import org.kie.api.builder.ReleaseIdComparator.ComparableVersion;
+import org.kie.api.definition.process.Node;
 
 class KafkaServerRegistration {
-
-    private Map<String, Map<SignalDesc, Collection<String>>> topic2Signal = new HashMap<>();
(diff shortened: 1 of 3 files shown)


kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
```diff
@@ -14,30 +14,87 @@
...
+        }
+
+        @Override
+        public boolean equals(Object other) {
+            if (this == other) {
+                return true;
+            }
+            if (other instanceof DeploymentId) {
+                DeploymentId deployment = (DeploymentId) other;
+                return Objects.equal(this.groupId, deployment.groupId) && Objects.equal(this.artifactId, deployment.artifactId);
```
elguardian, line 91: This means will only have one deployment per signal so if we have 2 or more different versions of the same deployment it will happen that only one of them will get the signal processed. keep in mind that a process in version 1 won't go to vesion 2 because they belong to different kie containers, so it does not seems to me like a good idea.

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
```diff
@@ -93,42 +159,42 @@ void forEachMessage(ConsumerRecord<String, byte[]> event, KafkaServerEventProces
...
-    private <T extends SignalDescBase> void addTopics(Map<String, Map<T, Collection<String>>> topic2SignalBase,
-                                                      String deploymentId,
+    private <T extends SignalDescBase> void addTopics(Map<String, Map<T, Map<DeploymentId, String>>> topic2SignalBase,
+                                                      DeploymentId deploymentId,
                                                       Collection<T> signals) {
         for (T signal : signals) {
             if (!signal.getIncomingNodes().isEmpty()) {
                 topic2SignalBase.computeIfAbsent(topicFromSignal(signal), k -> new HashMap<>()).computeIfAbsent(
-                        signal, k -> new ArrayList<>()).add(deploymentId);
+                                                                                                                signal, k -> new LinkedHashMap<>()).put(deploymentId, deploymentId.toString());
```
elguardian, line 182: this map will cause a race condition as the last to be deployed it will the one getting all the attention

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
fjtirado: Since currently we are just notifying deployments (to start process instances) not process instance, the purpose of this JIRA is to avoid starting two process for the "same" deployment. 
https://kie.zulipchat.com/#narrow/stream/232678-jbpm/topic/Container.20upgrade.20and.20Kafka.20message

kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerRegistration.java
fjtirado: thats the purpose, to start process in the latest deployed one.

fjtirado:
Working on different approach

fjtirado:
This depends on later versions being deployed later, this is not neccesarilty true after a restart, so a different approach is needed

Ginxo:
Jenkins retest this

fjtirado:
jenkins retest this

elguardian (approved):
OK.  Just a minor comment related to version comparison. Either you can leave it like this or try to do numeric comparisson.

gmunozfe (changes requested):
It fails the version comparator over 10.
Could you parameterize the test to check these cases:
9.0, 10.0
1.9, 1.10
1.0.9, 1.0.10

gmunozfe (approved):
Looks good to me, nailed down @fjtirado !

fjtirado:
jenkins retes this

fjtirado:
jenkins (re)test this

fjtirado:
jenkins retest this

elguardian:
@gmunozfe do you mind to check the latest changes ?

elguardian:
Jenkins retest  this please

gmunozfe (approved):
LGTM, tests ended ok but couldn't store artifacts, let's see again

gmunozfe:
jenkins retest this

PR #2655, opened 2021-11-25
RHPAM-3535: Is possible to create a container with a version not built

akumar074:
**JIRA**: 

[RHPAM-3535](https://issues.redhat.com/browse/RHPAM-3375)

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

### DIFF
--- a/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
+++ b/kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
@@ -50,7 +50,7 @@ public class RestSpecManagementServiceImpl {
             @ApiResponse(code = 400, message = "Controller exception"),
             @ApiResponse(code = 500, message = "Unexpected error")
     })
-    @PUT
+    @POST
     @Path("servers/{serverTemplateId}/containers/{containerId}")
     @Consumes({MediaType.APPLICATION_XML, MediaType.APPLICATION_JSON})
     public Response saveContainerSpec(@Context HttpHeaders headers,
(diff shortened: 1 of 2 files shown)


akumar074:
Jenkins run fdb

akumar074:
Jenkins run fdb

kie-server-parent/kie-server-controller/kie-server-controller-rest/src/main/java/org/kie/server/controller/rest/RestSpecManagementServiceImpl.java
```diff
@@ -90,7 +90,7 @@ public Response saveContainerSpec(@Context HttpHeaders headers,
             @ApiResponse(code = 400, message = "Controller exception"),
             @ApiResponse(code = 500, message = "Unexpected error")
     })
-    @POST
+    @PUT
```
sutaakar, line 93: That breaks backward compatibility.
What is the reason for such change?

akumar074: Hi @sutaakar If you see the code inside the POST & PUT methods currently we have opposite implementations. This change fixes the readability and API definition.
^ @cristianonicolai

sutaakar: So, you want to use POST to create a resource identified explicitly bu URL?
The backward compatibility issue I meant that if users use the POST endpoint to update containers in server template then this change will break it, forcing user to update the implementation.

akumar074: @sutaakar I agree this would need users to change the implementation in the existing applications, but I think it is also important to fix the API definition with the correct methods. 
Hi @cristianonicolai, Can you please share your thoughts on this?
Thanks

cristianonicolai: @akumar074 I think it is a valid concern, although I agree we should probably change the annotations, this is very likely to break users of the API. If we look in the original JIRA, this change wouldn't affect the reported issue in any way, so doing this change would be only recommended if extremely necessary. My 2 cents

PR #2667, opened 2021-12-15
[7.52.x][RHPAM-4038] Allow 2waySSL authentication in PIM

ruromero:
Signed-off-by: ruromero <rromerom@redhat.com>

**JIRA**:

[RHPAM-4038](https://issues.redhat.com/browse/RHPAM-4038)

* Allows KieServerHTTPRequests to use Client certificate for authentication.
* Allows PIM to define an identity certificate


elguardian:
@lazarotti  this is for 7.52 which is close. This needs to be in main branch first unless somebody says any different.
@mareknovotny  ^^

elguardian:
Jenkins retetes this please.

mareknovotny:
our streams are only main and 7.59.x so closing this.

afalhambra:
I believe, it's done this way cause 7.52.x is the last branch where PIM was still on Thorntail and since this is a specific requirement based on a customer unable to move to new PIM (on Quarkus). 
But I guess this is up to @ruromero to confirm

ruromero:
as @afalhambra just said, this PR was against this branch because it's a specific change for Thorntail. There are other 2 PR linked for main and for the Quarkus version of PIM

elguardian:
@afalhambra @ruromero @lazarotti 
upstream always first to avoid blockers and regresions

mareknovotny:
> as @afalhambra just said, this PR was against this branch because it's a specific change for Thorntail. There are other 2 PR linked for main and for the Quarkus version of PIM

we don't do here one-off patch(es), please talk to our support representative and provide that patch there.

afalhambra:
> > as @afalhambra just said, this PR was against this branch because it's a specific change for Thorntail. There are other 2 PR linked for main and for the Quarkus version of PIM
> 
> we don't do here one-off patch(es), please talk to our support representative and provide that patch there.

This is a tricky one since the last PIM on Thorntail was in 7.52.x branch. Support usually provides a one-off based on a PR/commit, and once merged, we (QE) start with the verification. 
If not possible, then this requires a different approach, @martinweiler? FYI

martinweiler:
Adding @abhijithumbe as he is checking with the customer which version the patch should be built against. Once we know, support can start creating the patch based on this PR so that QE can verify it on the patch branch. Would this work @afalhambra ?
How do we ensure similar functionality is provided in the Quarkus version, once the customer is able to upgrade @ruromero ?

afalhambra:
@martinweiler @abhijithumbe

- Regarding Quarkus, there is another PR raised for the same: https://github.com/kiegroup/process-migration-service/pull/39
Just updated jira ticket [RHPAM-4038](https://issues.redhat.com/browse/RHPAM-4038) with the 3 PRs involved here.

- Regarding the patch, it will work, but since we cannot merge this PR into any particular branch, we need to first review it properly and once this PR is approved, we can start working on the patch. 

Does it make sense for you?

lazarotti:
+1 to @afalhambra proposal - this PR should be reviewed so CEE can safely cherry-pick to one-off patch. @abhijithumbe as discussed in the support ticket, they are using in production 7.9.1, so as discussed also in a side-thread that is the one that a patch should be built. They are also requesting a patch over 7.11.1 as they are migrating to it, but I don't think it is the same priority as 7.9.1, so probably the second one-off can be done/tested after the first one is released to the customer. I will send an email about this.

lazarotti:
@afalhambra @ruromero @elguardian @abhijithumbe Citi just confirmed that a patch should be built on top of 7.11.1 (they are going live with the version in January), so 7.9.1 is not necessary.

PR #2732, opened 2022-03-10
JBPM-9979: Mark async active nodes and change default active node border

nmirasch:
Jira: https://issues.redhat.com/browse/JBPM-9979  

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>


nmirasch:
@bxf12315 I've created this PR based in your commit related with the errorBorderColor in order to:
- Keep backward compatibility just adding new calls to allow to use the new color option.
- Replace the references to errorBorderColor by asyncActiveborderColor.
- I've replaced the test by something we can found in real execution to try the fullLogs logic with a real case

Could you take a look on this one to see if there is something missing here?

bxf12315:
> @bxf12315 I've created this PR based in your commit related with the errorBorderColor in order to:
> 
> * Keep backward compatibility just adding new calls to allow to use the new color option.
I do not know why we need to create a new call? the reset call's any params could be null.  the rest API just using one places.  

> * Replace the references to errorBorderColor by asyncActiveborderColor.
Maybe failedAsyncNodeBoraderColor is better.
> * I've replaced the test by something we can found in real execution to try the fullLogs logic with a real case
> 
> Could you take a look on this one to see if there is something missing here?

nmirasch:
@bxf12315 Thanks,
* the thing is to avoid to change the current existing calls introducing that needed null. Just adding a new call if it needs to used the new border color conf, in order to avoid break in case someone is using the old one..
* failedAsyncNodeBorderColor implies it's failing but it's not completely true. They are async nodes that not have been completed yet.. At this point we can't differentiate between in 'error' or 'queued' ones.. that's the reason to use 'active' instead

nmirasch:
Jenkins retest this

bxf12315:
> * oid to change the current existing calls introducing t



> @bxf12315 Thanks,
> 
> * the thing is to avoid to change the current existing calls introducing that needed null. Just adding a new call if it needs to used the new border color conf, in order to avoid break in case someone is using the old one..
> * failedAsyncNodeBorderColor implies it's failing but it's not completely true. They are async nodes that not have been completed yet.. At this point we can't differentiate between in 'error' or 'queued' ones.. that's the reason to use 'active' instead

The two rest API has the same URL path, just has different query param and it's a GET. Maybe you new param is not work.
and I think it do not break other exits call, if the param is null, the the call also work.
Do you try the code, the async nodes is work?

nmirasch:
@bxf12315  yes, I agree we can keep just one rest endpoint.(I'm going to add that change now)  The issue was regarding the api changes method java.lang.String org.kie.server.client.UIServicesClient::getProcessInstanceImageCustomColor with different number or parameters could make mandatory to introduce changes in older calls, to use kie-server with this change.

bxf12315:
@nmirasch the code also not work. QueryParam's ASYNC_ACTIVE_BORDER_COLOR, will not work. you need to use "SVG_NODE_ASYNC_ACTIVE_BORDER_COLOR"

cristianonicolai (approved):

kie-server-parent/kie-server-remote/kie-server-client/src/main/java/org/kie/server/client/UIServicesClient.java
```diff
@@ -188,6 +188,20 @@
...
 
+    /**
+     * Returns process image (svg) with annotated active and completed nodes for given process instance
+     * that belongs to given container
+     * @param containerId container identifier where process resides
+     * @param processInstanceId unique process instance id
+     * @param completeNodeColor process instance image complete node color
+     * @param completeNodeBorderColor process instance image complete node border color
+     * @param activeNodeBorderColor process instance image active node border color
+     * @param activeAsyncNodeBorderColor process instance image error node border color
```
barboras7, line 199:
```suggestion
     * @param activeAsyncNodeBorderColor process instance image active async node border color
```

barboras7 (changes requested):

cristianonicolai:
jenkins please retest this

bxf12315:
> jenkins please retest this

Please don't merge this, it's a duplicate PR.

barboras7 (approved):

PR #2751, opened 2022-03-22
[JBPM-10046] Missing PreUndeployOperations.doNothing call for development mode and no dispose flag

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10046


gmunozfe (commented):
Add assert in test

gmunozfe (approved):
Looks good to me, thanks for fixing it, great work!
Just waiting the CI to be green

PR #2818, opened 2022-08-03
JBPM-10109: Add support for container alias in user task claim operation

martinweiler:
**JIRA**: https://issues.redhat.com/browse/JBPM-10109

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java
@@ -89,6 +89,7 @@ public class UserTaskServiceBase {
     }
 
     public void claim(String containerId, Number taskId, String userId) {
+        containerId = context.getContainerId(containerId, new ByTaskIdContainerLocator(taskId.longValue()));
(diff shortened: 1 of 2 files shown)


martinweiler:
@elguardian @fjtirado could you please take a look? I have tested the change in RHPAM and it works as expected, a user is able to claim a task when using the container alias (similar to other task operations in the same class). 

However, the integration test is not working as expected, as it is passing even without applying the change to UserTaskServiceBase, as the claim method is called with the CONTAINER_ID ("definition-project") and not the CONTAINER_ID_ALIAS ("definition-project-alias"). I could not figure out why, so I'd appreciate any hints. Thanks!

fjtirado (commented):
@martinweiler This is better than my proposal in the JIRA, so congrats!!!
I think that for completeness we might want to add the same call here
https://github.com/kiegroup/droolsjbpm-integration/blob/main/kie-server-parent/kie-server-services/kie-server-services-jbpm/src/main/java/org/kie/server/services/jbpm/UserTaskServiceBase.java#L98
so batch call also works

martinweiler:
@fjtirado thanks for your input. I added the container check to the bulk claim operation as well.

I also managed to verify the fix in the test class, it just seems to be a bit finicky to get the updates into the deployed web app used during the test.

fjtirado (approved):

fjtirado:
jenkins retest this

PR #2844, opened 2022-08-23
[JBPM-10101] kafka-clients misalignment with any supported AMQ Streams version

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10101

depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2038


mareknovotny:
@jiripetrlik  suggested to remove drools-ha completely from git, do you want to do it :) ? @elguardian

mareknovotny:
compilation error

jiripetrlik (changes requested):
@elguardian 
Is there any reason to trying to fix this instead of removing "drools-ha" completely?

mareknovotny:
i still need to remove openshift-drools-hacep repository so it is ok to keep that removal of drools-ha module for that. fyi @jiripetrlik  @elguardian

elguardian:
@mareknovotny that solution was after this PR. I am merging. You can remove it aftewards if you like.

PR #2961, opened 2023-05-21
[RHPAM-4714] wildfly-elytron as provided

lampajr:
**Thank you for submitting this pull request**

**JIRA**:

[
RHPAM-4714](https://issues.redhat.com/browse/RHPAM-4714)

**referenced Pull Requests**: 

* https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2298
* https://github.com/kiegroup/droolsjbpm-integration/pull/2961

<details>
<summary>
</summary>

</details>

<details>
<summary>
</summary>

</details>

<details>
<summary>
How to backport a pull request to a different branch?
</summary>

In order to automatically create a **backporting pull request** please add one or more labels having the following format `backport-<branch-name>`, where `<branch-name>` is the name of the branch where the pull request must be backported to (e.g., `backport-7.67.x` to backport the original PR to the `7.67.x` branch).

> **NOTE**: **backporting** is an action aiming to move a change (usually a commit) from a branch (usually the main one) to another one, which is generally referring to a still maintained release branch. Keeping it simple: it is about to move a specific change or a set of them from one branch to another.

Once the original pull request is successfully merged, the automated action will create one backporting pull request per each label (with the previous format) that has been added.

If something goes wrong, the author will be notified and at this point a manual backporting is needed.

> **NOTE**: this automated backporting is triggered whenever a pull request on `main` branch is labeled or closed, but both conditions must be satisfied to get the new PR created.
</details>

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
@@ -76,6 +76,7 @@
     <dependency>
       <groupId>org.wildfly.security</groupId>
       <artifactId>wildfly-elytron</artifactId>
+      <scope>provided</scope>
     </dependency>
 
     <!-- security -->
--- a/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
@@ -215,4 +215,21 @@
 
   </dependencies>
 
+  <profiles>
+    <profile>
+      <id>tomcat9</id>
+      <activation>
+        <property>
+          <name>container.profile</name>
+          <value>tomcat9</value>
+        </property>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>org.wildfly.security</groupId>
+          <artifactId>wildfly-elytron-auth-server</artifactId>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
 </project>
--- a/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/pom.xml
+++ b/kie-server-parent/kie-server-remote/kie-server-rest/kie-server-rest-common/pom.xml
@@ -98,6 +98,12 @@
       <artifactId>simpleclient_httpserver</artifactId>
     </dependency>
 
+    <!-- Test - Wildfly Elytron -->
+    <dependency>
+      <groupId>org.wildfly.security</groupId>
+      <artifactId>wildfly-elytron-auth-server</artifactId>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
--- a/kie-server-parent/kie-server-services/kie-server-services-jbpm/pom.xml
+++ b/kie-server-parent/kie-server-services/kie-server-services-jbpm/pom.xml
@@ -162,6 +162,7 @@
     <dependency>
       <groupId>org.wildfly.security</groupId>
       <artifactId>wildfly-elytron</artifactId>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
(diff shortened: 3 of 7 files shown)


lampajr:
Jenkins run fdb

lampajr:
Jenkins retest this

sutaakar:
It seems that https://github.com/kiegroup/droolsjbpm-integration/blob/main/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/security/ElytronIdentityProvider.java#L33 should also check for existence of SecurityDomain class in classpath.

mareknovotny:
also with this option https://github.com/wildfly-security/wildfly-elytron/blob/1.x/wildfly-elytron/pom.xml#LL171C30-L171C56 it should not be used as a dependency in other project(s) where they are not the leaf/final module see

lampajr:
@mareknovotny @sutaakar I just tried adding `wildfly-elytron-auth-server` as `test` dependency, let's see if that is enough to fix the failures.

> The root cause is to use that wildfly-elytron agregated and shaded deps jar.
the only possible even annoying is to take what we really need from all modules there :(

@mareknovotny if we want to switch to this approach I think we just need (from compilation point of view) the following deps:
- wildfly-elytron-auth-server
- wildfly-elytron-realm

but I don't know if this could introduce additional issues at runtime

mareknovotny:
the direction is still not clear as documentation for wildfly-elytron purpose is not anywhere clear to me, and it was created for easier migration from previous security module, but as i filed at that time https://issues.redhat.com/browse/ELY-1971 and it was ignored while the already 2.x and 3.x streams are existing.

lampajr:
looks like tests are fixed now

Ginxo (approved):
LGTM

kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
```diff
@@ -76,6 +76,7 @@
     <dependency>
       <groupId>org.wildfly.security</groupId>
       <artifactId>wildfly-elytron</artifactId>
+      <scope>provided</scope>
```
sutaakar, line 79: This works only for EAP.
For Tomcat and others  this approach produce ClassNotFoundException because SecurityDomain is expected to be on classpath according to current implementation - https://github.com/kiegroup/droolsjbpm-integration/blob/main/kie-server-parent/kie-server-services/kie-server-services-common/src/main/java/org/kie/server/services/impl/security/ElytronIdentityProvider.java#L33

So either the code should handle the case if SecurityDomain is not on classpath or wildfly-elytron-auth-server needs to be bundled to war files for all application servers.

sutaakar:
@MarianMacik @gmunozfe Can you please keep an eye on this PR for this week (I will be on PTO)?

kie-server-parent/kie-server-services/kie-server-services-common/pom.xml
lampajr: we could try something like that:
```
    <dependency>
      <groupId>org.wildfly.security</groupId>
      <artifactId>wildfly-elytron</artifactId>
      <scope>provided</scope>
      <exclusions>
        <exclusion>
          <groupId>org.wildfly.security</groupId>
          <artifactId>wildfly-elytron-auth-server</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
    <dependency>
      <groupId>org.wildfly.security</groupId>
      <artifactId>wildfly-elytron-auth-server</artifactId>
    </dependency>
```
In this way `wildfly-elytron-auth-server` should be bundled to war files for all application servers, wdyt?

lampajr: or even add something like that for `tomcat`:
```
  <profiles>
    <profile>
      <id>tomcat9</id>
      <activation>
        <property>
          <name>container.profile</name>
          <value>tomcat9</value>
        </property>
      </activation>
      <dependencies>
        <dependency>
          <groupId>org.wildfly.security</groupId>
          <artifactId>wildfly-elytron-auth-server</artifactId>
        </dependency>
      </dependencies>
    </profile>
  </profiles>
```

mareknovotny: that's great, thanks @lampajr !

mareknovotny (approved):

gmunozfe (approved):
looks good to me, outstanding work @lampajr !

PR #3033, opened 2024-02-29
[FIX_KAFKA_EXECUTOR] fix executor service

elguardian:
The current approach behaves like monothread.

### DIFF
--- a/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerConsumer.java
+++ b/kie-server-parent/kie-server-services/kie-server-services-kafka/src/main/java/org/kie/server/services/jbpm/kafka/KafkaServerConsumer.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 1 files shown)


martinweiler (approved):
This works properly, thanks @elguardian

porcelli (approved):
LGTM

fjtirado (changes requested):
I do not think is a good idea to create a fixed number of thread. 
According to ThreadPoolExecutor [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#ThreadPoolExecutor-int-int-long-java.util.concurrent.TimeUnit-java.util.concurrent.BlockingQueue-), new threads will be created only if needed. 
I think the behaviour @martinweiler described in the JIRA is caused because the events are already queued by the consumer. Why not testing two different topics?

fjtirado:
Ok, I misundertoos the doc (bold is mine)
_Unbounded queues. Using an unbounded queue (for example a [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html) without a predefined capacity) will cause new tasks to wait in the queue when all corePoolSize threads are busy. **Thus, no more than corePoolSize threads will ever be created. (And the value of the maximumPoolSize therefore doesn't have any effect.)** This may be appropriate when each task is completely independent of others, so tasks cannot affect each others execution; for example, in a web page server. While this style of queuing can be useful in smoothing out transient bursts of requests, it admits the possibility of unbounded work queue growth when commands continue to arrive on average faster than they can be processed._

fjtirado (approved):

PR #3046, opened 2024-04-22
[incubator-kie-issues-1105] Polymorphic json typing causes long delays in first request

elguardian:
Closes https://github.com/apache/incubator-kie-issues/issues/1105


kie-server-parent/kie-server-api/pom.xml
```diff
@@ -38,11 +38,17 @@
   </properties>
 
   <dependencies>
+    <dependency>
+        <groupId>org.projectlombok</groupId>
+        <artifactId>lombok</artifactId>
+        <version>1.18.32</version>
+        <scope>provided</scope>
+    </dependency>
```
martinweiler, line 46: I don't think we should / need to add lombok here, we could just remove all the related annotations from the dummy classes

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -347,6 +354,10 @@ public TypeDeserializer buildTypeDeserializer(DeserializationConfig config, Java
             typer2 = typer2.init(JsonTypeInfo.Id.CLASS, null);
             typer2 = typer2.inclusion(JsonTypeInfo.As.WRAPPER_OBJECT);
             deserializeObjectMapper.setDefaultTyping(typer2);
+            // this avoids the use of default typing
+            if(!defaultTypingEnabled) {
+                deserializeObjectMapper.deactivateDefaultTyping();
```
martinweiler, line 359: if we disable default typing here, wouldn't it make sense to wrap lines 327-356 in a `if(defaultTypingEnabled)` clause?

kie-server-parent/kie-server-api/src/test/java/com/example/manyrules/JSONMarshallerLargeTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assume.assumeThat;
+
+public class JSONMarshallerLargeTest {
+
+    private Logger logger = LoggerFactory.getLogger(JSONMarshallerLargeTest.class);
+
+    @Test(timeout=5000L)
+    public void testLargeNumberOfClasses() throws IOException {
+        assumeThat(JSONMarshaller.isDefaultTypingActivate(), is(false));
```
martinweiler, line 27: We should run the test by default with `org.kie.server.json.defaultTyping.enabled=false` and not skip it

kie-server-parent/kie-server-api/pom.xml
elguardian: removed

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
elguardian: done

kie-server-parent/kie-server-api/src/test/java/com/example/manyrules/JSONMarshallerLargeTest.java
elguardian: done

martinweiler (approved):
Great work @elguardian

PR #3052, opened 2024-07-16
[FIX] custom fix when xml adapter has names

elguardian:
The performance problem was related to the polymorphic feature of jackson. it computes the subtypes of a given types.
Before the patch we were saying that all classes contained in the kjar/project where subtypes of every class causing a huge impact in performance
That performance problem was causing another problem, it was overwriting the proper names for types
Every class when you are serializing you have the full qualified name and the name used for serialized and deserialized
So before the patch we were overwriting that information making this feature not working properly
Now because the subtypes are being computed properly we are not overwriting the names therefore now the name is being taken from the annotation like it should be

### DIFF
--- a/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
+++ b/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
@@ -39,9 +39,25 @@ import java.util.Set;
 import java.util.regex.Pattern;
 
 import javax.xml.bind.annotation.XmlAnyElement;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlElements;
 import javax.xml.bind.annotation.adapters.XmlAdapter;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapters;
+
+import org.drools.core.xml.jaxb.util.JaxbListAdapter;
+import org.drools.core.xml.jaxb.util.JaxbListWrapper;
+import org.drools.core.xml.jaxb.util.JaxbUnknownAdapter;
+import org.kie.server.api.KieServerConstants;
+import org.kie.server.api.marshalling.Marshaller;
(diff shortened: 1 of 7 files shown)


elguardian:
Jenkins retest this please

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -503,20 +502,34 @@ public ExtendedJaxbAnnotationIntrospector(List<NamedType> customClasses, ObjectM
...
                     Class<?> clazz = namedType.getType();
                     if (!a.getRawType().equals(clazz) && a.getRawType().isAssignableFrom(clazz)) {
                         complete.add(namedType);
                     }
                 }
             }
+
+            XmlElements elements = findAnnotation(XmlElements.class, a, false, false, false);
+            if (elements != null) {
+                for (XmlElement elem : elements.value()) {
```
porcelli, line 517: can't we check jsonHandleXmlAnyElementsNames before the loop, so we can avoid unnecessary processing (I know that over time JIT will do the job... but still)

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
```diff
@@ -1160,4 +1171,8 @@ public void setClassLoader(ClassLoader classLoader) {
     public ClassLoader getClassLoader() {
         return classLoader;
     }
+
+    public void setXmlAnyElementsNames(boolean jsonHandleXmlAnyElementsNames) {
```
porcelli, line 1175: is this method used only for testing purposes? if yes... maybe there's a better way to do so.... I personally like Anonymous classes for hack the tested class... but I also understand that it might not be the most popular design choice.

kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/json/JSONMarshaller.java
elguardian: it won't give the same outcome.

martinweiler (approved):
Excellent work @elguardian

